Issue #1, opened 2015-04-15
Draft of the spec


eyakubovich:
The spec is currently being drafted here. It'll be moved into an .md file in this repository shorly.

https://docs.google.com/document/d/1pGJiQhA_5lZsCeswsKDm5v_brkeqaT1Bmn4zANCynGg/edit?usp=sharing

jonboulle:
deprecated by the diff in #2

PR #2, opened 2015-04-16
*: first cut at SPEC and README

jonboulle:


jonboulle:
fixes #1

SPEC.md
```diff
@@ -0,0 +1,171 @@
...
+  // type (plugin) specific
+  "bridge": "ovs0",
+  "vxlanID": 42
+  // ipam specific
+  "ipam": {
+    "type": "dhcp"
+    "routes": [ "10.3.0.0/16", "10.4.0.0/16" ]
+  }
+}
+
```
eyakubovich: missing ```

jonboulle: thanks, fixed

joeswaminathan:
When specifying the routes, can we have a consistent syntax to use dictionary always, rather than a list of CIDR or a dictionary.  The "gw" can be an optional element. It will be easy parse.

joeswaminathan:
On Error is there a standard output syntax. Or just the error code is good enough. It would be better to define a standard syntax (perhaps a dictionary of "error-code", "error-msg", "error-details")

SPEC.md
```diff
@@ -0,0 +1,172 @@
...
+  "name": "dbnet",
+  "type": "bridge",
+  // type (plugin) specific
+  "bridge": "CNI0",
+  "addIf": "eth0",
+  // ipam specific
+  "ipam": {
+    "type": "host-local",
+    "subnet": "10.1.0.0/16",
+    "gateway": "10.1.0.1"
```
rajatchopra: Did you mean to write this here? The doc above does not seem to match the example, unless I misunderstood.
```
"ipam": {
   "type": "host-local",
    "routes" : [ { "dst": "10.1.0.0/16",
                       "gw": "10.1.0.1"
                     }]
}
```

eyakubovich:
We'll update to use consistent route syntax.

As for errors, they're communicated via error codes. However, stderr should be passed into the plugin and could be utilized to print error messages, logging, etc -- completely unstructured. The container runtime can just pass /dev/null if it doesn't want the user to see it. We can definitely add structured error output as well but then I would prefer that to go on stdout (this way stdout always gets JSON) and still utilize stderr as an optional unstructured dump.

rajatchopra:
General question on IPAM plugin: is the split into a different executable and standardizing its input/output only so that a main CNI vendor can utilize another vendor's IPAM plugin? Or is the container execution system ever expected to execute the IPAM plugin directly?

eyakubovich:
@rajatchopra The split is to allow users to mix and match CNI and IPAM plugins, even if they're from different vendors. It's the job of the top-level plugin to call the IPAM one. The container executioner does not invoke IPAM. We'll need to clarify that.

SPEC.md
eyakubovich: This example meant to demonstrates that "host-local" plugin supports `subnet` and `gateway` attributes. These are specific to this plugin, they're not part of the well known set. `routes` is not specified here. Now, in practice, this example makes little sense. If there were routes with just a dst but without gw, the gateway value could be used as default. I'll update to something more sane. Thanks.

joeswaminathan:
@eyakubovich I like the idea of structured output in the stdout and unstructured output in the stderr. I believe if you can standardize the structured output the container runtime can make sense out of it across different drivers. At the end the orchestration layer has to create a meaningful log for failures to facilitate trouble shooting. A standard format will help in my opinion. 

At the least a partial standardization of error code is needed (like http error codes) in my opinion

joeswaminathan:
I see the routes are defined as part of network. This would mean any container connecting to this network will have exact same route specification. But I wonder if there are any exceptions in a real world use case where some ports connecting to the same network have a different route requirements. 

The reason I am asking is, in this spec a network is synonymous to an IP subnet. There are use cases where a single IP subnet is used across different isolation domain and a firewall prevents cross talk. In such case the routes could be  different for each isolation domain.

I would suggest to move the route specification to "Connect container to Network" operation.

errordeveloper:
I have previously suggested that passing JSON on standard input would be cleaner then using environment variables for input and JSON for everything else (i.e. config file and output)... Aside from this minor detail, it all looks good to me.

errordeveloper:
I have previously suggested that passing JSON on standard input would be cleaner then using environment variables for input and JSON for everything else (i.e. config file and output)... Aside from this minor detail, it all looks good to me as far as initial revision is considered.

eyakubovich:
@joeswaminathan During the ADD command, when a container is added to the network, the job of the IPAM plugin is to return the IP and routes for _this_ particular container. It can use whatever configuration information it finds in `ipam` section together with CNI_CONTAINERID and CNI_ARGS to calculate the proper subnet/routes for this container. The intention of CNI_ARGS is actually to pass in plugin specific values for this particular container (think per-port). So CNI_ARGS can have some labels that the IPAM plugin can use to place the container into one subnet or another.

We don't have anything in the result of IPAM call that specifies the ACLs to set up. The main plugin can set up the deny rules for subnets outside of the `routes` subnets. I would be interested to see suggestions on how we can better handle ACLs here.

eyakubovich:
@errordeveloper The stdin is used to pass in JSON network configuration. The idea here is that network configuration is network-wide and env vars are container specific.

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+- Delete container from network
+  - Parameters:
+    - **Container ID**, as defined above.
+    - **Network namespace path**, as defined above.
+    - **Network configuration**, as defined above.
+    - **Extra arguments**, as defined above.
+    - **Name of the interface inside the container**, as defined above.
+
+The executable command-line API uses the type of network (see [Network Configuration](#network-configuration) below) as the name of the executable to invoke. It will then look for this executable in a list of predefined directories. Once found, it will invoke the executable using the following environment variables for argument passing:
+- `CNI_COMMAND`: indicates the desired operation; either `ADD` or `DEL`
```
rajatchopra: Not necessarily required, but we might also need an 'inpsect/show/status' command. Once the container is created with its network etc, I should be able to inspect/check-health later also.
Or the app-spec will have to store the info returned by ADD. (I don't see the ipv4/ipv6 stored there, did I miss it?)

errordeveloper:
> The stdin is used to pass in JSON network configuration. The idea here is that network configuration is network-wide and env vars are container specific.

That's why I couldn't see configuration files being mentioned any more; it makes sense this way.

SPEC.md
eyakubovich: The plugins return a JSON with the IP/net info to the caller (container runtime). I will expand it to also return gateway and routes that IPAM plugin returned. The container runtime will be responsible for storing this information and providing it to the user via inspect/show/status.

eyakubovich:
PR for plugin code: https://github.com/appc/cni/pull/3

shettyg:
I have some higher level comments with the use case being multi-host networking + multi-tenants + network virtualization.

* One should be able to pick and choose networks the container should join.
If I have a total of 100 possible networks, I should be able to choose the networks the container should join. i.e., with 100 networks, I now have 100 conf files in /etc/rkt/net.d. I should be able to start a container with the network names. This option was present in the v2 of the design doc(via -net option), but I did not see it in SPEC.md in this issue. So I wanted to make sure that this is still the case.

* The presence of files inside "/etc/rkt/net.d" should be optional.
Networks can be created and destroyed dynamically. i.e., at any given time, there could be variable number of networks in your datacenter. In any given container host, any container could belong to any network. So going ahead and creating and deleting these files in /etc/rkt/net.d is an overhead. So my suggestion is that, if a container is created via a command of the form:

rkt run --net:NAME1:PLUGINNAME1:ARGS --net:NAME2:PLUGINNAME2:ARGS image

If NAME.conf is not seen in /etc/rkt/net.d, you can simply look for PLUGINNAME and pass it the args (plus ethX name) needed.

It is possible that I have missed the reason on why this configuration file needs to be mandatory.

* Assuming the config file in /etc/rkt/net.d needs to be mandatory for some reason, in the config files placed in /etc/rkt/net.d, there is an assumption in the spec that the "name" identifies a "network". I think we should not force associate that name to a "network". It should be any random label that the plugin understands. This way, you can end up with just one conf file that simply tells the plugin name.
This though would mean that one should be able to repeat the name multiple times, like this:

rkt run --net:NAME1:ARGS --net:NAME1:ARGS image

joeswaminathan:
+1 regarding not mandating the network config file. The device could employ an offline mechanism to get the necessary config associated with the named network. Container run time could just walk through all the plugins and whichever accepts this name can take action. This way there is no need to specify the plugin in the container run time command line.

jonboulle:
@shettyg @joeswaminathan I am a bit confused what you are referring to, since we explicitly removed all mentions of the filesystem from the SPEC and only refer to loading configs via stdin, for the very reasons you outlined - i.e. this PR is _not_ just a copy of the rkt document. Can you take a look at let me know? thanks.

shettyg:
@jonboulle 
Oops. I am sorry. I looked at the reference to the google doc in spec.md and assumed that what was discussed in the doc holds true to spec.md too. To avoid confusion, could you give me an example of a typical rkt run command that this new doc recommends?

eyakubovich:
@shettyg The `rkt` command should stay largely the same as now. And while we don't have `--net` implemented now, that's still planned, even with this spec.

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+}
+```
+
+```json
+{
+  "name": "pci",
+  "type": "ovs",
+  // type (plugin) specific
+  "bridge": "ovs0",
+  "vxlanID": 42
```
jonboulle: trailing comma

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+  // ipam specific
+  "ipam": {
+    "type": "dhcp"
```
jonboulle: trailing comma

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+}
+```
+
+```json
+{
+  "name": "wan",
+  "type": "macvlan",
+  // ipam specific
+  "ipam": {
+    "type": "dhcp"
```
jonboulle: trailing comma

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+A CNI plugin is responsible for inserting a network interface into the container network namespace (e.g. one end of a veth pair) and making any necessary changes on the host (e.g. attaching other end of veth into a bridge). It should then assign the IP to the interface and setup the routes consistent with IP Address Management section by invoking appropriate IPAM plugin.
+
+### Parameters
+
+The operations that the CNI plugin needs to support are:
+
+
+- Add container to network
+  - Parameters:
+    - **Container ID**. This is optional but recommended, and should be unique across an administrative domain. For example, an environment with an IPAM system may require that each container is allocated a unique ID and that each IP allocation can thus be correlated back to a particular container. As another example, in appc implementations this would be the _pod ID_.
```
thockin: time and space unique or just space unique (e.g. GUID vs name) ?

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+    - **Network namespace path**. This represents the path to the network namespace to be added, i.e. /proc/[pid]/ns/net or a bind-mount/link to it.
+    - **Network configuration**. This is a JSON document describing a network to which a container can be joined. The schema is described below.
+    - **Extra arguments**. This allows granular configuration of CNI plugins on a per-container basis.
+    - **Name of the interface inside the container**. This is the name that should be assigned to the interface created inside the container (network namespace); consequently it must comply with the standard Linux restrictions on interface names.
```
thockin: What about installations that want multiple interfaces to be manifested in-container?

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+  - Result:
+    - **IPs assigned to the interface**. This is either an IPv4 address, an IPv6 address, or both.
+
+- Delete container from network
+  - Parameters:
+    - **Container ID**, as defined above.
+    - **Network namespace path**, as defined above.
+    - **Network configuration**, as defined above.
+    - **Extra arguments**, as defined above.
+    - **Name of the interface inside the container**, as defined above.
```
thockin: Why do you need the name on delete?  Are you trying to support partial delete?

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+
+The executable command-line API uses the type of network (see [Network Configuration](#network-configuration) below) as the name of the executable to invoke. It will then look for this executable in a list of predefined directories. Once found, it will invoke the executable using the following environment variables for argument passing:
+- `CNI_COMMAND`: indicates the desired operation; either `ADD` or `DEL`
+- `CNI_CONTAINERID`: Container ID
+- `CNI_NETNS`: Path to network namespace file
+- `CNI_IFNAME`: Interface name to set up
+- `CNI_ARGS`: Extra arguments passed in by the user at invocation time. Alphanumeric key-value pairs separated by semicolons; for example, "FOO=BAR;ABC=123"
+- `CNI_PATH`: Colon-separated list of paths to search for CNI plugin executables
```
thockin: This implies that one plugin might chain to others - is that the intent?

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+```
+
+Errors are indicated by a non-zero return code and error details printed to stderr.
+
+### Network Configuration
+
+The network configuration is described in JSON form. The configuration can be stored on disk or generated from other sources by the container runtime. The following fields are well-known and have the following meaning:
+- `name` (string): Network name. This should be unique across all containers on the host (or other administrative domain).
+- `type` (string): Refers to the filename of the CNI plugin executable.
+- `ipMasq` (boolean): Optional (if supported by the plugin). Set up an IP masquerade on the host for this network.
```
thockin: It's not clear to me how the ipMasq flag is to be used, or what it is for

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+- `ipam`: Dictionary with IPAM specific values:
+  - `type` (string): Refers to the filename of the CNI plugin executable.
```
thockin: of which CNI plugin?  The currently executing one?  A different one?  if so which?

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+  - `routes` (list): List of subnets (in CIDR notation) that the CNI plugin should ensure are reachable by routing them through the network. Each entry is a dictionary containing:
```
thockin: why is routes under IPAM?

SPEC.md
eyakubovich: Space unique. No two containers that are currently running should have the same Container ID. The space is defined somewhat loosely as "administrative domain".

SPEC.md
eyakubovich: This is exactly to accommodate multiple interfaces inside a single container. Each plugin is invoked to connect a container to a single "network". This operations results in the plugin creating an interface inside the network namespace. However, the container runtime can invoke multiple plugins against the same container/network namespace. This allows the container to join multiple networks (e.g. container-to-container overlay and Internet). To make sure that the plugins don't collide on the interface name, it's passed in (e.g. first one gets `eth0`, second one `eth1` and so on).

SPEC.md
eyakubovich: See above. This is the name of the interface that this network set up. This allows the plugins not to store network name -> ifname mapping. We want plugins to be as stateless as possible.

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+    "routes": <list-of-ipv6-routes>    (optional)
+  }
+}
+```
+
+`gateway` is the default gateway for this subnet, if one exists.
+It does not instruct the CNI plugin to add any routes with this gateway: routes to add are specified separately via the `routes` field.
+An example use of this value is for the CNI plugin to add this IP address to the linux-bridge to make it a gateway.
+
+Each route entry is a dictionary with the following fields:
```
thockin: Are routes expected to be added by the parent plugin or is this declaring that they have been created?

SPEC.md
eyakubovich: Yes. The primary use is for the "main" plugin to invoke the IPAM plugin to get the IP and routes.

SPEC.md
```diff
@@ -0,0 +1,176 @@
...
+Errors are indicated by a non-zero return code and error details printed to stderr.
+
+IPAM plugin examples:
+ - **host-local**: Select an unused (by other containers on the same host) IP within the specified range.
+ - **dhcp**: Use DHCP protocol to acquire and maintain a lease. The DHCP requests will be sent via the created container interface; therefore, the associated network must support broadcast.
+
+#### Notes
+
+ - Routes are expected to be added with a 0 metric.
+ - A default route may be specified via "0.0.0.0/0". Since another network might have already configured the default route, the CNI plugin should be prepared to skip over its default route definition.
```
thockin: I don't understand this.

SPEC.md
eyakubovich: If joining a network requires multiple interfaces, it would not be supported out of the box. It can be emulated by defining two networks that are backed by the same plugin. Do you think this is a common case?

SPEC.md
thockin: My point was that you should clarify the doc :)

SPEC.md
thockin: It's a different approach than I had considered.  I envisioned a single plugin being active, and that plugin being responsible for everything.  If you wanted to compose multiple plugins, you would write a wrapper plugin that invoked others.

Just so I understand, tell me where I get it wrong:

Admin configures the local machine to say "use net plugins Foo and Bar".

Pod gets created

Runtime calls Foo with CNI_COMMAND=ADD and CNI_IFNAME=eth0

Runtime calls Bar with CNI_COMMAND=ADD and CNI_IFNAME=eth1

Runtime creates app containers

thockin: I find this API a bit proscriptive - I feel like there are interesting things I might do in a plugin that don't involve creating an interface or involve creating more than one.  Can't this be more opaque?  Call plugins, in succession.  Let each plugin return some status which you merge together into the final result.  Let them deal with allocating a unique interface name (maybe offer a psedocode algorithm) if they need it.  Let them return status strings that might be useful in logs/events.  

Or put another way - what does this more rigid interface you have defined offer in exchange for its rigidness?

SPEC.md
jonboulle: This was a struggle since it's intentionally broad - any suggestions for clarifying "administrative domain"? (it might _also_ be time unique depending on environment requirements...)

README.md
```diff
@@ -1 +1,18 @@
...
+## Why develop CNI?
+
+Application containers on Linux are a rapidly evolving area, and within this space networking is a particularly unsolved problem, as it is highly environment-specific.
+We believe that every container runtime will seek to solve the same problem of making the network layer pluggable.
+In order to avoid duplication, we think it is prudent to define a common interface between the network plugins and container execution.
+Hence we are proposing this specification, along with an initial set of plugins that can be used by different container runtime systems.
+
+## How do I use CNI?
+
+TODO...
```
jonboulle: This should be rebased against master to capture the plugins now

SPEC.md
thockin: As with the above, I'd rather see this be opaque.  Plugins can store "state" in the interface name, so they don't have to be fed that information but can instead discover it.

SPEC.md
thockin: I read that further on, but I am worried about the comingling of plugins that serve different purposes.  If I have a DHCP plugin for IPAM, what happens if I accidentally configure my runtime to use DHCP as a top-level plugin?  I think there should be a wall between CNI plugins and and CNI/IPAM plugins.  This also leaves room for other "second level" plugin flavors to emerge

SPEC.md
jonboulle: This is a bug, should be "IPAM plugin". We struggled greatly with naming here, of course. Technically both types are CNI plugins - originally we called them "main plugin" and "IPAM plugin", ended up compromising on "CNI plugin" and "IPAM plugin". Less than ideal, suggestions welcomed.

SPEC.md
jonboulle: So essentially you are insisting that plugins be cooperative, right? I think that's one of the major issues I see with making this opaque.

SPEC.md
jonboulle: Makes sense, we should very explicitly delineate the two.

SPEC.md
thockin: simply saying "at a point in time" or "as long as this network is active" or noting that this ID may be reused over time.

SPEC.md
jonboulle: Well assuming single default route (no policy routing), it's "first default route wins".

SPEC.md
thockin: I don't see a better choice.  Either:

a) Assert a very rigid boundary in every way imaginable (ifnames, routes, iptables, sysctls, ...)

b) Assert that plugins should try to be cooperative, try to detect when they are not

c) Assert that a runtime uses a single plugin, and punt multi-plugin operation down to the user or system-integrator to make and test a wrapper plugin that works with known wrappee plugins

I was leaning towards C, but I could see B.

SPEC.md
eyakubovich: @thockin First off, your understanding of the operation is correct.

The reason we wanted to support multiple plugins (or more precisely, joining multiple networks) without writing wrapper code is to make it easy for the user (or admin). I think it will be a common enough operation that it'd be great to do that without custom scripts/code. The big motivator was actually rkt itself: we needed the container to always join the "default" network that would be able to reach the metadata service. It then got expanded to also provide the default route (for Internet) via the host with IP masquerade enabled. The user could then specify additional networks (like container-to-container).

The prescriptive design of 1 network = 1 interface was chosen to:
1) Make it easy to give familiar and predictable ethN interface names. Otherwise the plugin would need to probe to see what name is available.
2) To make status/diagnostics for the container runtime easier. We know that every network results in one interface with one IP address (ok, can be two, IPv4 and IPv6).
3)  Since our abstraction is "joining a network", it seemed like it would be difficult to do so without creating an interface. They only way I could see that it could be accomplished is by re-using an existing interface. However, this would make networks not independent. I can see a case for creating multiple interfaces to join a single network. But that should be rare enough that it can be worked around. For example, go ahead and create the first one with CNI_IFNAME, subsequent ones with random letters and report up only the first one.

SPEC.md
eyakubovich: The reason we need to support IP masquerade is because very often the IPs assigned to the interface are not routable outside their subnet. e.g., that's an issue with Docker, no news here. Even with something like flannel, these IPs are not routable outside of flannel network. So if a network/user wishes to provide access to "outside" subnets, the container's IP needs to be NAT'ed at the host. That's what this flag is for.

Now, this flag is completely up to the plugin to implement or not. In general, this spec does not want to define fields that are plugin specific. However, to avoid every plugin calling it differently, we decided to standardize on the name.

SPEC.md
eyakubovich: This field is very much like `ipMasq` -- it's interpreted by the plugin but in the interest of consistent naming, we put its definition in the spec. We're modeling IPAM plugin semantics from DHCP server. DHCP server will return IP, subnet, and optionally, routes to install. That's why the result of IPAM execution has routes to install. These routes (in config) are specified by the user as something that he wishes to add. The IPAM plugin should return them as part of the result (e.g. merge them routes from DHCP response).

SPEC.md
eyakubovich: Routes (and IP) are assigned by the parent plugin. The IPAM plugin just returns the values but doesn't actually apply them.

SPEC.md
eyakubovich: The assumption now is that the plugins are co-operative to the extent that all these resources (routes, iptables, etc) are shared so they need to:
a) take reasonable steps to namespace keys: e.g. iptables routes should be prefixed by something unique, like network name.
b) be ready that a shared resource may be busy (e.g. default route already installed) and either skip what their action or error out.

Of course, not every combo is going to work. We have punted on creating safety nets to ensure conflicting plugins don't step on each other (w/ exception of ifname being passed in).

SPEC.md
thockin: You provided your own example of a plugin that doesn't create an interface: IP masquerade

Re: (1) - having the plugin probe for names is not so bad.  If this is a big deal, you could do something like pass the index of the first available interface in.  If the plugin does not return a new interface, you pass that same index to the next plugin.  e.g. first plugin gets index 0.  If it creates eth0, the second plugin gets index 1.  But even this feels wrong.  You should not be dictating the name of interfaces - that should be determined by the plugins based on what kind of interface they are.

Re: (2) - I don't buy it.  You're targetting a universal standard (aren't we all?).  You need to be looser.  What about plugins that simply tear down an interface?  E.g. Kubernetes might need an "un-docker" plugin.  What about plugins that do iptables work and no interface.

I'm not actually a networking expert (I just play one on the Internet), but to methis feels overly bound to a single pre-supposed model of operation.  It's your plugin API, you just asked for feedback, so here it is.

SPEC.md
thockin: Your words say B, but your spec is closer to A.

SPEC.md
thockin: But who decides whether or not to set this field?  It feel weirdly out of place and disconnected...

SPEC.md
thockin: I don't follow.  If DHCP is an IPAM module, this should be OUTPUT from a plugin not input?

SPEC.md
```diff
@@ -0,0 +1,209 @@
...
+    - **Container ID**, as defined above.
+    - **Network namespace path**, as defined above.
+    - **Network configuration**, as defined above.
+    - **Extra arguments**, as defined above.
+    - **Name of the interface inside the container**, as defined above.
+
+The executable command-line API uses the type of network (see [Network Configuration](#network-configuration) below) as the name of the executable to invoke. It will then look for this executable in a list of predefined directories. Once found, it will invoke the executable using the following environment variables for argument passing:
+- `CNI_COMMAND`: indicates the desired operation; either `ADD` or `DEL`
+- `CNI_CONTAINERID`: Container ID
+- `CNI_NETNS`: Path to network namespace file
```
jonboulle: We need to clarify that this can potentially change and hence cannot be relied on to be consistent for the life of a container - https://github.com/coreos/rkt/issues/851#issuecomment-98267431

eyakubovich:
Merging as is for now. From now on, please file issues.

thockin:
I'm happy to discuss more, because I really want this to be successful.  That said, I am going to have a hard time filing and following up on a bunch of somewhat fundamental issues right now.  If you are open to considering some of the bigger things I have brought up, maybe we can figure out what the most time-efficient next steps are?

PR #8, opened 2015-05-03
plugins: add ipvlan main plugin

### LINKED ISSUE #6: plugins: add ipvlan main plugin
Support for *ipvlan* devices would be nice to have. This should not be much different from *macvlan*.

steveeJ:
Add the ipvlan main plugin which is heavily based on the macvlan plugin.
Availabile modes for this plugin are l2 and l3, wheres l2 has the higher
compatibility due to support for multicast and broadcasts. L2 has therefore been
chosen as the default mode.
See the official docs at
https://www.kernel.org/doc/Documentation/networking/ipvlan.txt
for more information.

Fixes #6.


eyakubovich:
LGTM. Thank you for contributing.

PR #11, opened 2015-05-06
Do not use netns as ID or for entropy

### LINKED ISSUE #5: plugins: audit use of netns path, mandate containerid?
As discovered and discussed in https://github.com/coreos/rkt/issues/851 - there are problems with the ways that some of the plugins utilise the `CNI_NETNS` parameter:
- the veth plugin [uses it as a source of entropy](https://github.com/appc/cni/blob/88377fa3466eea30817b038c69ac06686e7655b7/plugins/main/veth/veth.go#L50) for generating interface names for the host. Unfortunately, rkt supplies a (consistent) _relative_ path to this file (which is useful for other reasons and not strictly prohibited by the current wording of the spec). Consequently, multiple invocations of the veth plugin for different containers end up using conflicting "entropy".
- macvlan [has a similar issue](https://github.com/appc/cni/blob/88377fa3466eea30817b038c69ac06686e7655b7/plugins/main/macvlan/macvlan.go#L114)
- the host-local IPAM allocator [uses the (raw) netns name](https://github.com/appc/cni/blob/88377fa3466eea30817b038c69ac06686e7655b7/plugins/ipam/host-local/main.go#L51) as an identifier for its local store; in this case, multiple allocations happen to succeed but [removals will conflict](https://github.com/appc/cni/blob/88377fa3466eea30817b038c69ac06686e7655b7/plugins/ipam/host-local/backend/disk/backend.go#L71)

Some of these problems would be addressed by mandating that the `CNI_NETNS` parameter must be an absolute path (which should by definition mean it is unique to that container), but this introduces other issues (for example, in rkt this path might change between when its networking is being set up and destroyed).

A better solution is probably going to be to make `CNI_CONTAINERID` required, rather than optional as it currently is in the latest draft of the spec, and have plugins use this as a unique container identifier (and safe source of entropy) instead.

eyakubovich:
ContainerID is now required so use that
or generate random bytes.

Fixes #5

### DIFF
--- a/pkg/ip/link.go
+++ b/pkg/ip/link.go
@@ -15,7 +15,7 @@
 package ip
 
 import (
-	"crypto/sha512"
+	"crypto/rand"
 	"fmt"
 	"net"
 	"os"
@@ -23,7 +23,7 @@ import (
 	"github.com/appc/cni/Godeps/_workspace/src/github.com/vishvananda/netlink"
 )
 
-func makeVeth(name, peer string, mtu int) (netlink.Link, error) {
+func makeVethPair(name, peer string, mtu int) (netlink.Link, error) {
 	veth := &netlink.Veth{
 		LinkAttrs: netlink.LinkAttrs{
 			Name:  name,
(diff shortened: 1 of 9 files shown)


pkg/ip/link.go
```diff
@@ -39,37 +39,61 @@ func makeVeth(name, peer string, mtu int) (netlink.Link, error) {
 	return veth, nil
 }
 
+func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err error) {
+	for {
```
jonboulle: infinite loops scare me, can you add a sanity limit/panic?

jonboulle:
lgtm

pkg/ip/link.go
eyakubovich: Trust very small probabilities not occurring. But not in my coding skills so I'll add a sanity check.

jonboulle: Yeah the chance now might be minuscule but in anticipation of something changing or a bug elsewhere, etc etc (especially as this calls other functions)

PR #19, opened 2015-06-08
Propagate json error object to the caller

eyakubovich:
When plugin errors out, it prints out a JSON object to stdout
describing the failure. This object needs to be propagated out
through the plugins and to the container runtime. This change
also adds Print method to both the result and error structs
for easy serialization to stdout.

### DIFF
--- a/pkg/skel/skel.go
+++ b/pkg/skel/skel.go
@@ -64,12 +64,12 @@ func PluginMain(cmdAdd, cmdDel func(_ *CmdArgs) error) {
 	}
 
 	if argsMissing {
-		die("required env variables missing")
+		dieMsg("required env variables missing")
 	}
 
 	stdinData, err := ioutil.ReadAll(os.Stdin)
 	if err != nil {
-		die("error reading from stdin: %v", err)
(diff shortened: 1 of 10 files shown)


jonboulle:
s/propogate/propagate fyi

pkg/skel/skel.go
```diff
@@ -93,14 +93,20 @@ func PluginMain(cmdAdd, cmdDel func(_ *CmdArgs) error) {
 	}
 
 	if err != nil {
+		if e, ok := err.(*plugin.Error); ok {
+			// don't wrap Error in Error
+			e.Print()
```
jonboulle: this is swallowing the error from Print?

eyakubovich: yeah but what you gonna do. it's already exiting with non-zero status and failed to write the error. i'll add a log to stderr.

jonboulle: OK. Also ideally it would be nice if the Exit() was in a single place that these two paths shared...

eyakubovich:
@jonboulle PTAL

jonboulle:
lgtm

PR #20, opened 2015-06-11
add flannel plugin

eyakubovich:
This introduces a notion of a "meta" plugin. A meta plugin
is one that delegates the actual work of setting up the interface
to the main plugin. The meta plugin is used to select and dynamically
configure the main plugin. The sequence of events, is as follows:

Given netconf like:

```
{
    "name": "mynet",
    "type": "flannel",
    "delegate": {
        "type": "bridge"
    }
}
```
flannel fills in values like "mtu", "ipam.subnet" and delegates to
"bridge" main plugin. "bridge" plugin will operate as usual, calling
into ipam module for IP assignment.

Delegate dictionary should not contain "name" field as it will be
filled in by the flannel plugin.

### DIFF
--- a/plugins/meta/flannel/flannel.go
+++ b/plugins/meta/flannel/flannel.go
@@ -1,0 +1,216 @@
+// Copyright 2015 CNI Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This is a "meta-plugin". It reads in its own netconf, combines it with
+// the data from flannel generated subnet file and then invokes a plugin
+// like bridge or ipvlan to do the real work.
+
+package main
+
+import (
+	"bufio"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/appc/cni/pkg/plugin"
+	"github.com/appc/cni/pkg/skel"
+)
+
+const (
+	defaultSubnetFile = "/run/flannel/subnet.env"
+	stateDir          = "/var/lib/cni/flannel"
+)
+
+type NetConf struct {
+	plugin.NetConf
+	SubnetFile string                 `json:"subnetFile"`
+	Delegate   map[string]interface{} `json:"delegate"`
+}
+
+type subnetEnv struct {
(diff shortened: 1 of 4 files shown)


plugins/meta/flannel/flannel.go
```diff
@@ -0,0 +1,213 @@
+// Copyright 2014 CoreOS, Inc.
```
jonboulle: 2015 fwiw

philips: haven't we sort of settled on the "Flannel Authors" style of copyright notices?

plugins/meta/flannel/flannel.go
```diff
@@ -0,0 +1,213 @@
...
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/appc/cni/pkg/plugin"
+	"github.com/appc/cni/pkg/skel"
+)
+
+const defaultSubnetFile = "/run/flannel/subnet.env"
+const scratchDir = "/var/lib/cni/flannel"
```
jonboulle: why is this scratch?

plugins/meta/flannel/flannel.go
jonboulle: I'd prefer that across the board but no-one else has really followed it

jonboulle: @eyakubovich global "CNI Authors" change after this?

jonboulle:
LGTM overall, but if possible I would like to make this stuff more discoverable... e.g. the explanation of the meta-plugin, perhaps in `plugins/meta/flannel/doc.go` rather than in just this commit message.

plugins/meta/flannel/flannel.go
eyakubovich: Yes, should be CNI Authors, I tend to blindly copy/paste these headers

plugins/meta/flannel/flannel.go
eyakubovich: This is a directory that this plugin uses to write out extra state that it needs for the DEL command. So it's kind of a scratch paper.

eyakubovich:
Yeah, i'm struggling whether to put plugin documentation in CNI or in rkt. On the one hand, it obviously belongs in CNI repo but from a practical point of view, everyone will be looking it up in rkt (or whatever container runtime that ship these). But I'll at least add a link to the README.md

plugins/meta/flannel/flannel.go
jonboulle: ehh, "scratch" to me implies it can/should be thrown away at any time whereas we really want this to be persistent between add/deletes

eyakubovich: You got a point there. I'll think of a better name.

jonboulle:
@eyakubovich yeah in this case I am talking about more general structural documentation for developers/users of this repository. But on your point, I think docs should be here and we can work on linking them approrpiately from rkt.

plugins/meta/flannel/flannel.go
```diff
@@ -0,0 +1,215 @@
+// Copyright 2015 CNI Authoris.
```
jonboulle: authoris.

PR #21, opened 2015-06-12
No more path rewriting

eyakubovich:
Path rewriting causes too many problems when vendoring
vendored code. When CNI code is vendored into rkt,
godep has problems code already vendored by CNI.

### DIFF
--- a/build
+++ b/build
@@ -9,7 +9,7 @@ if [ ! -h gopath/src/${REPO_PATH} ]; then
 fi
 
 export GOBIN=${PWD}/bin
-export GOPATH=${PWD}/gopath
+export GOPATH=${PWD}/gopath:$(pwd)/Godeps/_workspace
 
 echo "Building plugins"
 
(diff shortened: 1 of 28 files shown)


build
```diff
@@ -9,7 +9,7 @@ if [ ! -h gopath/src/${REPO_PATH} ]; then
 fi
 
 export GOBIN=${PWD}/bin
-export GOPATH=${PWD}/gopath
+export GOPATH=${PWD}/gopath:`pwd`/Godeps/_workspace
```
jonboulle: Godeps should be first, surely

jonboulle: also backtick? what is this 1999?

eyakubovich: Why Godeps first? Wouldn't you want code local to this repo to be searched before dependencies?

jonboulle: I misread slightly, thought it was including predefined GOPATH. In this
case, either option seems equally unintuitive to me, so this is fine :P

jonboulle:
ok

PR #22, opened 2015-06-15
dhcp: don't swallow err from netns switch

eyakubovich:


plugins/ipam/dhcp/lease.go
```diff
@@ -72,29 +72,34 @@ func AcquireLease(clientID, netns, ifName string) (*DHCPLease, error) {
...
-			errCh <- fmt.Errorf("error looking up %q", ifName)
-			return
-		}
+	go func() {
+		err := ns.WithNetNSPath(netns, true, func(_ *os.File) (e error) {
+			defer l.wg.Done()
+
+			link, err := netlink.LinkByName(ifName)
+			if err != nil {
+				errCh <- fmt.Errorf("error looking up %q", ifName)
```
jonboulle: Shouldn't you be removing these sends, otherwise you're pushing the error onto the channel twice

jonboulle: actually, different errors...

plugins/ipam/dhcp/lease.go
```diff
@@ -72,29 +72,34 @@ func AcquireLease(clientID, netns, ifName string) (*DHCPLease, error) {
...
 
-		l.link = link
+			l.link = link
 
-		if err = l.acquire(); err != nil {
-			errCh <- err
-			return
-		}
+			if err = l.acquire(); err != nil {
+				errCh <- err
```
jonboulle: Likewise, although here it's the same error twice

plugins/ipam/dhcp/lease.go
```diff
@@ -72,29 +72,34 @@ func AcquireLease(clientID, netns, ifName string) (*DHCPLease, error) {
...
+				return
+			}
 
-		log.Printf("%v: lease acquired, expiration is %v", l.clientID, l.expireTime)
+			log.Printf("%v: lease acquired, expiration is %v", l.clientID, l.expireTime)
 
-		errCh <- nil
+			errCh <- nil
```
jonboulle: omit?

plugins/ipam/dhcp/lease.go
```diff
@@ -72,29 +72,34 @@ func AcquireLease(clientID, netns, ifName string) (*DHCPLease, error) {
...
-	go ns.WithNetNSPath(netns, true, func(_ *os.File) (e error) {
-		defer l.wg.Done()
-
-		link, err := netlink.LinkByName(ifName)
-		if err != nil {
-			errCh <- fmt.Errorf("error looking up %q", ifName)
-			return
-		}
+	go func() {
+		err := ns.WithNetNSPath(netns, true, func(_ *os.File) (e error) {
```
jonboulle: The way I'd see it happening would be `errCh <- ns.WithNetNSPath()` and for the closure to return any errors to allow the wrapper to propagate - but maybe you'retrying to do something slightly different?

plugins/ipam/dhcp/lease.go
eyakubovich: I guess I can omit the `nil` and just close the channel. But there needs to be something to tell the other goroutine that the lease has been acquired ok.

jonboulle: https://github.com/appc/cni/pull/22/files#r32458513

plugins/ipam/dhcp/lease.go
eyakubovich: The tricky part here is that in case of error, that closure returns. If it succeeds in getting a lease, it lingers around to keep the lease maintained. `l.maintain()` doesn't return until it gets signaled by `l.Stop()`

jonboulle: oh! OK

plugins/ipam/dhcp/lease.go
eyakubovich: Don't follow. The error is only pushed once and then the closure immediately returns. But the closure always returns `nil`. I know that it's messy but that was the best I came up with after multiple iterations.

jonboulle: The closure returns err (!= nil), then the goroutine also pushes that error into the ch on line 100 right?

jonboulle:
LGTM, meatspace was easier.

PR #24, opened 2015-06-15
dhcp: make netns path absolute prior to RPC

eyakubovich:
The dhcp daemon maybe running with a different cwd
and so the netns paths need to be absolute. This patch
also refactors the code to pull out the common, RPC
parts, out into a separate function.

### DIFF
--- a/plugins/ipam/dhcp/main.go
+++ b/plugins/ipam/dhcp/main.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"net/rpc"
 	"os"
+	"path/filepath"
 
 	"github.com/appc/cni/pkg/plugin"
 	"github.com/appc/cni/pkg/skel"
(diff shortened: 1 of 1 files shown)


plugins/ipam/dhcp/main.go
```diff
@@ -34,30 +35,38 @@ func main() {
...
+
+func rpcCall(method string, args *skel.CmdArgs, result interface{}) error {
 	client, err := rpc.DialHTTP("unix", socketPath)
 	if err != nil {
 		return fmt.Errorf("error dialing DHCP daemon: %v", err)
 	}
 
-	dummy := struct{}{}
-	err = client.Call("DHCP.Release", args, &dummy)
+	// The daemon maybe running under a differend working dir
```
jonboulle: may be, different

jonboulle:
Two typos but LGTM

PR #32, opened 2015-07-10
ipvlan: wrong name was used during link creation

eyakubovich:
Instead of temp (random) name, the final name (e.g. eth0)
was used during link creation. This would collide on hosts
that already had the an interface with such a name.


eyakubovich:
@steveeJ PTAL

steveeJ:
I wonder how this slipped in, since I essentially forked macvlan. Also weird that the call to `renameLink` doesn't error out.

Anyways: good catch! :+1:

eyakubovich:
@steveeJ actually renameLink was erroring for me -- that's how I caught it. So LGTM?

steveeJ:
yep, LGTM!

PR #33, opened 2015-07-10
Rename veth plugin to ptp and add docs.

eyakubovich:
ptp (point-to-point) better describes the nature of the plugin
as veth pairs are also used by the bridge plugin.


iaguis:
LGTM

PR #41, opened 2015-08-11
flannel: set up route for the entire flannel network

eyakubovich:
Without it, packets leave via default route iface
and come back via interface added by flannel plugin.
If rp_filter=1, the packets are dropped.

### DIFF
--- a/plugins/meta/flannel/flannel.go
+++ b/plugins/meta/flannel/flannel.go
@@ -45,6 +45,7 @@ type NetConf struct {
 }
 
 type subnetEnv struct {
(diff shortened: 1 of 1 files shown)


eyakubovich:
@steveeJ PTAL

steveeJ:
Approach and code look good on first sight. Allow me some time on the weekend to do a flannel set up for a full verification.

steveeJ:
I was able to reproduce the situation and verify that your PR allows communication with other subnets in the flannel network.

LGTM

Issue #39, opened 2015-08-05
plugins: define runtime configurable attributes


steveeJ:
Which plugins should support configurable attributes?
What should these be?

---
- main
  - bridge
    - [ ] MAC
  - macvlan
    - [ ] MAC
- ipam
  - host-local(-ptp)
    - [x] IP

squeed:
we totally do this now with conventions. I'm marking this closed.

PR #42, opened 2015-08-16
host-local: allow ip request via CNI_ARGS

steveeJ:
A specific IP can now be requested via the environment variable CNI_ARGS, e.g.
`CNI_ARGS=ip=1.2.3.4`.
The plugin will try to reserve the specified ip.
If this is not successful the execution will fail.

Working on #39.

### DIFF
--- a/pkg/plugin/args.go
+++ b/pkg/plugin/args.go
@@ -1,0 +1,36 @@
+package plugin
+
+import (
+	"encoding"
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+func LoadArgs(args string, container interface{}) error {
+	if args == "" {
+		return nil
+	}
+
+	containerValue := reflect.ValueOf(container)
+
+	pairs := strings.Split(args, ",")
+	for _, pair := range pairs {
+		kv := strings.Split(pair, "=")
+		if len(kv) != 2 {
+			return fmt.Errorf("ARGS: invalid pair %q", pair)
+		}
+		keyString := kv[0]
+		valueString := kv[1]
+		keyField := containerValue.Elem().FieldByName(keyString)
+		if !keyField.IsValid() {
+			return fmt.Errorf("ARGS: invalid key %q", keyString)
+		}
+		u := keyField.Addr().Interface().(encoding.TextUnmarshaler)
+		err := u.UnmarshalText([]byte(valueString))
+		if err != nil {
+			return fmt.Errorf("ARGS: error parsing value of pair %q: %v)", pair, err)
+		}
+	}
+	return nil
+}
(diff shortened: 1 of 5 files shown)


pkg/plugin/args.go
```diff
@@ -0,0 +1,29 @@
...
+			return fmt.Errorf("No valid argument pair: %v", pair)
+		}
+		js += fmt.Sprintf(`"%v":"%v"`, kv[0], kv[1])
+		if i+1 < len(pairs) {
+			js += `,`
+		}
+	}
+	js += `}`
+
+	return json.Unmarshal([]byte(js), container)
```
eyakubovich: Bonus points if you can do this function without serializing to JSON first. Want to learn `reflect` api?

plugins/ipam/host-local/allocator.go
```diff
@@ -78,6 +78,40 @@ func (a *IPAllocator) Get(id string) (*plugin.IPConfig, error) {
 		gw = ip.NextIP(a.conf.Subnet.IP)
 	}
 
+	var requestedIP *net.IP
```
eyakubovich: `net.IP` instead of `*net.IP`. `net.IP` is a byte slice and so can be `nil`.

plugins/ipam/host-local/allocator.go
```diff
@@ -78,6 +78,40 @@ func (a *IPAllocator) Get(id string) (*plugin.IPConfig, error) {
...
+	if a.conf.Args != nil {
+		requestedIP = &a.conf.Args.IP
+	}
+
+	if requestedIP != nil {
+		if gw != nil && gw.Equal(a.conf.Args.IP) {
+			return nil, fmt.Errorf("requested IP must not equal IP")
```
eyakubovich: This error message is odd.

plugins/ipam/host-local/config.go
```diff
@@ -32,6 +32,11 @@ type IPAMConfig struct {
 	Subnet     ip.IPNet       `json:"subnet"`
 	Gateway    net.IP         `json:"gateway"`
 	Routes     []plugin.Route `json:"routes"`
+	Args       *IPAMArgs
```
eyakubovich: Add `json:"-"` to avoid picking it up from the conf file.

eyakubovich:
@steveeJ Appologies for late review.

steveeJ:
@eyakubovich Updated, thanks for the review!

pkg/plugin/args.go
```diff
@@ -0,0 +1,36 @@
...
+	for _, pair := range pairs {
+		kv := strings.Split(pair, "=")
+		if len(kv) != 2 {
+			return fmt.Errorf("ARGS: no valid pair '%v'", pair)
+		}
+		keyString := kv[0]
+		valueString := kv[1]
+		keyField := containerValue.Elem().FieldByName(keyString)
+		if !keyField.IsValid() {
+			return fmt.Errorf("ARGS: invalid key '%s'", keyString)
```
eyakubovich: It's fine but there's `%q` for quoted string -- really nifty.

eyakubovich:
LGTM

PR #50, opened 2015-09-04
bridge: add isDefaultGateway field

### LINKED ISSUE #27: bridge: does not set the gateway on the route
First raised in https://github.com/coreos/rkt/issues/862

bridge does correctly sets the IP on on the bridge (if `isGateway` is true) but not on the route.

### LINKED ISSUE #184: flannel plugin should allow setting default routes
Currently the flannel plugin, when generating config for the `delegate` IPAM seems to be setting only routes to the flannel network. However, it should be setting up a default route as well. Otherwise containers launched using the flannel plugin cannot communicate with anything other than containers within the flannel network.

steveeJ:
When isDefaultGateway is true it automatically sets isGateway to true.
The default route will be added via the (bridge's) gateway IP.
If a default gateway has been configured via IPAM in the same configuration file, the plugin will error out.

Fixes #27.
Fixes #184.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -37,6 +37,7 @@ type NetConf struct {
 	types.NetConf
 	BrName      string `json:"bridge"`
 	IsGW        bool   `json:"isGateway"`
+	IsDefaultGW bool   `json:"isDefaultGateway"`
 	IPMasq      bool   `json:"ipMasq"`
 	MTU         int    `json:"mtu"`
 	HairpinMode bool   `json:"hairpinMode"`
@@ -185,6 +186,10 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
+	if n.IsDefaultGW {
+		n.IsGW = true
+	}
+
 	br, err := setupBridge(n)
 	if err != nil {
 		return err
@@ -206,6 +211,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
+	// TODO: make this optional when IPv6 is supported
 	if result.IP4 == nil {
 		return errors.New("IPAM plugin returned missing IPv4 config")
 	}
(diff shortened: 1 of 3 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -196,6 +201,28 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	// set the default gateway if requested
+	if n.IsDefaultGW {
+		_, defaultNet, err := net.ParseCIDR("0.0.0.0/0")
+		if err != nil {
+			return err
+		}
+		for i, route := range result.IP4.Routes {
+			if defaultNet.String() == route.Dst.String() {
+				if route.GW != nil && !route.GW.Equal(result.IP4.Gateway) {
+					return errors.New(fmt.Sprintf(
```
eyakubovich: `fmt.Errorf`

eyakubovich:
I think it's a reasonable solution but I need to sleep on this. There are some implications for integrating it with `rkt` and it setting up a default route.

steveeJ:
I'd like to point out again that the default behaviour hasn't been changed, it does require a configuration change to make this happen. Only if `isDefaultGateway` is provided *instead of* or in addition to `isGateway` the default gateway will be set.

eyakubovich:
@steveeJ I guess you're right. However there's also fact that this almost sugar on top of IPAM routes. You could do the same with `"routes": { "dst": "0.0.0.0/0", "gw": "$GW }` in the IPAM section. At the same this maybe inconvenient but in that case it should be handled in the IPAM section/plugins and not the bridge.

steveeJ:
I see your point. I followed the already existing `isGateway`, which could also replaced via IPAM by adding a route that has the IPAM subnet as the destination.
One thing that has bothered me so far is that the gateway is automatically assumed to be the first subnet address using the `isGateway` attribute. In case the gateway is different, people would have to configure the routes manually via IPAM, and the `isGateway` would not fit in anymore.

So, I would like a solution of that moves the gateway attributes to IPAM and allows for better flexibility. 
Opinions?

eyakubovich:
The first address is only assumed if IPAM did not provide a gateway address (https://github.com/appc/cni/blob/master/plugins/main/bridge/bridge.go#L195). The "gateway" field in the IPAM result JSON is optional, so there is a problem that if IPAM returns some routes but not the gateway, it can get messed up. Maybe we can issue a warning or something in this case.

steveeJ:
@eyakubovich: my understanding on this is that we'll move the gateway attributes to IPAM and leave the behavior as is, but add a convenience option to set the default GW as already implemented in this PR.

Please confirm and I'll update the implementation accordingly.

plugins/main/bridge/bridge.go
```diff
@@ -196,6 +201,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
+	// TODO: make this optional when IPv6 is supported
```
jellonek: After inclusion of IPv6 next line should be expanded with `&& result.IP6 == nil` so this comment is hmmm... not precise.

jellonek:
LGTM

plugins/main/bridge/bridge.go
steveeJ: It's a hint that there's going to be multiple options, logically `IPv4 || IPv6`

squaremo:
LGTM. As discussed on IRC, it's worth waiting for #211 and rebasing, for the tests, before merging.

PR #58, opened 2015-09-09
ipvlan/macvlan: remove ipMasq related code

### LINKED ISSUE #14: Investigate ipMasq support from macvlan and ipvlan
I don't think it's really possible to get IP masquerade to work with macvlan as it operates at L2 and probably does not call into netfilter code. In that case, support for it should be removed. ipvlan maybe can work (maybe in one of the modes) but needs to be investigated.

steveeJ:
Luckily the docs haven't mentioned support for ipMasq for both plugins so far.
Even if anyone has attempted to enable the feature in their configuration files it didn't have the desired effect for the network.

Closes #14.


jonboulle:
LGTM

eyakubovich:
LGTM

PR #59, opened 2015-09-14
Factor an API out into a module

squaremo:
This takes some of the machinery from CNI and from the rkt networking
code, and turns it into a library that can be linked into go apps.

Included is an example command-line application that uses the library,
called `cnitool`.

Other headline changes:

 * Plugin exec'ing is factored out

The motivation here is to factor out the protocol for invoking
plugins. To that end, a generalisation of the code from
pkg/plugin/ipam.go goes into pkg/invoke/exec.go.

 * Move argument-handling and conf-loading into public API

The fact that the arguments get turned into an environment for the
plugin is incidental to the API; so, in the API we provide a way of
supplying them as a struct or saying "just use the same arguments
as I got" (the latter is for IPAM plugins).


api.go
```diff
@@ -0,0 +1,68 @@
...
+
+type RuntimeConf struct {
+	ContainerID string
+	NetNS       string
+	IfName      string
+	Args        [][2]string
+}
+
+type NetworkConfig struct {
+	types.NetConf
```
eyakubovich: I think it should be a named member instead of embedding for the pathological case of somebody putting "bytes" as a key in JSON (either by accident or b/c some plugin decides to have a field like that)

conf.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	return confFiles, nil
+}
+
+func LoadConf(dir, name string) (*NetworkConfig, error) {
+	files, err := ConfFiles(dir)
+	switch {
+	case err != nil:
+		return nil, err
+	case files == nil || len(files) == 0:
```
eyakubovich: `files == nil` implies `len(files) == 0`
http://play.golang.org/p/txZz6bZjCf

api.go
squaremo: Mmmhmm, good call, I'll change it.

eyakubovich:
I don't know if any .go files should be at the top-level. I think they should all be some place under `pkg`.

squaremo:
> I don't know if any .go files should be at the top-level. I think they should all be some place under pkg.

That can be arranged; I did like the simplicity of

    import (
      "github.com/appc/cni"
    )

Would you prefer `github.com/appc/cni/pkg/api` (and `.../conf`) though?

eyakubovich:
I do like the simplicity as well but I want to emphasize that CNI (and the repo) is first and foremost a spec. That's why SPEC.md is at the top level. The included plugins, client and helper code is secondary. From the practical point of view, the library code is more important to people. Let's compromise and put it under `github.com/appc/cni/api`

conf.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, fmt.Errorf("No net configurations found")
```
jonboulle: s/No/no/

README.md
```diff
@@ -78,6 +81,8 @@ lo        Link encap:Local Loopback
           RX bytes:0 (0.0 B)  TX bytes:0 (0.0 B)
 ```
 
+The environment variable `CNI_PATH` tells the scripts and library where to look for plugins executables.
```
jonboulle: s/plugins/plugin/

jonboulle:
IMHO it would be great for reviewability and auditability if the file renames could be split out into a separate commit from the actual changes. Not going to block this, but just sayin :-)

squaremo:
> Let's compromise and put it under github.com/appc/cni/api

Actually, how about `github.com/appc/cni/libcni`, so the package name can (per convention) be `libcni` rather than `api`.

squaremo:
> it would be great for reviewability and auditability if the file renames could be split out into a separate commit from the actual changes

Yeah, that's a fair cop. GitHub does an OK job of showing file moves combined with changes, mostly -- but other tools, not so much.

squaremo:
Comments addressed, including moving the top-level go files to `libcni/`; I'll squash them if it all looks ok.

eyakubovich:
LGTM. Please squash.

eyakubovich:
@squaremo You are now a maintainer with write access. Feel free to click the merge button!

PR #60, opened 2015-09-18
bug fix: exec of DEL cmd caused JSON decode error

eyakubovich:
When plugin is executed with a DEL command, it does not
print result to stdout unless there is an error. Therefore
it stdout bytes should not be passed to json.Unmarshal.


eyakubovich:
@squaremo PTAL

squaremo:
Good spot; LGTM by eye, and testing manually.

PR #62, opened 2015-09-18
No host local ptp

eyakubovich:


plugins/main/ptp/ptp.go
```diff
@@ -75,13 +131,17 @@ func setupHostVeth(vethName string, ipConf *types.IPConfig) error {
 	// TODO(eyakubovich): IPv6
```
steveeJ: By now we should've set the host interface up. For convenience here's something to copy&paste" 

```
    err = netlink.LinkSetUp(veth)                                                                                    
    if err != nil {                                                                                                  
        return fmt.Errorf("failed to set %q up: %v", vethName, err)                                                  
    }
```

steveeJ:
Superseded by #63

PR #63, opened 2015-09-23
Merge host-local and host-local-ptp

steveeJ:
To have an atomic PR, this picks up #62 and adds a fix for the netlink behaviour


crawford:
_Minor_ NITs. Otherwise, looks good.

PR #69, opened 2015-09-29
Route del would fail due to wrong scope; better errors

eyakubovich:
In some cases the route deletion would fail due to wrong
scope. It should be NOWHERE when deleting (per iproute2).
This also adds more verbose error messages.


steveeJ:
LGTM

PR #73, opened 2015-10-07
flannel: doc fix and check for missing subnet.env fields

### LINKED ISSUE #72: Missing FLANNEL_NETWORK in flannel documentation.
In https://github.com/appc/cni/blob/master/Documentation/flannel.md example `subnet.env` does not have `FLANNEL_NETWORK`  variable, without which https://github.com/appc/cni/blob/master/plugins/meta/flannel/flannel.go#L205 will panic.

Btw. there is no error checking if required variables are in `subnet.env` file.
If this variables could be missing, then in https://github.com/appc/cni/blob/master/plugins/meta/flannel/flannel.go#L72 pointer struct members should be initialized with empty values like:
```
se := &subnetEnv{
  nw: &net.IPNet{},
  sn: &net.IPNet{},
}
```

eyakubovich:
- Document that flannel outputs FLANNEL_NETWORK b/c we use it
- Check and error out if expecited fields are missing in subnetEnv file

Fixes #72

### DIFF
--- a/plugins/meta/flannel/flannel.go
+++ b/plugins/meta/flannel/flannel.go
@@ -48,8 +48,26 @@ type NetConf struct {
 type subnetEnv struct {
 	nw     *net.IPNet
 	sn     *net.IPNet
-	mtu    uint
-	ipmasq bool
+	mtu    *uint
+	ipmasq *bool
+}
+
+func (se *subnetEnv) missing() string {
+	m := []string{}
+
+	if se.nw == nil {
+		m = append(m, "FLANNEL_NETWORK")
+	}
+	if se.sn == nil {
+		m = append(m, "FLANNEL_SUBNET")
+	}
+	if se.mtu == nil {
+		m = append(m, "FLANNEL_MTU")
+	}
+	if se.ipmasq == nil {
+		m = append(m, "FLANNEL_IPMASQ")
+	}
+	return strings.Join(m, ", ")
 }
 
 func loadFlannelNetConf(bytes []byte) (*NetConf, error) {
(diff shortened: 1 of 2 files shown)


plugins/meta/flannel/flannel.go
```diff
@@ -92,16 +108,22 @@ func loadFlannelSubnetEnv(fn string) (*subnetEnv, error) {
 			if err != nil {
 				return nil, err
 			}
-			se.mtu = uint(mtu)
+			se.mtu = new(uint)
```
jonboulle: what's wrong with se.mtu = &mtu?

plugins/meta/flannel/flannel.go
```diff
@@ -92,16 +108,22 @@ func loadFlannelSubnetEnv(fn string) (*subnetEnv, error) {
...
+			ipmasq := parts[1] == "true"
+			se.ipmasq = &ipmasq
 		}
 	}
 	if err := s.Err(); err != nil {
 		return nil, err
 	}
 
+	if m := se.missing(); m != "" {
+		return nil, fmt.Errorf("%q is missing %q", fn, m)
```
jonboulle: you don't want to be nice and print all the missing ones? ;-)

plugins/meta/flannel/flannel.go
eyakubovich: you'll get a type mismatch. `mtu` is `uint64` and `se.mtu` is `*uint`. There are other ways to write this code, of course.

plugins/meta/flannel/flannel.go
eyakubovich: fine, since you ask :)

plugins/meta/flannel/flannel.go
jonboulle: oh ok clearly I didn't read the spec of ParseUint. *cringe* sometimes I miss Java's extreme type verbosity..

eyakubovich:
@jonboulle PTAL

plugins/meta/flannel/flannel.go
```diff
@@ -182,7 +206,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 
 	if !hasKey(n.Delegate, "ipMasq") {
 		// if flannel is not doing ipmasq, we should
-		ipmasq := !fenv.ipmasq
+		ipmasq := !*fenv.ipmasq
```
jonboulle: did I miss this change in the previous review? if not, did it actually compile!?

jonboulle:
LGTM although I am slightly concerned about the above

plugins/meta/flannel/flannel.go
eyakubovich: I made all fields pointers so I can use `nil` as not set.

jonboulle: I got that, I just didn't notice you change this line in the original
commit.

PR #76, opened 2015-11-04
spec: add "dns" field to the result JSON

### LINKED ISSUE #67: passing dns resolvers out of a plugin
We need to define the where in the returned JSON metadata a plugin can tell the caller what the DNS configuration should be. This will essentially be a priority ordered list of IP addresses:

```
"dns-resolvers": ["8.8.8.8", "8.8.4.4"]
```

eyakubovich:
This adds an ability for networks to communicate
about the DNS nameservers they know about to the
container runtime.

Fixes #67

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -2,7 +2,8 @@
 
 ## Overview
 
-This document proposes a generic plugin-based networking solution for application containers on Linux, the _Container Networking Interface_, or _CNI_. It is derived from the [rkt Networking Proposal][rkt-networking-proposal], which aimed to satisfy many of the [design considerations][rkt-networking-design] for networking in [rkt][rkt-github].
+This document proposes a generic plugin-based networking solution for application containers on Linux, the _Container Networking Interface_, or _CNI_.
+It is derived from the [rkt Networking Proposal][rkt-networking-proposal], which aimed to satisfy many of the [design considerations][rkt-networking-design] for networking in [rkt][rkt-github].
 
 For the purposes of this proposal, we define two terms very specifically:
 - _container_ can be considered synonymous with a [Linux _network namespace_][namespaces]. What unit this corresponds to depends on a particular container runtime implementation: for example, in implementations of the [App Container Spec][appc-github] like rkt, each _pod_ runs in a unique network namespace. In [Docker][docker], on the other hand, network namespaces generally exist for each separate Docker container.
@@ -48,6 +49,7 @@ The operations that the CNI plugin needs to support are:
     - **Name of the interface inside the container**. This is the name that should be assigned to the interface created inside the container (network namespace); consequently it must comply with the standard Linux restrictions on interface names.
   - Result:
     - **IPs assigned to the interface**. This is either an IPv4 address, an IPv6 address, or both.
+    - **List of DNS nameservers**. This is a priority-ordered list of IPv4 and IPv6 addresses of DNS nameservers.
 
 - Delete container from network
   - Parameters:
@@ -83,9 +85,16 @@ Success is indicated by a return code of zero and the following JSON printed to 
     "ip": <ipv6-and-subnet-in-CIDR>,
     "gateway": <ipv6-of-the-gateway>,  (optional)
     "routes": <list-of-ipv6-routes>    (optional)
-  }
+  },
+  "dns": <list-of-DNS-nameservers>     (optional)
 }
 ```
+
+The "dns" field contains a list of a priority-ordered list of DNS nameservers that this network is aware of.
+Each entry in the list is a string containing either an IPv4 or an IPv6 address.
+Typically this value would just be the value returned by the IPAM plugin.
+It is outside the scope of this specification how the container runtime uses the list of DNS nameservers from each of the networks to provide name resolution services to the container.
+Examples of how this list could be used include generating an `/etc/resolv.conf` file to be injected into the container filesystem or running a DNS forwarder on the host.
 
 Errors are indicated by a non-zero return code and the following JSON being printed to stdout:
 ```
@@ -127,7 +136,7 @@ The network configuration is described in JSON form. The configuration can be st
     // ipam specific
     "subnet": "10.1.0.0/16",
     "gateway": "10.1.0.1"
-  },
+  }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


steveeJ:
Any particular reason why DNS servers should not be differentiated between IP versions?

philips:
@steveeJ is that common to differentiate by protocol? Can resolv.conf do that?

squaremo:
I can think of three appropriate ways of providing nameservers:

    * entirely out-of-band with respect to CNI (e.g., if Kubernetes runs its own DNS server);
    * in the configuration of a network;
    * supplied by an IPAM plugin.

Allowing that I don't always have the right end of the stick, I would be interested to know what suggests this change, since it doesn't (quite) cover off the above situations.

eyakubovich:
@steveeJ That is a great point and I spent a while thinking about whether to add `dns` at top level or inside the `ip4`, `ip6` sections. The reason I decided to go with top level is to allow the plugin to specify priority/preference between IPv4 and IPv6 nameservers in dual stack scenarios. However, it's a small win so definitely open to doing it the other way if there's a reason.

eyakubovich:
@squaremo I'm not sure I understand your concern. There is no right way to do DNS resolution. This change just allows the CNI plugin to communicate about the nameservers it happened to know about (e.g. b/c it learned it via DHCP). It would still be up to the runtime to choose the implementation.

eyakubovich:
@squaremo I see what you're saying. The IPAM section in the spec is missing the `dns` field. It will ultimately be coming from there and propagated up. I will fix.

steveeJ:
@philips resolv.conf doesn't do it but I wouldn't only solve the resolv.conf use-case in CNI. We provide routes/gateway per protocol, applying that on DNS would be consistent and nice to have for applications that could make use with the extra information.

SPEC.md
```diff
@@ -48,6 +49,7 @@ The operations that the CNI plugin needs to support are:
     - **Name of the interface inside the container**. This is the name that should be assigned to the interface created inside the container (network namespace); consequently it must comply with the standard Linux restrictions on interface names.
   - Result:
     - **IPs assigned to the interface**. This is either an IPv4 address, an IPv6 address, or both.
+    - **List of DNS nameservers**. This is priority ordered list of IPv4 and IPv6 addresses of DNS nameservers.
```
jonboulle: s/priority ordered/a priority-ordered/

SPEC.md
```diff
@@ -83,10 +85,18 @@ Success is indicated by a return code of zero and the following JSON printed to
     "ip": <ipv6-and-subnet-in-CIDR>,
     "gateway": <ipv6-of-the-gateway>,  (optional)
     "routes": <list-of-ipv6-routes>    (optional)
-  }
+  },
+  "dns": <list-of-DNS-nameservers>     (optional)
 }
 ```
 
+The "dns" field contains a list of priority ordered list of DNS nameservers that this network is aware of.
```
jonboulle: s/a list of priority ordered list of/ a priority-ordered list of/

SPEC.md
```diff
@@ -83,10 +85,18 @@ Success is indicated by a return code of zero and the following JSON printed to
...
+Each entry in the list is a string containing either an IPv4 or an IPv6 address.
+Typically this value would just be the value returned by the IPAM plugin.
+It is outside the scope of this specification on how the container runtime uses the list of DNS nameservers from each of the networks to create a final list for the container.
+In addition, it is outside the scope of this specification how the final list is used to provide name resolution services to the container.
```
jonboulle: doesn't this imply line-96 anyway? maybe we could just have one

SPEC.md
```diff
@@ -193,6 +204,10 @@ Each route entry is a dictionary with the following fields:
 - `dst` (string): Destination subnet specified in CIDR notation.
 - `gw` (string): IP of the gateway. If omitted, a default gateway is assumed (as determined by the CNI plugin).
 
+The "dns" field contains a list of priority ordered list of DNS nameservers that this network is aware of.
```
jonboulle: s/a list of priority ordered list of/ a priority-ordered list of/

jonboulle:
Few minor syntactical issues, but other than that this approach lgtm.

SPEC.md
dcbw: > doesn't this imply line-96 anyway? maybe we could just have one

Yeah, how about:

"It is outside the scope of this specification how the container runtime uses the list of DNS nameservers from each of the networks to provide name resolution services to the container."

dcbw:
The only scenario I can think of that this wouldn't work for is split DNS, where queries for certain domains are directed to specific nameservers and the rest go somewhere else.  eg, VPN-like functionality.  That would require that the 'dns' option be a map of string::list-of-string though.  And maybe too complicated for now, since if this actually becomes a need we could add a 'split-dns' key or something. Just food for thought.

But generally LGTM too.  There is value in having the list priority ordered, which implies keeping it separate and not in the IP-version-specific blocks.

dcbw:
@rajatchopra look good to you too?

rajatchopra:
Useful and good to go I think.

One question which should not block this PR is handling of /etc/hosts, if at all. Its mention is removed from Open Questions, but this PR does not address it.

eyakubovich:
@rajatchopra This PR takes a stance that `/etc/hosts`, like `/etc/resolve.conf`, is outside the scope of the spec (at least for now). The only thing this spec covers in terms of name resolution is the `dns` key as specified in this PR.

eyakubovich:
@dcbw I think with split-dns, it's the DNS server that serves up alternate IPs for a some DNS zone. It's usually not done by the local resolver (e.g. `/etc/resolv.conf`) so I think it's reasonable not to support it here as well. But like you say, we can extend the spec if the need arises.

dcbw:
> I think with split-dns, it's the DNS server that serves up alternate IPs for a some DNS zone.

@eyakubovich Not really.  It comes into play when you have multiple sources of network information.  For example, suppose you want your container to connect to multiple networks; one for the local cluster and a VPN-like one.  Each network is handled by a different CNI plugin (or the same plugin, doesn't matter really), and each plugin can return whatever DNS information it wants.  Which do you choose to be "primary"?

Even though priority and how DNS resolution is handled is out of scope, the thing that calls the CNI plugin needs some way to figure that out.  And passing the search domains back from the CNI plugin is pretty crucial for split DNS, since that's the only way you know what domains to send to that network's DNS server.

I'm not saying we need to add it now...  Just that we should think about the use-case as more and more stuff becomes multi-network capable.

PR #79, opened 2015-11-16
spec: add ability for plugins to return state to persist

eyakubovich:
Currently any state that the plugin needs to maintain has
to be serialized to disk by the plugin itself. To reduce
this burden, the current spec provides the DEL command
all the same information as the ADD command. This can be
tricky for the runtime to achieve, requiring it to make
a copy of network configuration. Since it has to do the
work of writing to disk, this change allows the plugin to
return extra state for it to persist. The DEL command
interface has been changed to pipe in previously returned
state. Therefore the DEL command no longer needs to receive
all the information that was provided to ADD as the plugin
can easily save it via the "state" mechanism.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -33,69 +33,35 @@ Upon completion of the container lifecycle, the runtime executes the plugins in 
 Each CNI plugin is implemented as an executable that is invoked by the container management system (e.g. rkt or Docker).
 
 A CNI plugin is responsible for inserting a network interface into the container network namespace (e.g. one end of a veth pair) and making any necessary changes on the host (e.g. attaching other end of veth into a bridge).
-It should then assign the IP to the interface and setup the routes consistent with IP Address Management section by invoking appropriate IPAM plugin.
-
-### Parameters
+It should then assign the IP to the interface and set up the routes consistent with IP Address Management section by invoking appropriate IPAM plugin.
 
 The operations that the CNI plugin needs to support are:
+- Adding a container to the network
+- Deleting a container from the network
 
+### ADD command
 
-- Add container to network
-  - Parameters:
-    - **Container ID**. This is optional but recommended, and should be unique across an administrative domain while the container is live (it may be reused in the future). For example, an environment with an IPAM system may require that each container is allocated a unique ID and that each IP allocation can thus be correlated back to a particular container. As another example, in appc implementations this would be the _pod ID_.
-    - **Network namespace path**. This represents the path to the network namespace to be added, i.e. /proc/[pid]/ns/net or a bind-mount/link to it.
-    - **Network configuration**. This is a JSON document describing a network to which a container can be joined. The schema is described below.
-    - **Extra arguments**. This allows granular configuration of CNI plugins on a per-container basis.
-    - **Name of the interface inside the container**. This is the name that should be assigned to the interface created inside the container (network namespace); consequently it must comply with the standard Linux restrictions on interface names.
-  - Result:
-    - **IPs assigned to the interface**. This is either an IPv4 address, an IPv6 address, or both.
-    - **List of DNS nameservers**. This is a priority-ordered list of IPv4 and IPv6 addresses of DNS nameservers.
+The ADD command is executed to add (join) the container to a network.
+It is typically done as part of the container creation process.
 
-- Delete container from network
-  - Parameters:
-    - **Container ID**, as defined above.
-    - **Network namespace path**, as defined above.
-    - **Network configuration**, as defined above.
-    - **Extra arguments**, as defined above.
-    - **Name of the interface inside the container**, as defined above.
+#### Parameters passed via environment variables
+- `CNI_COMMAND`: `ADD`
+- `CNI_CONTAINERID`: Container ID. This is optional but recommended, and should be unique across an administrative domain while the container is live (it may be reused in the future). For example, an environment with an IPAM system may require that each container is allocated a unique ID and that each IP allocation can thus be correlated back to a particular container. As another example, in appc implementations this would be the _pod ID_.
+- `CNI_NETNS`: Path to network namespace file representing network namespace to which the container is to be added, i.e. /proc/[pid]/ns/net or a bind-mount/link to it.
+- `CNI_IFNAME`: Interface name to set up. This is the name that should be assigned to the interface created inside the container (network namespace); consequently it must comply with the standard Linux restrictions on interface names.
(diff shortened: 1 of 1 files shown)


dcbw:
Looks OK to me; looking at Kube we'll have to store the state somewhere, possibly in pkg/kubelet/container/runtime.go's Pod struct via some network plugin data object there.  Later I'd hope we could stick the data into whatever enhanced Network Plugin API we come up with, or simply require the enhanced network plugin to store the data itself, indexed by Pod.ID or such.

thockin:
SGTM

SPEC.md
```diff
@@ -106,10 +72,29 @@ Errors are indicated by a non-zero return code and the following JSON being prin
...
+It is typically done as part of the container teardown.
+
+#### Parameters passed via environment variables
+- `CNI_COMMAND`: `DEL`
+- `CNI_CONTAINERID`: Container ID. This must be the same Container ID that was passed in during the ADD command. If no Container ID was passed into the ADD command, no ID must be passed in during the DEL command.
+- `CNI_NETNS`: Path to network namespace file representing network namespace from which the container is to be deleted, i.e. `/proc/[pid]/ns/net` or a bind-mount/link to it.
+- `CNI_PATH`: Colon-separated list of paths to search for CNI plugin executables.
+
+#### State streamed via standard input
+This is the JSON value that was returned in the `state` field of the successful result JSON document during the ADD command.
```
jonboulle: To make this valid JSON, it should actually be an object:
`{"state": <value>}`

SPEC.md
```diff
@@ -33,69 +33,35 @@ Upon completion of the container lifecycle, the runtime executes the plugins in
...
-
-The "dns" field contains a list of a priority-ordered list of DNS nameservers that this network is aware of.
-Each entry in the list is a string containing either an IPv4 or an IPv6 address.
-Typically this value would just be the value returned by the IPAM plugin.
-It is outside the scope of this specification how the container runtime uses the list of DNS nameservers from each of the networks to provide name resolution services to the container.
-Examples of how this list could be used include generating an `/etc/resolv.conf` file to be injected into the container filesystem or running a DNS forwarder on the host.
+#### Result streamed via standard output
+Success is indicated by an exit code of zero and a JSON result printed to stdout.
+This should be the same output as was returned by the IPAM plugin (see [IP Allocation](#ip-allocation) for details) with the exception of `state` field.
+`state` field should contain both the state that the plugin itself needs the runtime to persist as well as any `state` returned by the IPAM plugin.
```
jonboulle: So to be clear, the plugin decides how to amalgamate its own state and any IPAM plugin state? And then is responsible for splitting them apart again as necessary? I'm wondering if there's any reason we would want to formalise this separately instead (e.g. `{"state": <val>, "ipamState": <val>}` .. but I can't actually think of a use case right now)

rajatchopra:
Can we have an example of the state? Also it seems odd that a state that is created by the plugin is asked to be maintained by runtime. In kubelet itself there will be trouble if we have to persist something on disk.

dcbw:
Also, given that the CNI plugin might further exec an IPAM plugin, how would the IPAM plugin persist state too?  I suppose the main plugin could fold that state into its opaque state string and unpack it on DEL, but maybe the states should be separate to simplify plugins?

jonboulle:
+1 on a state example.

@dcbw does my suggestion of the structured object mitigate your concern? The caller just has to persist one state, but it's merged in a well-defined (CNI-prescribed) way (and trivial for plugins to split apart)

@rajatchopra a few weeks ago several of us met to discuss k8s/cni integration, and we concluded it is OK for kubelet to persist this state.

SPEC.md
eyakubovich: JSON document/text can be an object, array, number, string, bool, or null (http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf). So any value that was in `state` field is itself a valid JSON doc.

SPEC.md
eyakubovich: You're correct that it's up to the plugin to amalgamate. For example, if the plugin does not want to keep any state, it could just pass IPAM's state up as its own.

As for specifying it rules or conventions on how to combine state is an option. If we go down that route, I'd like to keep it generic as I would also like to support more arbitrary plugin composability. For example, one plugin might call onto 2 helpers. That's why I decided not to prescribe this.

eyakubovich:
I'm not sure why having the outer plugin do the combining and splitting is such a chore. It's as easy as assigning value of `state` field that it received in result JSON into some field in it's result JSON.

SPEC.md
jonboulle: Hmm. My interpretation is a little different, from the RFC:
https://tools.ietf.org/html/rfc4627#section-2


On Wed, Nov 18, 2015 at 3:57 PM, Eugene Yakubovich <notifications@github.com

jonboulle: specifically

>   A JSON text is a serialized object or array.


On Wed, Nov 18, 2015 at 4:29 PM, Jonathan Boulle <jonathan.boulle@coreos.com
> wrote:

> Hmm. My interpretation is a little different, from the RFC:
> https://tools.ietf.org/html/rfc4627#section-2
>
>
> On Wed, Nov 18, 2015 at 3:57 PM, Eugene Yakubovich <

jonboulle: anyway putting aside the spec itself (since your reference presumably supersedes mine), I'd still argue in favour of the slightly more structured approach.

eyakubovich: We can wrap it in an object if we expect to add other stuff to it down the road. Otherwise it's just a dictionary with a single element for no reason.

eyakubovich:
@rajatchopra Example of the state would be:
- any values from network config that are needed for the teardown. The DEL command no longer receives the config.
- Value of `CNI_IFNAME` that was passed in ADD command
- Some ID of an iptables rule that was added and needs to be removed (maybe ID stored in iptables rule comment).

dcbw:
> Some ID of an iptables rule that was added and needs to be removed (maybe ID stored in iptables rule comment).

Or for the OpenShift case, we could generate a randomized cookie when adding pod-specific OVS rules and return that cookie, and use it to delete those rules on DEL.

dcbw:
> I'm not sure why having the outer plugin do the combining and splitting is such a chore. It's as easy as assigning value of state field that it received in result JSON into some field in it's result JSON.

It's not hard, just that it would duplicate code and that seems kinda pointless.  But if as you say the state could be an amalgamation of many plugins, not just this plugin and the IPAM plugin, I see your point.  So whatever way that combination/split happens is OK with me.

squeed:
I think we don't want this. I'm going to close it; feel free to re-open if this offends you.

PR #80, opened 2015-11-17
spec: add cniVersion to results and config

eyakubovich:
Version all artifacts using semantic versioning
so that plugins and container runtimes coded against
different versions of the spec can interoperate.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -42,6 +42,7 @@ The operations that the CNI plugin needs to support are:
 
 - Add container to network
   - Parameters:
+    - **Version**. The version of CNI spec that the caller is using (container management system or the invoking plugin).
     - **Container ID**. This is optional but recommended, and should be unique across an administrative domain while the container is live (it may be reused in the future). For example, an environment with an IPAM system may require that each container is allocated a unique ID and that each IP allocation can thus be correlated back to a particular container. As another example, in appc implementations this would be the _pod ID_.
     - **Network namespace path**. This represents the path to the network namespace to be added, i.e. /proc/[pid]/ns/net or a bind-mount/link to it.
     - **Network configuration**. This is a JSON document describing a network to which a container can be joined. The schema is described below.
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -42,6 +42,7 @@ The operations that the CNI plugin needs to support are:
 
 - Add container to network
   - Parameters:
+    - **Version**. The version of CNI spec that the caller is using (container managment system or the invoking plugin).
```
jonboulle: management

SPEC.md
```diff
@@ -60,6 +61,7 @@ The operations that the CNI plugin needs to support are:
     - **Name of the interface inside the container**, as defined above.
 
 The executable command-line API uses the type of network (see [Network Configuration](#network-configuration) below) as the name of the executable to invoke. It will then look for this executable in a list of predefined directories. Once found, it will invoke the executable using the following environment variables for argument passing:
+- `CNI_VERSION`:  [Semantic Version](http://semver.org) of CNI specification. This effectively versions the CNI_XXX environment variables.
```
jonboulle: please specify semver 2.0.0

SPEC.md
```diff
@@ -76,6 +78,7 @@ Success is indicated by a return code of zero and the following JSON printed to
 
 ```
 {
+  "cniVersion": "0.1",
```
jonboulle: 0.1.0

dcbw:
I think we need a Version parameter for the DEL action too.  If we version one of the calls, we should version all of them.

Since CNI plugins are executables and not shared libraries loaded at the start of the management entity, the plugin could be upgraded through rpm/dpkg/whatever while a container is running.  Yeah, Don't Do That.  But the plugin should be able to return an intelligent error in that case too.

Next question is about backwards compatibility; in your model it looks like the plugin itself is responsible for determining whether the version passed to it is OK or not.  What if a CMS wanted to support two versions of plugins for backwards compat?  This could be achieved by versioning the variable names so that the call would be compatible with multiple plugins.  Otherwise, the CMS would have to choose between (a) having a flag day for older plugins or (b) having to introspect the plugin's version through some out-of-band mechanism and pick which API level to use.

eyakubovich:
@dcbw You're right, we absolutely need version in DEL action. I missed putting it there -- great catch.

Also good point on backward compatibility stuff. Versioning env variables (on major version) is very reasonable. I would like to see what others think about this. /cc @thockin -- I know you wanted CNI versioning.

thockin:
I'm a fan of versioning

SPEC.md
```diff
@@ -216,3 +229,6 @@ IPAM plugin examples:
 #### Notes
  - Routes are expected to be added with a 0 metric.
  - A default route may be specified via "0.0.0.0/0". Since another network might have already configured the default route, the CNI plugin should be prepared to skip over its default route definition.
+
+## Well-known Error Codes
+- `1` - Incompatabile CNI version
```
jonboulle: incompatible

jonboulle: are we going to define incompatibility? e.g. appc has this to say (not saying it's necessarily exactly what we want but to give you an idea):
```
The App Container specification aims to follow semantic versioning and retain forward and backwards compatibility within major versions. For example, if an implementation is compliant against version 1.0.1 of the spec, it is compatible with the complete 1.x series.

```

eyakubovich: That's what I'm saying here (https://github.com/appc/cni/pull/80/files#diff-916b3e1e005fd96e3a0546715235477dR65) just with not so many words. Having backwards compat within same major version is part of SemVer spec.

jonboulle:
@eyakubovich this LGTM but please fix the "Incompatabile" typo

eyakubovich:
@dcbw I tried making changes to the spec to version the variables: e.g. `CNI_V1_COMMAND`. Since SemVer2 says that only major versions are not backward compat, we only need to version on those. On the surface it looks OK.

It's hard to predict the future but it seems like if we make big changes that warrant next major version, it's far from a given that they can live side by side. For example, if next version starts using totally different commands, this scheme won't actually help. Even if the container runtime could find out the version of the plugin before invoking it, it's hard to say that the necessary setup that the runtime has to do will work with say, both v2 and v3 plugins.

At this point I actually think we should not try to version variables and just restrict everything to running the  same major version.

dcbw:
@eyakubovich ok, we can also punt the backwards compat discussion until we actually have a new major version.  Which should be part of the discussion at that point anyway.

PR #82, opened 2015-11-19
dhcp: detect if link is down and attempt to set up

### LINKED ISSUE #81: DHCP plugin fails with "network is down" error
Seems to be happening with both macvlan and ipvlan. Might be that the interface is DOWN when IPAM is invoked.

steveeJ:
With this we still get the "network is down" error, but after the 2nd
retry the interface is up and the dhcp request works.

Fixes #81.


eyakubovich:
LGTM. I think this is short term solution. We need to specify in the spec what plugin should bring the interface up.

PR #83, opened 2015-11-20
Send ClientID DHCP option in DISCOVER/ACK msgs

### LINKED ISSUE #17: Add support for ClientID to DHCP
With ipvlan, all interfaces will share the MAC address so the DHCP server will return the same lease. Need to add support for ClientID (DHCP option 61). The value can be ContainerID by default but over-rideable via `CNI_ARGS`.

eyakubovich:
ClientID is "containerid/netname" and can help
with visibility into allocated IPs in DHCP server.
Unforunately it still does not make it possible
to use DHCP with ipvlan as the messages coming
from the DHCP server can't be forwarded to the
correct ipvlan interface since they don't yet
have the IP assigned.

Fixes #17

### DIFF
--- a/Documentation/dhcp.md
+++ b/Documentation/dhcp.md
@@ -33,3 +33,5 @@ With the daemon running, containers using the dhcp plugin can be launched.
 ## Network configuration reference
 
 * `type` (string, required): "dhcp"
+* `sendClientID` (boolean, optional): whether to include the Client Identifier option (61) in messages to the DHCP server. Defaults to true.
+* `broadcast` (boolean, optional): whether to set the Broacast flag in DHCP messages. Defaults to false.
--- a/Documentation/ipvlan.md
+++ b/Documentation/ipvlan.md
@@ -38,3 +38,31 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
 Therefore the container will not be able to reach the host via `ipvlan` interface.
 Be sure to also have container join a network that provides connectivity to the host (e.g. `ptp`).
 * A single master interface can not be enslaved by both `macvlan` and `ipvlan`.
+
+### Using ipvlan and dhcp
+There are two important considerations when using the "ipvlan" plugin together with the "dhcp" IPAM plugin.
+
+First, the kernel ipvlan driver uses the IP packet's destination address when choosing the slave interface to which it should be forwarded.
+The driver cannot perform this forwarding if the container interface has not yet been assigned an IP address.
+Since the DHCP packets are exchanged to determine which IP should be used, the interface will not yet have an IP address.
+DHCP servers typically unicast replies back to the client.
+The ipvlan driver will drop these replies.
+To work around this, the DHCP client must instruct the DHCP server to use broadcast replies instead.
+This can be accomplished by setting the "broadcast" flag in the "ipam" section of the network configuration:
+
+```
+{
+	"name": "mynet",
+	"type": "ipvlan",
+	"master": "eth0",
+	"ipam": {
+		"type": "dhcp",
+		"broadcast": true
+	}
+}
+```
+
+Second, because all ipvlan interfaces share the MAC address of the master interface, DHCP servers will have trouble identifying and tracking individual container interfaces.
+To mitigate this, it is necessary to include DHCP [option 61](http://tools.ietf.org/html/rfc2132#section-9.14) (Client Identifier) in DHCP messages.
+The DHCP plugin will emit such client identifiers if the sendClientID flag is set, as it is by default.
+The value of this Client Identifier option will be ContainerID catenated with the network name.
--- a/plugins/ipam/dhcp/daemon.go
+++ b/plugins/ipam/dhcp/daemon.go
@@ -37,6 +37,14 @@ const resendCount = 3
 
 var errNoMoreTries = errors.New("no more tries")
 
+type NetConf struct {
+	Name string `json:"name"`
+	IPAM struct {
+		SendClientID bool `json:"sendClientID"`
+		Broadcast    bool `json:"broadcast"`
+	} `json:"ipam"`
+}
+
 type DHCP struct {
 	mux    sync.Mutex
 	leases map[string]*DHCPLease
@@ -51,13 +59,15 @@ func newDHCP() *DHCP {
 // Allocate acquires an IP from a DHCP server for a specified container.
 // The acquired lease will be maintained until Release() is called.
 func (d *DHCP) Allocate(args *skel.CmdArgs, result *types.Result) error {
-	conf := types.NetConf{}
+	conf := NetConf{}
+	conf.IPAM.SendClientID = true
+
 	if err := json.Unmarshal(args.StdinData, &conf); err != nil {
 		return fmt.Errorf("error parsing netconf: %v", err)
 	}
 
 	clientID := args.ContainerID + "/" + conf.Name
-	l, err := AcquireLease(clientID, args.Netns, args.IfName)
+	l, err := AcquireLease(clientID, conf.IPAM.SendClientID, conf.IPAM.Broadcast, args.Netns, args.IfName)
 	if err != nil {
 		return err
 	}
--- a/plugins/ipam/dhcp/lease.go
+++ b/plugins/ipam/dhcp/lease.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
 	"log"
 	"math/rand"
@@ -35,6 +36,7 @@ import (
 // and randomized to +/- 1sec
 const resendDelay0 = 4 * time.Second
 const resendDelayMax = 32 * time.Second
+const recvTimeout = 5 * time.Second
 
 const (
 	leaseStateBound = iota
@@ -50,6 +52,8 @@ const (
 
 type DHCPLease struct {
 	clientID      string
+	sendClientID  bool
+	useBroadcast  bool
 	ack           *dhcp4.Packet
 	opts          dhcp4.Options
 	link          netlink.Link
@@ -63,11 +67,13 @@ type DHCPLease struct {
 // AcquireLease gets an DHCP lease and then maintains it in the background
 // by periodically renewing it. The acquired lease can be released by
 // calling DHCPLease.Stop()
-func AcquireLease(clientID, netns, ifName string) (*DHCPLease, error) {
+func AcquireLease(clientID string, sendClientID, bcast bool, netns, ifName string) (*DHCPLease, error) {
 	errCh := make(chan error, 1)
 	l := &DHCPLease{
-		clientID: clientID,
-		stop:     make(chan struct{}),
+		clientID:     clientID,
+		sendClientID: sendClientID,
+		useBroadcast: bcast,
+		stop:         make(chan struct{}),
 	}
 
 	log.Printf("%v: acquiring lease", clientID)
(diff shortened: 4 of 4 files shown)


dcbw:
You need to make sure to set the "broadcast" flag for the DHCP requests.  Then the request will get seen by all the ipvlan interfaces, and the one that made the request will find the correct response by ClientID.

eyakubovich:
@dcbw Thanks, I'll give it a try.

Documentation/ipvlan.md
```diff
@@ -38,3 +38,40 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
 Therefore the container will not be able to reach the host via `ipvlan` interface.
 Be sure to also have container join a network that provides connectivity to the host (e.g. `ptp`).
 * A single master interface can not be enslaved by both `macvlan` and `ipvlan`.
+
+### Using ipvlan and dhcp
+There are two considerations when using "ipvlan" plugin together with "dhcp" IPAM plugin.
```
jonboulle: There are two important considerations when using the "ipvlan" plugin together with the "dhcp" IPAM plugin.

Documentation/ipvlan.md
```diff
@@ -38,3 +38,40 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
...
+
+First, since ipvlan driver uses the destination address in the IP packet when deciding to which slave the packet should be forwarded to, it will not work if the container interface is not yet assigned an IP.
```
jonboulle: First, since the kernel ipvlan driver uses
s/slave/slave interface/
s/forwarded to/forward/

Documentation/ipvlan.md
```diff
@@ -38,3 +38,40 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
...
+	"master": "eth0",
+	"ipam": {
+		"type": "dhcp",
+		"broadcast": true
+	}
+}
+```
+
+Second, because all ipvlan interfaces share the MAC address of the master interface, DHCP servers will have trouble identifying and tracking individual container interfaces.
+To combat this, it is necessary to include DHCP option 61 (Client Identifier) in DHCP messages.
```
jonboulle: s/combat/mitigate/ or "work around"
Perhaps link to RFC? https://tools.ietf.org/html/draft-henry-DHCP-opt61-UUID-type-00

Documentation/ipvlan.md
```diff
@@ -38,3 +38,40 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
...
+DHCP plugin will do so if "sendClientID" flag is enabled (default).
```
jonboulle: The DHCP plugin will do so if the "sendClientID" flag

Documentation/dhcp.md
```diff
@@ -33,3 +33,5 @@ With the daemon running, containers using the dhcp plugin can be launched.
 ## Network configuration reference
 
 * `type` (string, required): "dhcp"
+* `sendClientID` (boolean, optional): whether to include client identifier option (61) in messages to the DHCP server. Defaults to true.
+* `broadcast` (boolean, optional): whether to set Broacast flag in DHCP messages. Defaults to false.
```
joshix: s/Broacast/Broadcast/ (also: capitalization neccessary?)

dcbw:
I looked into how ISC dhclient does receive matching since I was a bit surprised that it's necessary to reflect the client ID back in the server config.  dhclient matches on the combination of the 32-bit "transaction id" from the DHCP header and the interface hardware address:

    void dhcpoffer (packet)
    	struct packet *packet;
    {
    ...
	/* Find a client state that matches the xid... */
	for (client = ip -> client; client; client = client -> next)
		if (client -> xid == packet -> raw -> xid)
			break;

    	if (!client ||
    	    client -> state != S_SELECTING ||
    	    (packet -> interface -> hw_address.hlen - 1 !=
    	     packet -> raw -> hlen) ||
    	    (memcmp (&packet -> interface -> hw_address.hbuf [1],
    		     packet -> raw -> chaddr, packet -> raw -> hlen))) {

So I was originally wrong when I said that the DHCP client should look at the client ID to determine whether it should claim the response.  Instead, the DHCP client should look at the combination of the XID and the interface hardware address.  Obviously the XID should be randomized so that no two instances using the same hardware address use the same XID.  Doing this in the DHCP client should remove the need to modify server configuration.

Where the ClientID comes into play is mostly on the server-side, since without a ClientID the server has nothing to key off for different instances that share the same MAC address.  I've had some servers (Cisco SG300-10 small-business switch, for example) hand back the same lease for two ipvlan clients when they didn't send a ClientID.  We just can't trust servers to do the right thing here without a ClientID.

Documentation/ipvlan.md
joshix: First, since the kernel ipvlan driver uses the IP packet's destination address when choosing the slave interface to which it should be forwarded, the driver cannot perform this forwarding if the container interface has not yet been assigned an IP address.

Documentation/ipvlan.md
```diff
@@ -38,3 +38,40 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
 Therefore the container will not be able to reach the host via `ipvlan` interface.
 Be sure to also have container join a network that provides connectivity to the host (e.g. `ptp`).
 * A single master interface can not be enslaved by both `macvlan` and `ipvlan`.
+
+### Using ipvlan and dhcp
+There are two considerations when using "ipvlan" plugin together with "dhcp" IPAM plugin.
+
+First, since ipvlan driver uses the destination address in the IP packet when deciding to which slave the packet should be forwarded to, it will not work if the container interface is not yet assigned an IP.
+Since the DHCP packets are exchanged to determine which IP should be used, the interface will not yet have an IP address.
+A DHCP server typically unicasts the replies back to the client which will cause the ipvlan driver to drop the replies.
```
joshix: DHCP servers typically unicast replies back to the client. The ipvlan driver will drop these replies.

Documentation/ipvlan.md
```diff
@@ -38,3 +38,40 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
...
+To work around this, the DHCP client needs to instruct the DHCP server to use broadcasts instead.
```
joshix: To work around this, the DHCP client must instruct the DHCP server to use broadcast replies instead.

Documentation/dhcp.md
```diff
@@ -33,3 +33,5 @@ With the daemon running, containers using the dhcp plugin can be launched.
 ## Network configuration reference
 
 * `type` (string, required): "dhcp"
+* `sendClientID` (boolean, optional): whether to include client identifier option (61) in messages to the DHCP server. Defaults to true.
```
joshix: "client identifier" is Capitalized in other file(s) below. Make consistently Capital Phrases for these DHCP Terms of Art.

joshix: "...the Client Identifier option (`61`)..."

Documentation/ipvlan.md
joshix: The DHCP plugin will emit such client identifiers if the `sendClientID` flag is set, as it is by default.

Documentation/ipvlan.md
```diff
@@ -38,3 +38,40 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
...
+		"type": "dhcp",
+		"broadcast": true
+	}
+}
+```
+
+Second, because all ipvlan interfaces share the MAC address of the master interface, DHCP servers will have trouble identifying and tracking individual container interfaces.
+To combat this, it is necessary to include DHCP option 61 (Client Identifier) in DHCP messages.
+DHCP plugin will do so if "sendClientID" flag is enabled (default).
+The value of Client Identifier option will be ContainerID plus the network name.
```
joshix: The value of this Client Identifier option will be the ContainerID catenated with the network name.

Documentation/ipvlan.md
```diff
@@ -38,3 +38,40 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
...
+In addition, the DHCP server needs to echo option 61 (Client Identifier) in its replies.
```
joshix: The DHCP server must echo the respective option 61 Client Identifier in its replies.

Documentation/ipvlan.md
```diff
@@ -38,3 +38,40 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
...
+Not all DHCP servers are capable of doing so while others need to be explicitly configured to enable this behavior.
```
joshix: Not all DHCP servers are capable of properly echoing the Client Identifier, while others may need to be explicitly configured to enable this behavior.

Documentation/ipvlan.md
```diff
@@ -38,3 +38,40 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
...
+For example, ISC dhcpd needs to have the following directive in its dhcpd.conf:
```
joshix: For example, the ISC dhcpd is configured to echo Client Identifiers with the following `dhcpd.conf` directive:

joshix:
@eyakubovich Reviewed. Comments inline.

eyakubovich:
@dcbw You're right that XID is better than ClientID since servers don't echo it well. It's 4 bytes long, I think it's enough entropy. Obviously we're loosing the hardware address bits as they'll be all the same. I'll rework the patch.

eyakubovich:
@joshix Thank you for the review.

eyakubovich:
@dcbw Updated per your suggestion to just match on XID (it was already doing that). However it doesn't actually quite work either. Looks like both dnsmasq and dhcpd use the following algorithm when processing DHCPDISCOVER:

1. If ClientID option is there, use it to look for existing lease
2. If no ClientID or no lease found, use h/w addr (`chaddr`) to look for lease
3. If none found, create a new lease

The trouble is if the host's dhcp client (e.g. dhclient on my laptop) has already received a lease without sending a ClientID. When the CNI plugin makes a request, step 2 finds the host's lease and returns it.

dcbw:
@eyakubovich if I understand you correctly, you are saying that:

  1. a previous lease has been acquired using the MAC address of the ipvlan parent interface, but no ClientID was attached to the lease
  2. the new ipvlan DHCP request does send a client ID
  3. the DHCP server ignores the client ID and matches only on the MAC address, and returns the parent interface's lease to the ipvlan interface?

eyakubovich:
@dcbw Exactly.

steveeJ:
[RFC2131](http://www.faqs.org/rfcs/rfc2131.html) says that

    The protocol defines
    that the key will be (IP-subnet-number, hardware-address) unless the
    client explicitly supplies an identifier using the 'client
    identifier' option.

IIUC the behavior described by @eyakubovich doesn't conform to this RFC. The DHCP server should treat the request containing the 'client identifier' as a new client and not reuse the host's interface lease.

@eyakubovich are you going to pursue this in any of the upstream DHCP server implementations? 

I wonder how reliable the *client identifier* is in the real world with the numerous embedded DHCP server implementations in network equipment.

fnordahl:
@eyakubovich you could probably get the desired behaviour from the DHCP server by setting the chaddr field to zero in the discover and all subsequent messages sent to the server. (You may also have to set the htype and hlen to zero)

fnordahl:
@steveeJ you would be surprised as of how many interpertations and implementations there exist of the relatively simple DHCP protocol. A lot of network related issues have their source in their differences. Any implementation should be prepared to handle quirks. :-)

squeed:
The CNI repository has split; if you'd like to re-open this PR against the plugins repository, please do so.

PR #91, opened 2016-01-05
Update README.md

tomdee:
No need for full path to ifconfig (and the given path doesn't work...)


steveeJ:
Thanks for the update.

PR #98, opened 2016-01-22
spec: Removed reference to unsupported addIf attribute

krobertson:
Removes the "addIf" attribute from the example of a bridge network
definition. This doesn't appear to exist in the implementation.


jonboulle:
lgtm, thanks for catching this

PR #102, opened 2016-01-27
scripts/exec-plugins: print plugin results if DEBUG>0

steveeJ:


jonboulle:
Not sure what priv net run in the commit header is referencing. Other than that seems fine.

PR #103, opened 2016-01-27
*: DNS information as dictionary, adding domain, search domains and options

### LINKED ISSUE #100: *: DNS option only allows to configure nameserver entries
http://linux.die.net/man/5/resolv.conf has some more options than *nameserver*. 
I suggest to add *search* and *domain*. I'm not sure if we want/need the others.

/cc @alban

steveeJ:
/cc @alban 

@thockin might be interested and helpful too

Fixes #100.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -50,7 +50,7 @@ The operations that the CNI plugin needs to support are:
     - **Name of the interface inside the container**. This is the name that should be assigned to the interface created inside the container (network namespace); consequently it must comply with the standard Linux restrictions on interface names.
   - Result:
     - **IPs assigned to the interface**. This is either an IPv4 address, an IPv6 address, or both.
-    - **List of DNS nameservers**. This is a priority-ordered list of IPv4 and IPv6 addresses of DNS nameservers.
+    - **DNS information**. Dictionary that includes DNS information for nameservers, domain, search domains and options.
 
 - Delete container from network
   - Parameters:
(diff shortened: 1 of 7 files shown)


SPEC.md
```diff
@@ -130,6 +134,11 @@ The network configuration is described in JSON form. The configuration can be st
   - `routes` (list): List of subnets (in CIDR notation) that the CNI plugin should ensure are reachable by routing them through the network. Each entry is a dictionary containing:
     - `dst` (string): subnet in CIDR notation
     - `gw` (string): IP address of the gateway to use. If not specified, the default gateway for the subnet is assumed (as determined by the IPAM plugin).
+- `dns`: Dictionary with DNS specific values:
+  - `nameservers` (list of strings): list of a priority-ordered list of DNS nameservers that this network is aware of. Each entry in the list is a string containing either an IPv4 or an IPv6 address.
+  - `domain` (string): the local domain used for short hostname lookups.
```
thockin: might be worth noting that `domain` and `search` are usually mutually exclusive

thockin:
LGTM otherwise

SPEC.md
```diff
@@ -91,16 +91,20 @@ Success is indicated by a return code of zero and the following JSON printed to
...
 ```
 
 `cniVersion` specifies a [Semantic Version 2.0](http://semver.org) of CNI specification used by the plugin.
-The "dns" field contains a list of a priority-ordered list of DNS nameservers that this network is aware of.
-Each entry in the list is a string containing either an IPv4 or an IPv6 address.
-Typically this value would just be the value returned by the IPAM plugin.
-It is outside the scope of this specification how the container runtime uses the list of DNS nameservers from each of the networks to provide name resolution services to the container.
-Examples of how this list could be used include generating an `/etc/resolv.conf` file to be injected into the container filesystem or running a DNS forwarder on the host.
+`dns` field contains a dictionary consisting of common DNS information that this network is aware of.
+Please see how DNS can be [specified in the configuration](#network-configuration), as The result is returned in the same format.
```
jonboulle: The result is returned in the same format as specified in the [configuration](#network-configuration).

jonboulle:
LGTM

SPEC.md
dcbw: Definitely; in most (all?) cases the underlying DNS implementation will enforce some kind of mutual exclusivity, so that should be reflected in the docs here.  Plugins cannot expect that giving both will actually use both, unless there's some hackery elsewhere.  So it should be defined explicitly what the precedence is.

I vote for saying something like "if search is given, it is preferred over domain".

steveeJ: Thanks, I will add a note about how  most resolvers handle this.

Some background info:
The idea is to have a convenient way to pass a domain to the runtime which might want to set a new domain name in case it created a new UTS namespace. How runtimes handle multiple networks that pass along a domain is beyond scope though, as well as how it handles the case of *domain* and *search* given. Personally I'd prepend *domain* to *search* if it's not already included and use this as the new *search*.

steveeJ:
Thanks for the reviews!

PR #104, opened 2016-01-27
Revendor go-iptables to get --wait behavior

dcbw:
We need the wait behavior, otherwise we get:

Jan 27 19:40:39 kubernetes-node-1 kubelet[4897]: E0127 19:40:39.465904    4897 manager.go:1885] Failed to setup network for pod "heapster-v11-dgahz_kube-system(a55f02dc-c52c-11e5-803e-525400ebdc4d)" using network plugins "kubenet": Error adding container to network: exit status 4: Another app is currently holding the xtables lock. Perhaps you want to use the -w option?

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -7,7 +7,7 @@
 	"Deps": [
 		{
 			"ImportPath": "github.com/coreos/go-iptables/iptables",
-			"Rev": "83dfad0f13fd7310fb3c1cb8563248d8d604b95b"
+			"Rev": "90456be57fcb8185b264b77ce42a9539df42df25"
 		},
 		{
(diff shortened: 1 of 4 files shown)


dcbw:
@jonboulle seems like go-iptables is due for an update...

jonboulle:
LGTM

jonboulle:
@dcbw sorry, what do you mean? this patch, or something else?

dcbw:
@jonboulle when using CNI plugins with kubernetes in the above-referenced Kubernetes issue, we run into the iptables wait error when starting multiple pods at the same time.  So we need the wait handling to make sure pods can actually start with ipMasq support.  Thanks!

PR #109, opened 2016-01-31
bugfix in LoadArgs

aanm:
We should not fail if a given argument doesn't match any interface's
fields name.

Signed-off-by: André Martins <aanm90@gmail.com>

### DIFF
--- a/pkg/types/args.go
+++ b/pkg/types/args.go
@@ -38,7 +38,7 @@ func LoadArgs(args string, container interface{}) error {
 		valueString := kv[1]
 		keyField := containerValue.Elem().FieldByName(keyString)
 		if !keyField.IsValid() {
-			return fmt.Errorf("ARGS: invalid key %q", keyString)
+			continue
 		}
 		u := keyField.Addr().Interface().(encoding.TextUnmarshaler)
 		err := u.UnmarshalText([]byte(valueString))


steveeJ:
Apologies for my slowness, but I don't quite follow the use-case here. Please give us an example where this error occurs but we would rather ignore it.

aanm:
@steveeJ The code is forcefully checking for **all** arguments that are passed under `args` to be present in all fields of the given interface `container`. i.e. it fails if I pass the argument `foo` but the interface `container` doesn't have that field name, which I think it should discarded and not give any error about it.

steveeJ:
After thinking about this for a while I think I understand the problem but  I don't think your code solves the problem.

AFAIU CNI_ARGS might call `LoadArgs()` but don't have the field name in the passed interface, because CNI_ARGS contains keys that are not meant for all plugins. This is partially a specification problem because a simple `K=V` doesn't specify a target plugin. So when args contain keys that are to be passed through to cascaded plugins (host -> ipam in most cases) it will error out.

I see the following options:
* change the SPEC to where arguments must target plugin names
* keep track of all used arguments and error out if after processing all plugins there are arguments that haven't been used.

aanm:
>AFAIU CNI_ARGS might call LoadArgs() but don't have the field name in the passed interface, because CNI_ARGS **contains keys that are not meant for all** plugins.

Correct so it should discard the non-used keys, right?

>So when args contain keys that are to be passed through to cascaded plugins (host -> ipam in most cases) it will error out.

Exactly but they are still being passed from the host plugin to the ipam, so why should the ipam, for example, error out because of host plugin specific arguments?

steveeJ:
> Correct so it should discard the non-used keys, right?

Not quite, at least not only that. Might you have missed the suggestions from my previous comment? I'd appreciate feedback on that.

aanm:
Sorry I didn't miss them, I didn't want to reply before I test out something first.

So, without this patch, I get the following errors with kubernetes and cni:
```
# arguments passed to the plugin: "K8S_POD_NAMESPACE=default;K8S_POD_NAME=my-nginx-ca7m8;K8S_POD_INFRA_CONTAINER_ID=662e690afd1cda30efa039a97bebbc3369d415b361ae370a6612d5531435c74f"
E0212 02:19:30.720906    8292 cni.go:186] Error deleting network: ARGS: invalid key "K8S_POD_NAMESPACE"
```
Which they are set here: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/network/cni/cni.go#L201

>change the SPEC to where arguments must target plugin names

IIRC that would break the way bridge plugin calls the host-local plugin for example: https://github.com/appc/cni/blob/master/plugins/main/bridge/bridge.go#L187
>keep track of all used arguments and error out if after processing all plugins there are arguments that haven't been used.

I'm actually not following why is such a big deal to use all arguments passed to the argument, can you explain it? Even if the `CNI_ARGS` contains keys that are not meant for all plugins, the reason in here is to discard arguments that don't belong to any plugin at all, which IMHO it should be fine.

steveeJ:
> I'm actually not following why is such a big deal to use all arguments passed to the argument, can you explain it?

Simple, it enforces correctness or differently put prohibits sloppiness. Why should we allow sloppy behavior by design?

dcbw:
@steveeJ I don't think it's sloppiness.  Kubernetes may pass arguments that are optional which the plugin may not care about.  For example, the plugin may not care about the Kubernetes namespace name because it may have a cache mapping pod names to network settings, or something like that.  The plugin has no control over the arguments passed by a generic runtime like Kubernetes or rkt, and attempting to standardize arguments between the two is probably not going to work well.

dcbw:
Also, I'm not sure what "correctness" means WRT CNI args passed to plugins.  The runtime (Kube/rkt/whatever) has no idea what arguments the plugin expects.  Likewise the plugin may expect to receive some arguments from the runtime, and if it doesn't get what it expects, it is free to return an error the runtime.  I don't think it's possible to define what's "correct" here; well OK it's possible, but I don't think it's very useful.

aanm:
Exactly. IMO it is better to see a warning from the CNI plugin about args that couldn't be parsed than completely fail the allocation of an interface.

yifan-gu:
> Exactly. IMO it is better to see a warning from the CNI plugin about args that couldn't be parsed than completely fail the allocation of an interface.

@steveeJ I think we probably don't want CNI to error out in all these cases. A warning message seems fine to me. Or we can provide an option in the config, say whether we should allow optional args or not?

PR #117, opened 2016-02-08
new plugin: tuning

alban:
Allow users to tune net network parameters such as somaxconn.

With this patch, users can add a new network configuration:
```
> {
>   "name": "mytuning",
>   "type": "tuning",
>   "sysctl": {
>           "net.core.somaxconn": "500"
>   }
> }
```
The value /proc/sys/net/core/somaxconn will be set to 500 in the network
namespace but will remain unchanged on the host.

Only sysctl parameters that belong to the network subsystem can be
modified.

Related to: https://github.com/coreos/rkt/pull/2140

### DIFF
--- a/Documentation/tuning.md
+++ b/Documentation/tuning.md
@@ -1,0 +1,36 @@
+# tuning plugin
+
+## Overview
+
+This plugin can change some system controls (sysctls) in the network namespace.
+It does not create any network interfaces and therefore does not bring connectivity by itself.
+It is only useful when used in addition to other plugins.
+
+## Operation
+The following network configuration file
+```
+{
+  "name": "mytuning",
+  "type": "tuning",
+  "sysctl": {
+          "net.core.somaxconn": "500"
+  }
+}
+```
+will set /proc/sys/net/core/somaxconn to 500.
+Other sysctls can be modified as long as they belong to the network namespace (`/proc/sys/net/*`).
+
+A successful result would simply be:
+```
+{
+  "cniVersion": "0.1.0"
+}
+```
+
+## Network sysctls documentation
+
+Some network sysctls are documented in the Linux sources:
+
+- [Documentation/sysctl/net.txt](https://www.kernel.org/doc/Documentation/sysctl/net.txt)
+- [Documentation/networking/ip-sysctl.txt](https://www.kernel.org/doc/Documentation/networking/ip-sysctl.txt)
+- [Documentation/networking/](https://www.kernel.org/doc/Documentation/networking/)
--- a/plugins/meta/tuning/tuning.go
+++ b/plugins/meta/tuning/tuning.go
@@ -1,0 +1,83 @@
+// Copyright 2016 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This is a "meta-plugin". It reads in its own netconf, it does not create
+// any network interface but just changes the network sysctl.
+
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/appc/cni/pkg/ns"
+	"github.com/appc/cni/pkg/skel"
+	"github.com/appc/cni/pkg/types"
+)
+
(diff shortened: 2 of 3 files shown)


alban:
I am suggesting this "tuning" network plugin to CNI: https://github.com/appc/cni/pull/117

Documentation/tuning.md
```diff
@@ -0,0 +1,36 @@
+# tuning plugin
+
+## Overview
+
+This plugin can change some system controls (sysctls) in the network namespace.
+It does not create any network interface and therefore does not bring connectivity by itself.
```
steveeJ: interface**s**

Documentation/tuning.md
```diff
@@ -0,0 +1,36 @@
...
+It is useful when used in addition to other plugins.
```
steveeJ: It is *only* useful..

steveeJ:
Haven't finished reviewing the code because I was interrupted by the following thought:
If this is going to be used against a running container we should be able to reverse what it did.
So this should
1. store or return the previous values of what changed in the results
2. implement delete which would reset the values

I'm undecided about 1. It comes down to if we want to be responsible for persisting state for this plugin or not.

alban:
What should we be able to reverse what this plugin did?

steveeJ:
Taken from your example. If the plugin adjusts the following value

> > "sysctl": { > "net.core.somaxconn": "500" > }

it should first read what's defined to restore this value on plugin delete.

alban:
Oh sorry, I meant *why* should we do that? This will all go away when the network namespace is released.

steveeJ:
This plugin might be added and deleted at runtime of a container without the whole container going away. We need to design the plugins for such use-cases.

Documentation/tuning.md
alban: fixed

Documentation/tuning.md
alban: ok

alban:
> This plugin might be added and deleted at runtime of a container without the whole container going away. We need to design the plugins for such use-cases.

For this "tuning" plugin, I am not sure it is important in practice: for example, changing "net.core.somaxconn" back to 128 would not change the somaxconn of existing sockets but only of newly created ones. I could serialize and deserialize the selected net sysctl at the cost of more complex code. Should I do this in this PR or a following one?

Do you know if the format for writing and reading net sysctl is always the same? I am not aware of any net sysctl that couldn't be restored, I am just double-checking.

slallema:
I agree that dealing with state adds complexity and in that case I'm not sure it's worth it.
I can't think of a use case where you would change these setting back to default at runtime.

steveeJ:
@alban please add a TODO comment in the `cmdDel` function that reminds us that the revert scenario could/should eventually be added.

alban:
@steveeJ ok, updated.

steveeJ:
Tnanks @alban. I think that we need better category names than "meta" but let's not block your PR on that issue.

@squaremo PTAL and feel free to merge if you spot no problems.

jellonek:
I'm not quite sure that this functionality should belong to CNI, because it provides much greater functionality to tune every subsystem in `/proc/sys`, so also oom_killer settings, dirty pages and... hostname (`/proc/sys/kernel/{host,domain}name`).

Please look how many other things in `/proc/sys/{vm,kernel}` could be tuned that way but... what also could mess with such container invocation for other living containers (and host).

Very nice feature, but also very dangerous. Maybe we should make some whitelist for settings which could be modified by this plugin?

jellonek:
My point is: if this has to be set by sysadmin with knowledge about what he is doing - he also easily will do this using basic shell scripts, but if we provide such method for less experienced users - we should fortify it "a bit more", so they will not flip on something "funny" sent as a joke by other user as "usage example".

alban:
@jellonek Things like `/proc/sys/kernel/host` cannot be changed by this plugin because it checks that the path starts with `/proc/sys/net/`. The "net" directory should only contain settings that belong to the network namespace, so it should not be able to change settings on the host network namespace or other containers.

jellonek:
Argh, my mistake. Missreaded the code...

iaguis:
LGTM

squaremo:
> PTAL and feel free to merge

@steveeJ Sorry about that, I was on vacation. I'm OK with this being merged, ftr :)

I'm curious about the intended mode of use, though, @alban -- in my reading, each container would need a network to do the tuning in addition to which networks it was supposed to belong to. It seems like a more natural way to do it would be to decorate the "real" network with the tuning. Am I out of touch?

alban:
I don't have a strong opinion on "tuning" being a separate plugin or "tuning" as decoration. My arguments in favor of a separate plugin are:
- the plugins developed out of tree such as  [calico-cni](https://github.com/projectcalico/calico-cni) can automatically be used next to the "tuning" plugin without having to implement decorations.
- the settings from "tuning" are not specific to a network but for the whole network namespace. If the user has several networks with different decorations, the last one in the list will win. Being independent  from other networks look cleaner to me.

However, there could be other plugins where decoration seems better, like decoration adding traffic control rules on a network interface...

PR #127, opened 2016-02-17
Skip unknown ARGS in host-local rather than error

### LINKED ISSUE #126: host-local IPAM plugin doesn't work with kubernetes
When using the CNI plugin with Kubernetes, a number of additional arguments are [provided](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/network/cni/cni.go#L201). This causes the `host-local` plugin to exit with

```
2016-02-17 19:03:39,885 28126 [default/busybox] INFO Using IPAM plugin at: /opt/cni/bin/host-local
2016-02-17 19:03:39,888 28126 [default/busybox] ERROR IPAM plugin error (rc=1): {
    "code": 100,
    "msg": "ARGS: invalid key \"K8S_POD_NAMESPACE\""
}
```

tomdee:
Fixes #126 

Signed-off-by: Tom Denham <tom@tomdee.co.uk>

### DIFF
--- a/pkg/types/args.go
+++ b/pkg/types/args.go
@@ -38,7 +38,8 @@ func LoadArgs(args string, container interface{}) error {
 		valueString := kv[1]
 		keyField := containerValue.Elem().FieldByName(keyString)
 		if !keyField.IsValid() {
-			return fmt.Errorf("ARGS: invalid key %q", keyString)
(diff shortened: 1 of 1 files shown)


steveeJ:
Duplicate of #109.

PR #132, opened 2016-02-23
scripts: add DEBUG option

steveeJ:


achanda:
This won't work if an user sets `DEBUG=true`. How about just checking if the variable is set?
`if [[ -n ${DEBUG} ]]`

steveeJ:
I didn't expect any comments on this one ;-) Your suggestion however eliminates the option of actually introducing debug levels in the future.

achanda:
Expect the unexpected, I guess :)
I think some user might be confused when they set that var to `true` and there is no debug output. However, I do see your point here.

steveeJ:
I'll take this as is, since the scripts are mainly for testing purposes this shouldn't bother anyone too much. Still, thanks for the 2nd pair of eyes ;-)

PR #134, opened 2016-02-25
Return v4 and v6 info in host-local IPAM

### LINKED ISSUE #131: Can't get IPv6 addresses from the host-local plugin
The documentation says that it can allocate IPv6 addresses, but here is what I see

```
ubuntu@dev:~/gosrc/src/github.com/appc/cni/bin$ sudo ./host-local < /etc/cni/net.d/10-mynet.conf
{
    "ip4": {
        "ip": "3ffe:ffff:0:1ff::2/64",
        "gateway": "3ffe:ffff:0:1ff::1",
        "routes": [
            {
                "dst": "3ffe:ffff:0:1ff::1/64"
            }
        ]
    },
    "dns": {}
}
ubuntu@dev:~/gosrc/src/github.com/appc/cni/bin$ cat /etc/cni/net.d/10-mynet.conf
{
    "name": "ipv6",
    "ipam": {
        "type": "host-local",
        "subnet": "3ffe:ffff:0:01ff::/64",
        "range-start": "3ffe:ffff:0:01ff::0010",
        "range-end": "3ffe:ffff:0:01ff::0020",
        "routes": [
            { "dst": "3ffe:ffff:0:01ff::1/64" }
        ]
    }
}
ubuntu@dev:~/gosrc/src/github.com/appc/cni/bin$ env | grep CNI
CNI_COMMAND=ADD
CNI_IFNAME=em0
CNI_NETNS=test
CNI_CONTAINERID=f81d4fae-7dec-11d0-a765-00a0c91e6bf6
CNI_PATH=/etc/cni/net.d/10-mynet.conf
```

achanda:
Closes #131

### DIFF
--- a/plugins/ipam/host-local/config.go
+++ b/plugins/ipam/host-local/config.go
@@ -39,12 +39,12 @@ type IPAMArgs struct {
 }
 
 type Net struct {
-	Name string      `json:"name"`
-	IPAM *IPAMConfig `json:"ipam"`
+	Name string       `json:"name"`
+	IPAM []IPAMConfig `json:"ipam"`
 }
 
 // NewIPAMConfig creates a NetworkConfig from the given network name.
-func LoadIPAMConfig(bytes []byte, args string) (*IPAMConfig, error) {
+func LoadIPAMConfig(bytes []byte, args string) ([]IPAMConfig, error) {
 	n := Net{}
 	if err := json.Unmarshal(bytes, &n); err != nil {
 		return nil, err
@@ -61,9 +61,9 @@ func LoadIPAMConfig(bytes []byte, args string) (*IPAMConfig, error) {
 	if n.IPAM == nil {
 		return nil, fmt.Errorf("%q missing 'ipam' key")
 	}
-
-	// Copy net name into IPAM so not to drag Net struct around
-	n.IPAM.Name = n.Name
+	if len(n.IPAM) > 2 {
+		return nil, fmt.Errorf("Found more than two IPAM blocks")
+	}
 
 	return n.IPAM, nil
 }
--- a/plugins/ipam/host-local/main.go
+++ b/plugins/ipam/host-local/main.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"fmt"
 	"github.com/appc/cni/plugins/ipam/host-local/backend/disk"
 
 	"github.com/appc/cni/pkg/skel"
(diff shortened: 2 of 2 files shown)


achanda:
@steveeJ got something to work. This looks like this, please review.
```
ubuntu@dev:~/gosrc/src/github.com/appc/cni/bin$ sudo ./host-local < /etc/cni/net.d/20-mynet.conf
{
    "ip4": {
        "ip": "10.22.0.5/16",
        "gateway": "10.22.0.1",
        "routes": [
            {
                "dst": "0.0.0.0/0"
            }
        ]
    },
    "ip6": {
        "ip": "3ffe:ffff:0:1ff::5/64",
        "gateway": "3ffe:ffff:0:1ff::1",
        "routes": [
            {
                "dst": "3ffe:ffff:0:1ff::1/64"
            }
        ]
    },
    "dns": {}
}ubuntu@dev:~/gosrc/src/github.com/appc/cni/bin$ cat /etc/cni/net.d/20-mynet.conf
{
    "name": "mynet",
    "type": "ptp",
    "isGateway": true,
    "ipMasq": true,
    "ipam": [{
        "type": "host-local",
        "subnet": "10.22.0.0/16",
        "routes": [
            { "dst": "0.0.0.0/0" }
        ]
    },
    {
        "type": "host-local",
        "subnet": "3ffe:ffff:0:01ff::/64",
        "range-start": "3ffe:ffff:0:01ff::0010",
        "range-end": "3ffe:ffff:0:01ff::0020",
        "routes": [
            { "dst": "3ffe:ffff:0:01ff::1/64" }
        ]
    }]
}
```

steveeJ:
AFAICT this is not backwards compatible to currently valid configurations using the dictionary.

Besides, it breaks the spec ;-) which says: `ipam: Dictionary with IPAM specific values: ...`

dcbw:
A suggestion that doesn't break the spec (which IMHO doesn't need to be broken here, just extended) would be something like:

```
ubuntu@dev:~/gosrc/src/github.com/appc/cni/bin$ cat /etc/cni/net.d/20-mynet.conf
{
    "name": "mynet",
    "type": "ptp",
    "isGateway": true,
    "ipMasq": true,
    "ipam": {
        "type": "host-local",
        "subnet": "10.22.0.0/16",
        "routes": [
            { "dst": "0.0.0.0/0" }
        ]
        "subnet6": "3ffe:ffff:0:01ff::/64",
        "range6-start": "3ffe:ffff:0:01ff::0010",
        "range6-end": "3ffe:ffff:0:01ff::0020",
        "routes6": [
            { "dst": "3ffe:ffff:0:01ff::1/64" }
        ]
    }
}
```

Alternatively, if we thing there's a good reason to use separate IPAM plugins for IPv4 and IPv6 (and there is an argument for that, I think) then it would be best to have a new 'ipam6' block instead.  But I don't think we should be trying to use an array here.

dcbw:
The argument for multiple IPAM plugins would be something like this...  IPv6 addressing is very different from IPv4 and it's certainly possible to do things like actual DHCP for IPv4 and SLAAC for IPv6 (including DHCPv6 for information-only mode).  I'm not sure we should require all IPAM plugins to handle all addressing types.

achanda:
So, few more options now that array is not on the table
```
{
    "name": "mynet",
    "type": "ptp",
    "isGateway": true,
    "ipMasq": true,
    "ipam": {
        "type": "host-local",
        "subnet": "10.22.0.0/16",
        "routes": [
            { "dst": "0.0.0.0/0" }
        ]
        "subnet6": "3ffe:ffff:0:01ff::/64",
        "range6-start": "3ffe:ffff:0:01ff::0010",
        "range6-end": "3ffe:ffff:0:01ff::0020",
        "routes6": [
            { "dst": "3ffe:ffff:0:01ff::1/64" }
        ]
    }
}
```
and
```
{
	"name": "mynet",
	"type": "ptp",
	"isGateway": true,
	"ipMasq": true,
	"ipam": {
		"ip4": {
			"type": "host-local",
			"subnet": "10.22.0.0/16",
			"routes": [{
				"dst": "0.0.0.0/0"
			}]
		},
		"ip6": {
                        "type": "host-local"
			"subnet6": "3ffe:ffff:0:01ff::/64",
			"range6-start": "3ffe:ffff:0:01ff::0010",
			"range6-end": "3ffe:ffff:0:01ff::0020",
			"routes6": [{
				"dst": "3ffe:ffff:0:01ff::1/64"
			}]
		}
	}
}
```
and
```
{
	"name": "mynet",
	"type": "ptp",
	"isGateway": true,
	"ipMasq": true,
	"ipam": {
		"type": "host-local",
		"subnet": "10.22.0.0/16",
		"routes": [{
			"dst": "0.0.0.0/0"
		}]
	},
	"ipam6": {
		"type": "host-local",
		"subnet": "3ffe:ffff:0:01ff::/64",
		"range-start": "3ffe:ffff:0:01ff::0010",
		"range-end": "3ffe:ffff:0:01ff::0020",
		"routes": [{
			"dst": "3ffe:ffff:0:01ff::1/64"
		}]
	}
}
```
I believe all of these maintain backward compatibility. Any thoughts on which one should be implemented?

dcbw:
I don't think the second one (ipv4 and ipv6 members inside the ipam dict) keeps compat, since existing plugins will expect keys within the ipam struct already.

That said, due to the multi-ipam thing I talk about above, I'm slightly favor the "ipam6" one.

achanda:
Closing this since we are sure this is not what we want. Refer #137

PR #137, opened 2016-02-26
Enable IPv6 support in the host-local driver

achanda:
cc @steveeJ @dcbw @alban

### DIFF
--- a/plugins/ipam/host-local/config.go
+++ b/plugins/ipam/host-local/config.go
@@ -41,8 +41,8 @@ type IPAMArgs struct {
 
 type Net struct {
 	Name  string      `json:"name"`
-	IPAM  *IPAMConfig `json:"ipam"`
-	IPAM6 *IPAMConfig `json:"ipam6"`
+	IPAM  *IPAMConfig `json:"ipam,omitempty"`
+	IPAM6 *IPAMConfig `json:"ipam6,omitempty"`
 }
 
 // LoadIPAMConfig unmarshals a given byte slice to a Net object
@@ -53,10 +53,19 @@ func LoadIPAMConfig(bytes []byte, args string) (*IPAMConfig, *IPAMConfig, error)
 	}
 
 	if args != "" {
-		n.IPAM.Args = &IPAMArgs{}
-		err := types.LoadArgs(args, n.IPAM.Args)
-		if err != nil {
-			return nil, nil, err
+		if n.IPAM != nil {
+			n.IPAM.Args = &IPAMArgs{}
+			err := types.LoadArgs(args, n.IPAM.Args)
+			if err != nil {
+				return nil, nil, err
+			}
+		}
+		if n.IPAM6 != nil {
+			n.IPAM6.Args = &IPAMArgs{}
+			err := types.LoadArgs(args, n.IPAM6.Args)
+			if err != nil {
+				return nil, nil, err
+			}
 		}
 	}
 
--- a/plugins/ipam/host-local/config.go
+++ b/plugins/ipam/host-local/config.go
@@ -40,31 +40,29 @@ type IPAMArgs struct {
 }
 
 type Net struct {
-	Name string      `json:"name"`
-	IPAM *IPAMConfig `json:"ipam"`
+	Name  string      `json:"name"`
+	IPAM  *IPAMConfig `json:"ipam"`
+	IPAM6 *IPAMConfig `json:"ipam6"`
 }
 
-// NewIPAMConfig creates a NetworkConfig from the given network name.
-func LoadIPAMConfig(bytes []byte, args string) (*IPAMConfig, error) {
+// LoadIPAMConfig unmarshals a given byte slice to a Net object
+func LoadIPAMConfig(bytes []byte, args string) (*IPAMConfig, *IPAMConfig, error) {
 	n := Net{}
 	if err := json.Unmarshal(bytes, &n); err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	if args != "" {
 		n.IPAM.Args = &IPAMArgs{}
 		err := types.LoadArgs(args, n.IPAM.Args)
 		if err != nil {
-			return nil, err
+			return nil, nil, err
 		}
 	}
 
-	if n.IPAM == nil {
-		return nil, fmt.Errorf("%q missing 'ipam' key")
+	if n.IPAM == nil && n.IPAM6 == nil {
+		return nil, nil, fmt.Errorf("Need at least one of ipam or ipam6 key")
 	}
 
-	// Copy net name into IPAM so not to drag Net struct around
-	n.IPAM.Name = n.Name
-
-	return n.IPAM, nil
+	return n.IPAM, n.IPAM6, nil
 }
--- a/plugins/ipam/host-local/config.go
+++ b/plugins/ipam/host-local/config.go
@@ -19,12 +19,13 @@ import (
 	"fmt"
 	"net"
 
-	"github.com/appc/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types"
 )
 
 // IPAMConfig represents the IP related network configuration.
 type IPAMConfig struct {
 	Name       string
+	Version    string        `json:"version"`
 	Type       string        `json:"type"`
 	RangeStart net.IP        `json:"rangeStart"`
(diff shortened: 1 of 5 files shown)


plugins/ipam/host-local/config.go
```diff
@@ -39,31 +39,29 @@ type IPAMArgs struct {
 }
 
 type Net struct {
-	Name string      `json:"name"`
-	IPAM *IPAMConfig `json:"ipam"`
+	Name  string      `json:"name"`
+	IPAM  *IPAMConfig `json:"ipam"`
+	IPAM6 *IPAMConfig `json:"ipam6"`
```
dcbw: Shouldn't at least one of these be omit-empty?  Actually I think both of them have to be omit-empty, but then the code needs to check that at least one of them is present and error out if not.  Otherwise adding ipam6 will require all network config to pass both ipam + ipam6, and nobody is passing ipam6 right now.

plugins/ipam/host-local/config.go
```diff
@@ -39,31 +39,29 @@ type IPAMArgs struct {
...
 		}
 	}
 
-	if n.IPAM == nil {
-		return nil, fmt.Errorf("%q missing 'ipam' key")
+	if n.IPAM == nil && n.IPAM6 == nil {
+		return nil, nil, fmt.Errorf("Need at least one of ipam or ipam6 key")
 	}
 
-	// Copy net name into IPAM so not to drag Net struct around
```
dcbw: There should probably be some minimal validation here too, to ensure that IPv4 values aren't passed in 'ipam6' and vice versa.  Since NewIPAllocator() isn't passed AF_INET/AF_INET6 it has no idea what kind of config it's getting in the v4 and v6 blocks in main.go and it will happily accept IPv4 values in the IPv6 allocator, I think.

plugins/ipam/host-local/config.go
```diff
@@ -39,31 +39,29 @@ type IPAMArgs struct {
...
-	n.IPAM.Name = n.Name
```
dcbw: Also, the Name field of IPAMConfig doesn't get marshalled at all, so it's going to be "" in both structures now.  And that now gets passed directly to disk.New() and will be the same for both IPv4 and IPv6.  First, I think each IPAM struct needs a name, so maybe just make n.IPAM.Name = n.Name, and n.IPAM6.Name = n.Name + "6"?  That way the stores aren't confused either.

plugins/ipam/host-local/config.go
achanda: Are you saying the case where an user puts IPv6 addresses in a "ipam" block or vice versa? main.go calls NewIPAllocator with IPv4 info only if the ipam key is set.

achanda:
@zachgersh review please? :)

zachgersh:
hey @achanda saw you made the naming change that was suggested. Marked this as needs review. Would you be willing to add some tests to cover the new code?

Some example tests are here:
https://github.com/appc/cni/blob/master/pkg/ns/ns_test.go

achanda:
@zachgersh sure, I can get that done.

plugins/ipam/host-local/config.go
```diff
@@ -39,31 +39,37 @@ type IPAMArgs struct {
...
 
 type Net struct {
-	Name string      `json:"name"`
-	IPAM *IPAMConfig `json:"ipam"`
+	Name  string      `json:"name"`
+	IPAM  *IPAMConfig `json:"ipam,omitempty"`
+	IPAM6 *IPAMConfig `json:"ipam6,omitempty"`
 }
 
 // NewIPAMConfig creates a NetworkConfig from the given network name.
```
zachgersh: This docstring is now wrong and probably should be a bit more descriptive?

achanda:
@zachgersh code coverage for the loopback and host-local plugins are at 0. Any idea why those tests are not being run?

steveeJ:
@achanda Tests that use external binaries are not counted in. If you write a test that uses the functions directly it should be fine.

achanda:
@steveeJ not very sure how to run the host-local plugin directly, is there an example? The tests in loopback seems to be running a binary as well. I took that as an example.

zachgersh:
@achanda there is no way to use it directly due to the fact that it is a main - you have to run the code either as a binary or via `go run`. That's the only type of testing you can do here and it is still valuable, even if the code coverage numbers don't go up.

achanda:
@zachgersh makes sense, this is ready for review then.

steveeJ:
> there is no way to use it directly due to the fact that it is a main

Actually in this case `main()` is just a wrapper to functions we should be able to test.

> That's the only type of testing you can do here and it is still valuable, even if the code coverage numbers don't go up.

I think we can just call the plugin\s `cmd{Add,Del}` from the tests, can't we?

achanda:
@zachgersh @steveeJ modified tests to call `cmdAdd` and `cmdDel` directly. I could not figure out how to read stdout when I'm not opening a new process. Please let me know if there is a known way to do that.

plugins/ipam/host-local/config.go
steveeJ: @achanda Please address this comment. I agree that both should be able to be empty.

test
```diff
@@ -11,7 +11,7 @@ set -e
 
 source ./build
 
-TESTABLE="plugins/ipam/dhcp plugins/main/loopback pkg/invoke pkg/ns pkg/skel pkg/types"
+TESTABLE="plugins/ipam/dhcp plugins/main/loopback plugins/ipam/host-local pkg/invoke pkg/ns pkg/skel pkg/types"
 FORMATTABLE="$TESTABLE libcni pkg/ip pkg/ns pkg/types pkg/ipam plugins/ipam/host-local plugins/main/bridge plugins/meta/flannel plugins/meta/tuning"
```
steveeJ: While you're at it, would you mind removing all the package names from `FORMATTABLE` that are now present in `TESTABLE`?

achanda: Done

plugins/ipam/host-local/config.go
steveeJ: What @dcbw is saying that we have no type checks at all to distinguish and check if the given IP information is appropriate for the respective fields. Before this PR, it is possible to pass both IPv4 and IPv6 information within IPAM and `host-local` will happily allocate an address. And after this PR it will be possible to pass both versions to both fields, which should not be the case anymore as we now want to distinguish input and output.

plugins/ipam/host-local/config.go
steveeJ: Disregarding the name. should we add a version field to the IPAM struct?

steveeJ: @dcbw @achanda any thoughts?

plugins/ipam/host-local/config.go
steveeJ: @achanda any thoughts?

achanda: @steveeJ sorry, I did not understand your last comment. Can you please give me an example? Do you mean the case where an user passes IPv6 info in `ipam` or the reverse?

plugins/ipam/host-local/config.go
achanda: @steveeJ use a single `ipam` key for both 4 and 6 using the version to demux? That sounds good. I can implement that.

achanda: @steveeJ ping! Wanted to check if this is what we want to do before I implement it.

achanda:
@steveeJ should I keep working on this?

steveeJ:
@achanda we (foremost @dcbw) are currently sorting out the IPv6 issues and PRs as we honestly have lost overview. Thanks for your patience on this one and I'm certain we'll know more in a couple days.

achanda:
@steveeJ thanks for the update, I will wait.

tgraf:
@achanda @dcbw  @steveeJ Given the discussion in isssue #245. Can we rebase and move on with this?

achanda:
Looks like this has diverged a bit from master. I will close this one and open a new PR against master.

steveeJ:
@achanda Next time please rebase the PR instead of creating a new one. It prevents flooding the project and also keeps discussions in one place.

tgraf:
This PR is obsoleted by PR #279. Can we close this?

rosenhouse:
Yup, closing.

PR #138, opened 2016-02-26
firewalld: integration

alban:
On distributions using FirewallD like Fedora, we need to register the
new IP to FirewallD.

Related to: https://github.com/coreos/rkt/issues/2206

### DIFF
--- a/Documentation/bridge.md
+++ b/Documentation/bridge.md
@@ -22,6 +22,9 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
 	"ipam": {
 		"type": "host-local",
 		"subnet": "10.10.0.0/16"
+	},
+	"firewalld": {
+		"zone": "trusted"
 	}
 }
 ```
@@ -35,3 +38,4 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
 * `ipMasq` (boolean, optional): set up IP Masquerade on the host for traffic originating from this network and destined outside of it. Defaults to false.
 * `mtu` (integer, optional): explicitly set MTU to the specified value. Defaults to the value chosen by the kernel.
 * `ipam` (dictionary, required): IPAM configuration to be used for this network.
+* `firewalld` (dictionary, optional): [FirewallD](https://fedoraproject.org/wiki/FirewallD)'s configuration to be used for this network.
--- a/Documentation/ptp.md
+++ b/Documentation/ptp.md
@@ -17,6 +17,9 @@ The traffic of the container interface will be routed through the interface of t
 	},
 	"dns": {
 		"nameservers": [ "10.1.1.1", "8.8.8.8" ]
+	},
+	"firewalld": {
+		"zone": "trusted"
 	}
 }
 
@@ -28,3 +31,4 @@ The traffic of the container interface will be routed through the interface of t
 * `mtu` (integer, optional): explicitly set MTU to the specified value. Defaults to value chosen by the kernel.
 * `ipam` (dictionary, required): IPAM configuration to be used for this network.
 * `dns` (dictionary, optional): DNS information to return as described in the [Result](/SPEC.md#result).
+* `firewalld` (dictionary, optional): [FirewallD](https://fedoraproject.org/wiki/FirewallD)'s configuration to be used for this network.
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -1,6 +1,7 @@
 {
 	"ImportPath": "github.com/appc/cni",
 	"GoVersion": "go1.6",
+	"GodepVersion": "v60",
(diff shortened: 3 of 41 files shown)


dcbw:
Alternatively, maybe use godbus and do it with D-Bus so there's less shelling out to stuff.  One example (mostly of listening, but also a method call) is https://github.com/kubernetes/kubernetes/blob/master/pkg/util/iptables/iptables.go#L165

steveeJ:
I second @dcbw's suggestion

@alban In addition I think it would be useful to have *zone* as an option to the plugin instead of hardcoding it.

alban:
@dcbw I updated the code to use D-Bus.

@steveeJ For the zone, should we still have "trusted" as a default? I am not sure how the user or the container run-time should choose the zone.

steveeJ:
@alban the zone can be an attribute in the plugin, so the user simply writes that in the network configuration file.

alban:
@steveeJ updated

steveeJ:
Thanks for this @alban, it's definitely a good start. I'd like to find a way that the *pkg/firewalld* code is not called from within the *ptp* package, but handled in the outer scope, which would currently be the skeleton plugin. 
If we merge it like this we'll have to change every plugin's implementation and configuration spec, which seems unnecessary. Do you agree?

alban:
@steveeJ but do ptp, bridge and other need to configure the firewall in the same way, namely by calling `org.fedoraproject.FirewallD1.zone.addSource($IP, $zone)`? If yes, then ok the new configuration parameter should be moved to `types.NetConf`.

Also, the call to FirewallD needs to be done after the call to IPAM because we need to know the new IP. So we might need to have code in each plugins.

philips:
We should figure out a way that we can chain the plugins so that not everything needs to implement firewallD logic.

alban:
But we cannot just add the firewallD logic on all plugins: overlay networks like flannel should not require any changes for firewallD, do they?

steveeJ:
@philips 
> We should figure out a way that we can chain the plugins so that not everything needs to implement firewallD logic.

+1 on that. Please state your opinion on #147 which where I started a discussion about _how_ to do this.

@alban 
> But we cannot just add the firewallD logic on all plugins

Exactly, we need this to be more generic. So we need (from you :smiley:?) is to write a _firewalld_ plugin. The plugin takes all the parameters required to act on an existing namespace with it's interfaces and addresses. If you need more help in designing this please don't hesitate to ping me.

alban:
> Exactly, we need this to be more generic. So what we need (from you ￼?) is to write a firewalld plugin. The plugin takes all the parameters required to act on an existing namespace with it's interfaces and addresses. If you need more help in designing this please don't hesitate to ping me.

We could have a firewalld plugin that is executed from ptp and bridge in the same way as the ipam plugin is executed with `ipam.ExecAdd()`. But it makes this more complicated to pass the parameters: the firewalld plugin would need the IP addresses that are decided by ipam. The IP is not in the configuration file so it would need to be passed via [CNI_ARGS](https://github.com/appc/cni/blob/master/SPEC.md#parameters) like "IP=1.2.3.4;IPv6=...". That's a lot more complicated that just calling a function in the firewalld pkg as it is now, and I don't see the benefit.

In any cases, we still need to add code in each plugins because we don't want to call the firewalld plugin from all main plugins: ptp and bridge would need it and flannel would not.

I could move the FirewallDZone field from each plugin struct to types.NetConf to make it more generic, but I prefer to keep the current logic as a pkg rather than a new plugin.

alban:
@steveeJ what kind of tests do you want for this PR?

alban:
Rebased on master to fix the conflict in Godeps/Godeps.json

steveeJ:
> The IP is not in the configuration file so it would need to be passed via CNI_ARGS like "IP=1.2.3.4;IPv6=...". That's a lot more complicated that just calling a function in the firewalld pkg as it is now, and I don't see the benefit

You don't have to pass configuration for the plugin via the args, instead just pass a JSON with the plugin configuration. The plugin can have arbitrary configuration fields.

> In any cases, we still need to add code in each plugins because we don't want to call the firewalld plugin from all main plugins: ptp and bridge would need it and flannel would not.

The container engine is in control of which plugins are called by configuring the JSON files, there is no need to add any additional plugin execution code to existing CNI plugins.

alban:
@steveeJ do you have time to finish this?

squeed:
I think this would be an excellent candidate for a chained plugin once that part of the standard is finalized.

voxadam:
Has there been any movement on this issue?

squeed:
As it stands, no, the CNI team doesn't really have the time to take this through to completion. However, all are welcome to make an independent plugin that works via chaining. Feel free to submit a pull request adding it to the list of plugins, too.

PR #140, opened 2016-02-29
Fix path to net namespace

achanda:
namespaces are created in `/var/run/netns`. Earlier, it was trying
to open a file in a wrong location and failing

cc @alban @steveeJ

### DIFF
--- a/pkg/ns/ns.go
+++ b/pkg/ns/ns.go
@@ -52,10 +52,11 @@ func SetNS(f *os.File, flags uintptr) error {
 // locked to an OS thread. If lockThread arg is true, this function
 // locks the goroutine prior to change namespace and unlocks before
 // returning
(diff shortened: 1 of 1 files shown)


dcbw:
Hmm, the CNI spec says: "CNI_NETNS: Path to network namespace file" so I'd assume that's supposed to be the full path.  What container runtime are you using and what is it passing for CNI_NETNS?

dcbw:
Also, only /sbin/ip actually creates netnamespaces in /var/run/netns, that's not a standardized thing.  If you simply run a container, the netns is shown in /proc/<pid>/ns/net as a link to the actual netnamespace fd.  /var/run/netns won't exist in that case.

achanda:
I've been doing `ip netns add test` and setting `CNI_NETNS=test`. I do see where I was going wrong. Thanks.

PR #142, opened 2016-03-02
Fix loopback result

### LINKED ISSUE #141: plugins/loopback: unmarshalling error when added with cnitool
Placing the configuration `/etc/cni/net.d/00-lo.conf`:
```
{
    "name": "lo",
    "type": "loopback"
}
```

The attempt to add it with `cnitool` fails like so:
```
 bin/ # CNI_PATH=`pwd` ./cnitool add lo /proc/self/ns/net
add: unexpected end of JSON input
```
While *lo* is up the error is valid. The plugin doesn't return a valid result.

This should also be covered in a test.

/cc @zachgersh

steveeJ:
Fixes #141.

---
- [x] Fix result
- [x] Add testcase

### DIFF
--- a/plugins/main/loopback/loopback.go
+++ b/plugins/main/loopback/loopback.go
@@ -33,7 +33,24 @@ func cmdAdd(args *skel.CmdArgs) error {
 }
 
 func cmdDel(args *skel.CmdArgs) error {
-	// del does nothing, we're going to destroy the device anyway
+	args.IfName = "lo" // ignore config, this only works for loopback
+	err := ns.WithNetNSPath(args.Netns, false, func(hostNS *os.File) error {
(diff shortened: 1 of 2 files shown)


jonboulle:
pluings -> plugins in https://github.com/steveeJ/cni/commit/539b04fd3e538da9bc4d78f628676be0983a54bd commit message

steveeJ:
Thanks @jonboulle :-) Fixed and testcase added.

zachgersh:
@steveeJ - this is great!

Could we actually add a test case for del though? I could totally help with this as needed.

steveeJ:
@zachgersh sure! I added the testcase.

zachgersh:
@steveeJ Awesome! LGTM - once the tests pass ;)

PR #143, opened 2016-03-02
More tests

zachgersh:
Now that the ginkgo / gomega dependencies have been merged - the tests for find can be turned on.

I have also added a travis badge to show that the build is passing!


steveeJ:
This merely activates tests that are already there, merging right away since it's green.

Issue #123, opened 2016-02-11
flannel: Why does flannel call ipam.ExecAdd() to invoke delegated plugin


qianzhangxa:
In the `delegateAdd()` of flannel plugin (https://github.com/appc/cni/blob/master/plugins/meta/flannel/flannel.go#L158), I see `ipam.ExecAdd()` is called to invoke the delegated plugin (e.g., bridge), it looks a little strange to me that a IPAM method is called to invoke a main plugin, Why not directly call `invoke.ExecPluginWithResult()` to invoke the main plugin?

steveeJ:
@squaremo can you jump in here? Github told me you wrote the line in question :-)

squaremo:
Ahhh yes, I think I know this one -- the Exec* procedures originally lived in the `ipam` package. I factored them out so they could be used by libcni, leaving shims behind for convenience -- and apparently neglected to update this use. I'll correct it.

PR #144, opened 2016-03-02
Add invoke.Delegate{Add,Del} for use by meta-plugins

squaremo:
The 'flannel' meta plugin delegates to other plugins to do the actual
OS-level work. It used the ipam.Exec{Add,Del} procedures for this
delegation, since those do precisely what's needed.

However this is a bit misleading, since the flannel plugin _isn't_
doing this for IPAM, and the ipam.Exec* procedures aren't doing
something specific to IPAM plugins.

So: anticipating that there may be more meta plugins that want to
delegate in the same way, this commit moves generic delegation
procedures to `pkg/invoke`, and makes the `pkg/ipam` procedures (still
used, accurately, in the non-meta plugins) shims.

See issue #123.

### DIFF
--- a/pkg/invoke/delegate.go
+++ b/pkg/invoke/delegate.go
@@ -1,0 +1,39 @@
+package invoke
+
+import (
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/appc/cni/pkg/types"
+)
+
+func DelegateAdd(delegatePlugin string, netconf []byte) (*types.Result, error) {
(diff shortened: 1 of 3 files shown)


steveeJ:
IMHO the functions deserve their own file in the package, maybe `delegate.go`?
Otherwise, LGTM!

squaremo:
> the functions deserve their own file in the package, maybe `delegate.go`

Fair enough -- moved those, and rebased on current master.

Issue #112, opened 2016-01-31
What happens if the container already has the interface "inside the container"?


aanm:
> Name of the interface inside the container. This is the name that should be assigned to the interface created inside the container (network namespace); consequently it must comply with the standard Linux restrictions on interface names.

The spec doesn't have anything related to that. So what is the procedure? Are the plugins allowed to remove, make the necessary modifications, and then recreate the interface with the same name?

dcbw:
Removing the interface may break existing connections and cause apps inside the container to fail when their interface goes away and sockets are closed and things like that.  So I would imagine that completely removing the interface would not be desirable.

I'm assuming this item is in the spec because apps may reference the network interface by name or something, or maybe the runtime may need to modify the interface in some way.

It's worth noting that the libnetwork/CNM API doesn't pass an expected interface name into the plugin, but does receive the interface name back from the plugin.  This would let the plugin itself choose whether to create additional interfaces in the namespace, or to set up the existing interface for the new network.  The interface name is then passed up to the runtime.

So I guess I'm wondering why CNI sends the expected interface down to the plugins at all?

aanm:
>It's worth noting that the libnetwork/CNM API doesn't pass an expected interface name into the plugin, but does receive the interface name back from the plugin. This would let the plugin itself choose whether to create additional interfaces in the namespace, or to set up the existing interface for the new network. The interface name is then passed up to the runtime.

Accordingly with the spec the only thing passed up to the runtime is the IPs assigned and the DNS info.
>Result:
IPs assigned to the interface. This is either an IPv4 address, an IPv6 address, or both.
DNS information. Dictionary that includes DNS information for nameservers, domain, search domains and options.

dcbw:
Yeah, I think that should get changed.  The runtime should pass back the interface name and all that, instead of accepting an interface name from the runtime.  Otherwise I can't see how multiple networks in a container can work for some cases, even with the same plugin.

steveeJ:
@dcbw 
> The runtime should pass back the interface name and all that, instead of accepting an interface name from the runtime.
I would rather not guess which "runtime" to replace for "plugin". If I had to I'd say the first one?

steveeJ:
Ping @dcbw

dcbw:
@steveeJ I'm working on a PR to pass the interface name back to the runtime.  It doesn't yet re-try interface creation with different names, but at least it gets the API ready.

Issue #75, opened 2015-10-30
spec: add ability for ADD to return arbitrary k/v data


eyakubovich:
Current design requires the same set of args/data to be passed to both the ADD and DEL commands. This is slightly tricky b/c the net conf (JSON) is often stored on disk. In that case, the container runtime needs to save a copy of it in case it gets changed by the user prior to DEL command execution.

Since the runtime must be saving data to disk, we can clean the API a bit by having the ADD command return a dictionary with arbitrary key/values for the runtime to persist. It would then pipe it back to DEL command. It is more general b/c it allows the plugin to not have to serialize state to disk itself (at least for a subset of cases).

Example ADD:
```
$plugin_type < /etc/cni/mynet.conf
{
    "ip": "1.2.3.4/24",
    "data": {
        "masqChain": "xyz123",
        "bridgeName": "abc"
    }
}
```

Followed by DEL:
```
$plugin_type <<EOF
{
    "masqChain": "xyz123",
    "bridgeName": "abc"
}
EOF
```

steveeJ:
It is a good idea to pass this state out to the caller, but I would like to keep it optional/configurable.

This might be related to #116, because if we allow every plugin to have a state we should at minimum offer to take care of the state as we already do with the host-local allocator. It looks like this is developing to be a generic pattern so that every plugin can use a store backend to serialize state to. If the intention is to pass the state back to the caller, a store plugin possibly called stdout would just write it to stdout, a plugin possibly called embed would embed it in the plugin's result. The file backend we have so far would be an option too of course.

If this sounds reasonable not only to me I'll prepare a PR.

RFC @squaremo @thockin @jonboulle @zachgersh @jonboulle

steveeJ:
@zachgersh @squaremo @jonboulle @philips 
awaiting your feedback

rosenhouse:
FWIW, I like the idea of this as an optional CNI feature, but I'd suggest that when the k/v's are passed back to the plugin on `DEL`, that they also be nested underneath a `data` field.  Otherwise I'd be concerned contaminating the top-level namespace.

squaremo:
> This might be related to #116, because if we allow every plugin to have a state we should at minimum offer to take care of the state as we already do with the host-local allocator

I don't see this as the same situation -- so I don't think that _just because_ host-local happens to put its state on disk, it should be a general mechanism.

I'm not sure I see the need for making this an item of configuration -- is it really up to the operator to decide what kind of backend store is used? Or is it up to the client application?

[...]

Issue #46, opened 2015-08-27
Evaluate port mapping


eyakubovich:
Some containers need to have port mapping between them and the host. The current stance of CNI is that it is outside the scope (and rkt handles port mapping itself). The reasoning behind this is that port mapping is not associated with any particular network but rather is "global" to the container. At the same time port mapping is implemented as a DNAT rule that maps to a _particular_ IP in the container and thus into a network. Therefore it may also make sense to make that a part of the plugin's responsibilities.

asridharan:
Do we have any further insights/discussions on this particular topic? This pretty important from the perspective of orchestration engines such as Mesos since an answer to this question would indicate whether we need to incorporate port mapping as an implementation into the container runt time or use interfaces to pass this information to the plugin. 

Ideally we would like this responsibility to be offloaded to the plugin, reason being that there seems to be cases where the underlying network would like to perform the port mapping instead of relying on the linux kernel. One specific use-case that has come up in the Mesos community discussions is the case where  the data-plane being used is OVS instead of the kernel. In this case setting up iptable rules to perform DNAT might actually be deterimental to the underlying forwarding plane. 

One thought I had was to have port mapping specific plugin, and use "delegation" to allow plugins to offload port mapping to this plugin". Plugins that rely on the underlying data-plane to perform the port-mapping can simply avoid this delegation.

rosenhouse:
Your analysis makes sense.  When this has been discussed in the past, the conversation has gone from "port mapping" to "iptables in general" to "dynamic filter rules" to "maybe CNI needs some kind of dynamic UPDATE like action". 

There's certainly nothing stopping you from building a plugin that sets up port mappings, given some particular NetConf representation of the desired mappings.

I'd question whether delegation is the right way to invoke that plugin, since it requires code changes in the caller.  Instead, it might make sense for the container runtime to just invoke two plugins in a row, passing the output of the first one as input to the second one.

But more discussion or draft PRs would be interesting.

References:
- https://github.com/containernetworking/cni/issues/89#issuecomment-216910419
- https://github.com/containernetworking/cni/issues/151

asridharan:
@rosenhouse is there a way for operators to specify serialization of plugins in the CNI config? Didn't see any documentation on this? 

I am referring to your comment about "container runtime invoking multiple plugins". Currently Mesos doesn't do this. So wondering if this actually a thing?

rosenhouse:
[...]

PR #145, opened 2016-03-02
Pass interface name and MAC back to runtime

dcbw:
Add the infrastructure to pass interface names back from the plugin to the runtime.  This lays the groundwork for letting runtimes not pass CNI_IFNAME and allowing plugins to determine that themsevles.  This should be backwards compatible; if the runtime sends CNI_IFNAME then plugins must honor that and return an error if they cannot.  If the runtime does not send CNI_IFNAME, the plugins currently default to 'eth0'.

Example output:

```
{
    "interface": {
        "hostIfname": "veth511119b6",
        "hostMac": "56:95:eb:4c:a0:91",
        "containerIfname": "eth0",
        "containerMac": "6a:1f:52:95:86:2c"
    },
    "ip4": {
        "ip": "10.1.0.152/24",
        "gateway": "10.1.0.1",
        "routes": [
            {
                "dst": "0.0.0.0/0"
            }
        ]
    },
    "dns": {}
```

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1,7 +1,7 @@
 # Container Networking Interface Proposal
 
 ## Version
-This is CNI **spec** version **0.2.0**.
+This is CNI **spec** version **0.3.0**.
 
 Note that this is **independent from the version of the CNI library and plugins** in this repository (e.g. the versions of [releases](https://github.com/containernetworking/cni/releases)).
 
@@ -57,7 +57,8 @@ The operations that the CNI plugin needs to support are:
     - **Extra arguments**. This provides an alternative mechanism to allow simple configuration of CNI plugins on a per-container basis.
     - **Name of the interface inside the container**. This is the name that should be assigned to the interface created inside the container (network namespace); consequently it must comply with the standard Linux restrictions on interface names.
   - Result:
-    - **IPs assigned to the interface**. This is either an IPv4 address, an IPv6 address, or both.
+    - **Interfaces list**. Depending on the plugin, this can include the sandbox (eg, container or hypervisor) interface name and/or the host interface name, the hardware addresses of each interface, and details about the sandbox (if any) the interface is in.
+    - **IP configuration assigned to each interface**. The IPv4 and/or IPv6 addresses, gateways, and routes assigned to sandbox and/or host interfaces.
     - **DNS information**. Dictionary that includes DNS information for nameservers, domain, search domains and options.
 
 - Delete container from network
@@ -75,8 +76,8 @@ The operations that the CNI plugin needs to support are:
 
       ```
       {
-        "cniVersion": "0.2.0", // the version of the CNI spec in use for this output
-        "supportedVersions": [ "0.1.0", "0.2.0" ] // the list of CNI spec versions that this plugin supports
+        "cniVersion": "0.3.0", // the version of the CNI spec in use for this output
+        "supportedVersions": [ "0.1.0", "0.2.0", "0.3.0" ] // the list of CNI spec versions that this plugin supports
       }
       ```
 
@@ -85,7 +86,7 @@ It will then look for this executable in a list of predefined directories. Once 
 - `CNI_COMMAND`: indicates the desired operation; `ADD`, `DEL` or `VERSION`.
 - `CNI_CONTAINERID`: Container ID
 - `CNI_NETNS`: Path to network namespace file
-- `CNI_IFNAME`: Interface name to set up
+- `CNI_IFNAME`: Interface name to set up; plugin must honor this interface name or return an error
 - `CNI_ARGS`: Extra arguments passed in by the user at invocation time. Alphanumeric key-value pairs separated by semicolons; for example, "FOO=BAR;ABC=123"
 - `CNI_PATH`: Colon-separated list of paths to search for CNI plugin executables
 
@@ -94,33 +95,60 @@ Network configuration in JSON format is streamed to the plugin through stdin. Th
 
 ### Result
 
-Success is indicated by a return code of zero and the following JSON printed to stdout in the case of the ADD command. This should be the same output as was returned by the IPAM plugin (see [IP Allocation](#ip-allocation) for details).
+Note that IPAM plugins return an abbreviated `Result` structure as described in [IP Allocation](#ip-allocation).
+
+Success is indicated by a return code of zero and the following JSON printed to stdout in the case of the ADD command. The `ip` and `dns` items should be the same output as was returned by the IPAM plugin (see [IP Allocation](#ip-allocation) for details) except that the plugin should fill in the `interface` indexes appropriately, which are missing from IPAM plugin output since IPAM plugins should be unaware of interfaces.
 
 ```
 {
-  "cniVersion": "0.2.0",
-  "ip4": {
-    "ip": <ipv4-and-subnet-in-CIDR>,
-    "gateway": <ipv4-of-the-gateway>,  (optional)
-    "routes": <list-of-ipv4-routes>    (optional)
-  },
-  "ip6": {
-    "ip": <ipv6-and-subnet-in-CIDR>,
-    "gateway": <ipv6-of-the-gateway>,  (optional)
-    "routes": <list-of-ipv6-routes>    (optional)
-  },
+  "cniVersion": "0.3.0",
+  "interfaces": [                                            (this key omitted by IPAM plugins)
+      {
+          "name": "<name>",
+          "mac": "<MAC address>",                            (required if L2 addresses are meaningful)
+          "sandbox": "<netns path or hypervisor identifier>" (required for container/hypervisor interfaces, empty/omitted for host interfaces)
+      }
+  ],
+  "ip": [
+      {
+          "version": "<4-or-6>",
+          "address": "<ip-and-prefix-in-CIDR>",
+          "gateway": "<ip-address-of-the-gateway>",          (optional)
+          "interface": <numeric index into 'interfaces' list>
+      },
+      ...
+  ],
+  "routes": [                                                (optional)
+      {
+          "dst": "<ip-and-prefix-in-cidr>",
+          "gw": "<ip-of-next-hop>"                           (optional)
+      },
+      ...
+  ]
   "dns": {
-    "nameservers": <list-of-nameservers>           (optional)
-    "domain": <name-of-local-domain>               (optional)
-    "search": <list-of-additional-search-domains>  (optional)
-    "options": <list-of-options>                   (optional)
+    "nameservers": <list-of-nameservers>                     (optional)
+    "domain": <name-of-local-domain>                         (optional)
+    "search": <list-of-additional-search-domains>            (optional)
+    "options": <list-of-options>                             (optional)
   }
 }
 ```
 
 `cniVersion` specifies a [Semantic Version 2.0](http://semver.org) of CNI specification used by the plugin.
-`dns` field contains a dictionary consisting of common DNS information that this network is aware of.
-The result is returned in the same format as specified in the [configuration](#network-configuration).
+`interfaces` describes specific network interfaces the plugin created.
+If the `CNI_IFNAME` variable exists the plugin must use that name for the sandbox/hypervisor interface or return an error if it cannot.
+- `mac` (string): the hardware address of the interface.
+   If L2 addresses are not meaningful for the plugin then this field is optional.
+- `sandbox` (string): container/namespace-based environments should return the full filesystem path to the network namespace of that sandbox.
+   Hypervisor/VM-based plugins should return an ID unique to the virtualized sandbox the interface was created in.
+   This item must be provided for interfaces created or moved into a sandbox like a network namespace or a hypervisor/VM.
+
+The `ip` field is a list of IP configuration information.
+See the [IP well-known structure](#ip) section for more information.
+
+The `dns` field contains a dictionary consisting of common DNS information.
+See the [DNS well-known structure](#dns) section for more information.
+
 The specification does not declare how this information must be processed by CNI consumers.
 Examples include generating an `/etc/resolv.conf` file to be injected into the container filesystem or running a DNS forwarder on the host.
 
@@ -164,7 +192,7 @@ Plugins may define additional fields that they accept and may generate an error 
 
 ```json
 {
-  "cniVersion": "0.2.0",
+  "cniVersion": "0.3.0",
   "name": "dbnet",
   "type": "bridge",
   // type (plugin) specific
@@ -183,7 +211,7 @@ Plugins may define additional fields that they accept and may generate an error 
 
 ```json
 {
-  "cniVersion": "0.2.0",
+  "cniVersion": "0.3.0",
   "name": "pci",
   "type": "ovs",
   // type (plugin) specific
@@ -204,7 +232,7 @@ Plugins may define additional fields that they accept and may generate an error 
 
 ```json
 {
-  "cniVersion": "0.1",
+  "cniVersion": "0.3.0",
   "name": "wan",
   "type": "macvlan",
   // ipam specific
@@ -394,41 +422,40 @@ Success is indicated by a zero return code and the following JSON being printed 
 
 ```
 {
-  "cniVersion": "0.2.0",
-  "ip4": {
-    "ip": <ipv4-and-subnet-in-CIDR>,
-    "gateway": <ipv4-of-the-gateway>,  (optional)
-    "routes": <list-of-ipv4-routes>    (optional)
-  },
-  "ip6": {
-    "ip": <ipv6-and-subnet-in-CIDR>,
-    "gateway": <ipv6-of-the-gateway>,  (optional)
-    "routes": <list-of-ipv6-routes>    (optional)
-  },
+  "cniVersion": "0.3.0",
+  "ips": [
+      {
+          "version": "<4-or-6>",
+          "address": "<ip-and-prefix-in-CIDR>",
+          "gateway": "<ip-address-of-the-gateway>"  (optional)
+      },
+      ...
+  ],
+  "routes": [                                       (optional)
+      {
+          "dst": "<ip-and-prefix-in-cidr>",
+          "gw": "<ip-of-next-hop>"                  (optional)
+      },
+      ...
+  ]
   "dns": {
-    "nameservers": <list-of-nameservers>           (optional)
-    "domain": <name-of-local-domain>               (optional)
-    "search": <list-of-search-domains>             (optional)
-    "options": <list-of-options>                   (optional)
+    "nameservers": <list-of-nameservers>            (optional)
+    "domain": <name-of-local-domain>                (optional)
+    "search": <list-of-search-domains>              (optional)
+    "options": <list-of-options>                    (optional)
   }
 }
 ```
 
+Note that unlike regular CNI plugins, IPAM plugins return an abbreviated `Result` structure that does not include the `interfaces` key, since IPAM plugins should be unaware of interfaces configured by their parent plugin except those specifically required for IPAM (eg, like the `dhcp` IPAM plugin).
+
 `cniVersion` specifies a [Semantic Version 2.0](http://semver.org) of CNI specification used by the plugin.
-`gateway` is the default gateway for this subnet, if one exists.
-It does not instruct the CNI plugin to add any routes with this gateway: routes to add are specified separately via the `routes` field.
-An example use of this value is for the CNI plugin to add this IP address to the linux-bridge to make it a gateway.
 
-Each route entry is a dictionary with the following fields:
-- `dst` (string): Destination subnet specified in CIDR notation.
-- `gw` (string): IP of the gateway. If omitted, a default gateway is assumed (as determined by the CNI plugin).
+The `ips` field is a list of IP configuration information.
+See the [IP well-known structure](#ip) section for more information.
 
-The "dns" field contains a dictionary consisting of common DNS information. 
-- `nameservers` (list of strings): list of a priority-ordered list of DNS nameservers that this network is aware of. Each entry in the list is a string containing either an IPv4 or an IPv6 address.
-- `domain` (string): the local domain used for short hostname lookups.
-- `search` (list of strings): list of priority ordered search domains for short hostname lookups. Will be preferred over `domain` by most resolvers.
-- `options` (list of strings): list of options that can be passed to the resolver
-See [CNI Plugin Result](#result) section for more information.
+The `dns` field contains a dictionary consisting of common DNS information.
+See the [DNS well-known structure](#dns) section for more information.
 
 Errors and logs are communicated in the same way as the CNI plugin. See [CNI Plugin Result](#result) section for details.
 
@@ -439,6 +466,68 @@ IPAM plugin examples:
 #### Notes
  - Routes are expected to be added with a 0 metric.
  - A default route may be specified via "0.0.0.0/0". Since another network might have already configured the default route, the CNI plugin should be prepared to skip over its default route definition.
+
+### Well-known Structures
+
+#### IP
+
+```
+  "ips": [
+      {
+          "version": "<4-or-6>",
+          "address": "<ip-and-prefix-in-CIDR>",
+          "gateway": "<ip-address-of-the-gateway>",      (optional)
+          "interface": <numeric index into 'interfaces' list> (not required for IPAM plugins)
+      },
+      ...
+  ]
+```
+
+The `ip` field is a list of IP configuration information determined by the plugin. Each item is a dictionary describing of IP configuration for a network interface.
+IP configuration for multiple network interfaces and multiple IP configurations for a single interface may be returned as separate items in the `ip` list.
+All properties known to the plugin should be provided, even if not strictly required.
+- `version` (string): either "4" or "6" and corresponds to the IP version of the addresses in the entry.
+   All IP addresses and gateways provided must be valid for the given `version`.
+- `address` (string): an IP address in CIDR notation (eg "192.168.1.3/24").
+- `gateway` (string): the default gateway for this subnet, if one exists.
+   It does not instruct the CNI plugin to add any routes with this gateway: routes to add are specified separately via the `routes` field.
+   An example use of this value is for the CNI `bridge` plugin to add this IP address to the Linux bridge to make it a gateway.
+- `interface` (uint): the index into the `interfaces` list for a [CNI Plugin Result](#result) indicating which interface this IP configuration should be applied to.
+   IPAM plugins should not return this key since they have no information about network interfaces.
+
+#### Routes
+
+```
+  "routes": [
+      {
+          "dst": "<ip-and-prefix-in-cidr>",
+          "gw": "<ip-of-next-hop>"               (optional)
+      },
+      ...
+  ]
+```
+
+- Each `routes` entry is a dictionary with the following fields.  All IP addresses in the `routes` entry must be the same IP version, either 4 or 6.
+  - `dst` (string): destination subnet specified in CIDR notation.
+  - `gw` (string): IP of the gateway. If omitted, a default gateway is assumed (as determined by the CNI plugin).
+
+#### DNS
+
+```
+  "dns": {
+    "nameservers": <list-of-nameservers>                 (optional)
+    "domain": <name-of-local-domain>                     (optional)
+    "search": <list-of-additional-search-domains>        (optional)
+    "options": <list-of-options>                         (optional)
+  }
+```
+
+The `dns` field contains a dictionary consisting of common DNS information.
+- `nameservers` (list of strings): list of a priority-ordered list of DNS nameservers that this network is aware of. Each entry in the list is a string containing either an IPv4 or an IPv6 address.
+- `domain` (string): the local domain used for short hostname lookups.
+- `search` (list of strings): list of priority ordered search domains for short hostname lookups. Will be preferred over `domain` by most resolvers.
+- `options` (list of strings): list of options that can be passed to the resolver.
+  See [CNI Plugin Result](#result) section for more information.
 
 ## Well-known Error Codes
 - `1` - Incompatible CNI version
--- a/libcni/api_test.go
+++ b/libcni/api_test.go
@@ -58,7 +58,7 @@ func newPluginInfo(configKey, configValue, prevResult string, injectDebugFilePat
 	}
 	Expect(debug.WriteDebug(debugFilePath)).To(Succeed())
 
-	config := fmt.Sprintf(`{"type": "noop", "%s": "%s", "cniVersion": "0.2.0"`, configKey, configValue)
+	config := fmt.Sprintf(`{"type": "noop", "%s": "%s", "cniVersion": "0.3.0"`, configKey, configValue)
 	if prevResult != "" {
 		config += fmt.Sprintf(`, "prevResult": %s`, prevResult)
 	}
@@ -77,8 +77,10 @@ func newPluginInfo(configKey, configValue, prevResult string, injectDebugFilePat
 var _ = Describe("Invoking plugins", func() {
 	Describe("Invoking a single plugin", func() {
 		var (
-			plugin        pluginInfo
+			debugFilePath string
+			debug         *noop_debug.Debug
 			cniBinPath    string
+			pluginConfig  string
 			cniConfig     libcni.CNIConfig
 			netConfig     *libcni.NetworkConfig
 			runtimeConfig *libcni.RuntimeConf
@@ -87,31 +89,39 @@ var _ = Describe("Invoking plugins", func() {
 		)
 
 		BeforeEach(func() {
-			pluginResult := `{ "ip4": { "ip": "10.1.2.3/24" }, "dns": {} }`
-			plugin = newPluginInfo("some-key", "some-value", "", false, pluginResult)
+			debugFile, err := ioutil.TempFile("", "cni_debug")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(debugFile.Close()).To(Succeed())
+			debugFilePath = debugFile.Name()
+
+			debug = &noop_debug.Debug{
+				ReportResult: `{ "ips": [{ "version": "4", "address": "10.1.2.3/24" }], "dns": {} }`,
+			}
+			Expect(debug.WriteDebug(debugFilePath)).To(Succeed())
 
 			cniBinPath = filepath.Dir(pluginPaths["noop"])
+			pluginConfig = `{ "type": "noop", "some-key": "some-value", "cniVersion": "0.3.0" }`
 			cniConfig = libcni.CNIConfig{Path: []string{cniBinPath}}
 			netConfig = &libcni.NetworkConfig{
 				Network: &types.NetConf{
 					Type: "noop",
 				},
-				Bytes: []byte(plugin.config),
+				Bytes: []byte(pluginConfig),
 			}
 			runtimeConfig = &libcni.RuntimeConf{
 				ContainerID: "some-container-id",
 				NetNS:       "/some/netns/path",
 				IfName:      "some-eth0",
-				Args:        [][2]string{{"DEBUG", plugin.debugFilePath}},
+				Args:        [][2]string{[2]string{"DEBUG", debugFilePath}},
 			}
 
 			expectedCmdArgs = skel.CmdArgs{
 				ContainerID: "some-container-id",
 				Netns:       "/some/netns/path",
 				IfName:      "some-eth0",
-				Args:        "DEBUG=" + plugin.debugFilePath,
+				Args:        "DEBUG=" + debugFilePath,
 				Path:        cniBinPath,
-				StdinData:   []byte(plugin.config),
+				StdinData:   []byte(pluginConfig),
 			}
 		})
 
@@ -124,15 +134,18 @@ var _ = Describe("Invoking plugins", func() {
 				Expect(err).NotTo(HaveOccurred())
 
 				Expect(result).To(Equal(&current.Result{
-					IP4: &current.IPConfig{
-						IP: net.IPNet{
-							IP:   net.ParseIP("10.1.2.3"),
-							Mask: net.IPv4Mask(255, 255, 255, 0),
+					IPs: []*current.IPConfig{
+						{
+							Version: "4",
+							Address: net.IPNet{
+								IP:   net.ParseIP("10.1.2.3"),
+								Mask: net.IPv4Mask(255, 255, 255, 0),
+							},
 						},
 					},
 				}))
 
-				debug, err := noop_debug.ReadDebug(plugin.debugFilePath)
+				debug, err := noop_debug.ReadDebug(debugFilePath)
 				Expect(err).NotTo(HaveOccurred())
 				Expect(debug.Command).To(Equal("ADD"))
 				Expect(debug.CmdArgs).To(Equal(expectedCmdArgs))
@@ -151,8 +164,8 @@ var _ = Describe("Invoking plugins", func() {
 
 			Context("when the plugin errors", func() {
 				BeforeEach(func() {
-					plugin.debug.ReportError = "plugin error: banana"
-					Expect(plugin.debug.WriteDebug(plugin.debugFilePath)).To(Succeed())
+					debug.ReportError = "plugin error: banana"
+					Expect(debug.WriteDebug(debugFilePath)).To(Succeed())
 				})
 				It("unmarshals and returns the error", func() {
 					result, err := cniConfig.AddNetwork(netConfig, runtimeConfig)
@@ -167,7 +180,7 @@ var _ = Describe("Invoking plugins", func() {
 				err := cniConfig.DelNetwork(netConfig, runtimeConfig)
 				Expect(err).NotTo(HaveOccurred())
 
-				debug, err := noop_debug.ReadDebug(plugin.debugFilePath)
+				debug, err := noop_debug.ReadDebug(debugFilePath)
 				Expect(err).NotTo(HaveOccurred())
 				Expect(debug.Command).To(Equal("DEL"))
 				Expect(debug.CmdArgs).To(Equal(expectedCmdArgs))
@@ -186,8 +199,8 @@ var _ = Describe("Invoking plugins", func() {
 
 			Context("when the plugin errors", func() {
 				BeforeEach(func() {
-					plugin.debug.ReportError = "plugin error: banana"
-					Expect(plugin.debug.WriteDebug(plugin.debugFilePath)).To(Succeed())
+					debug.ReportError = "plugin error: banana"
+					Expect(debug.WriteDebug(debugFilePath)).To(Succeed())
 				})
 				It("unmarshals and returns the error", func() {
 					err := cniConfig.DelNetwork(netConfig, runtimeConfig)
@@ -203,7 +216,7 @@ var _ = Describe("Invoking plugins", func() {
 
 				Expect(versionInfo).NotTo(BeNil())
 				Expect(versionInfo.SupportedVersions()).To(Equal([]string{
-					"0.-42.0", "0.1.0", "0.2.0",
+					"0.-42.0", "0.1.0", "0.2.0", "0.3.0",
 				}))
 			})
 
@@ -229,13 +242,13 @@ var _ = Describe("Invoking plugins", func() {
 
 		BeforeEach(func() {
 			plugins = make([]pluginInfo, 3, 3)
-			plugins[0] = newPluginInfo("some-key", "some-value", "", true, `{"dns":{},"ip4":{"ip": "10.1.2.3/24"}}`)
-			plugins[1] = newPluginInfo("some-key", "some-other-value", `{"dns":{},"ip4":{"ip": "10.1.2.3/24"}}`, true, "PASSTHROUGH")
-			plugins[2] = newPluginInfo("some-key", "yet-another-value", `{"dns":{},"ip4":{"ip": "10.1.2.3/24"}}`, true, "INJECT-DNS")
+			plugins[0] = newPluginInfo("some-key", "some-value", "", true, `{"dns":{},"ips":[{"version": "4", "address": "10.1.2.3/24"}]}`)
+			plugins[1] = newPluginInfo("some-key", "some-other-value", `{"dns":{},"ips":[{"version": "4", "address": "10.1.2.3/24"}]}`, true, "PASSTHROUGH")
+			plugins[2] = newPluginInfo("some-key", "yet-another-value", `{"dns":{},"ips":[{"version": "4", "address": "10.1.2.3/24"}]}`, true, "INJECT-DNS")
 
 			configList := []byte(fmt.Sprintf(`{
   "name": "some-list",
-  "cniVersion": "0.2.0",
+  "cniVersion": "0.3.0",
   "plugins": [
     %s,
     %s,
@@ -275,10 +288,13 @@ var _ = Describe("Invoking plugins", func() {
 
 				Expect(result).To(Equal(&current.Result{
 					// IP4 added by first plugin
-					IP4: &current.IPConfig{
-						IP: net.IPNet{
-							IP:   net.ParseIP("10.1.2.3"),
-							Mask: net.IPv4Mask(255, 255, 255, 0),
+					IPs: []*current.IPConfig{
+						{
+							Version: "4",
+							Address: net.IPNet{
+								IP:   net.ParseIP("10.1.2.3"),
+								Mask: net.IPv4Mask(255, 255, 255, 0),
+							},
 						},
 					},
 					// DNS injected by last plugin
--- a/libcni/api_test.go
+++ b/libcni/api_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/containernetworking/cni/libcni"
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
 	noop_debug "github.com/containernetworking/cni/plugins/test/noop/debug"
 
 	. "github.com/onsi/ginkgo"
@@ -116,11 +117,14 @@ var _ = Describe("Invoking plugins", func() {
 
 		Describe("AddNetwork", func() {
 			It("executes the plugin with command ADD", func() {
-				result, err := cniConfig.AddNetwork(netConfig, runtimeConfig)
+				r, err := cniConfig.AddNetwork(netConfig, runtimeConfig)
 				Expect(err).NotTo(HaveOccurred())
 
-				Expect(result).To(Equal(&types.Result{
-					IP4: &types.IPConfig{
+				result, err := current.GetResult(r)
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(result).To(Equal(&current.Result{
+					IP4: &current.IPConfig{
 						IP: net.IPNet{
 							IP:   net.ParseIP("10.1.2.3"),
 							Mask: net.IPv4Mask(255, 255, 255, 0),
@@ -263,12 +267,15 @@ var _ = Describe("Invoking plugins", func() {
 
 		Describe("AddNetworkList", func() {
 			It("executes all plugins with command ADD and returns an intermediate result", func() {
-				result, err := cniConfig.AddNetworkList(netConfigList, runtimeConfig)
+				r, err := cniConfig.AddNetworkList(netConfigList, runtimeConfig)
 				Expect(err).NotTo(HaveOccurred())
 
-				Expect(result).To(Equal(&types.Result{
+				result, err := current.GetResult(r)
+				Expect(err).NotTo(HaveOccurred())
+
+				Expect(result).To(Equal(&current.Result{
 					// IP4 added by first plugin
-					IP4: &types.IPConfig{
+					IP4: &current.IPConfig{
 						IP: net.IPNet{
 							IP:   net.ParseIP("10.1.2.3"),
 							Mask: net.IPv4Mask(255, 255, 255, 0),
--- a/pkg/ip/link.go
+++ b/pkg/ip/link.go
@@ -90,6 +90,14 @@ func RandomVethName() (string, error) {
 	return fmt.Sprintf("veth%x", entropy), nil
 }
 
+func RenameLink(curName, newName string) error {
+	link, err := netlink.LinkByName(curName)
+	if err == nil {
+		err = netlink.LinkSetName(link, newName)
+	}
+	return err
+}
+
 // SetupVeth sets up a virtual ethernet link.
 // Should be in container netns, and will switch back to hostNS to set the host
 // veth end up.
--- a/pkg/ipam/ipam.go
+++ b/pkg/ipam/ipam.go
@@ -21,11 +21,12 @@ import (
 	"github.com/containernetworking/cni/pkg/invoke"
 	"github.com/containernetworking/cni/pkg/ip"
 	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/types/current"
 
 	"github.com/vishvananda/netlink"
 )
 
-func ExecAdd(plugin string, netconf []byte) (*types.Result, error) {
+func ExecAdd(plugin string, netconf []byte) (types.Result, error) {
 	return invoke.DelegateAdd(plugin, netconf)
 }
 
@@ -35,7 +36,7 @@ func ExecDel(plugin string, netconf []byte) error {
 
 // ConfigureIface takes the result of IPAM plugin and
 // applies to the ifName interface
-func ConfigureIface(ifName string, res *types.Result) error {
+func ConfigureIface(ifName string, res *current.Result) error {
 	link, err := netlink.LinkByName(ifName)
 	if err != nil {
 		return fmt.Errorf("failed to lookup %q: %v", ifName, err)
--- a/pkg/ipam/ipam.go
+++ b/pkg/ipam/ipam.go
@@ -16,6 +16,7 @@ package ipam
 
 import (
 	"fmt"
+	"net"
 	"os"
 
 	"github.com/containernetworking/cni/pkg/invoke"
@@ -37,6 +38,10 @@ func ExecDel(plugin string, netconf []byte) error {
 // ConfigureIface takes the result of IPAM plugin and
 // applies to the ifName interface
 func ConfigureIface(ifName string, res *current.Result) error {
+	if len(res.Interfaces) == 0 {
+		return fmt.Errorf("no interfaces to configure")
+	}
+
 	link, err := netlink.LinkByName(ifName)
 	if err != nil {
 		return fmt.Errorf("failed to lookup %q: %v", ifName, err)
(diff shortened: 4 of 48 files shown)


dcbw:
For example, with OpenShift we'd use the hostIfname so that we don't have to jump into the container with 'nsenter' and grab the peer ifindex, then look that up on the host side.  We also use nsenter to get the container veth MAC address, so this would also remove that step.

I'm going to plumb this through Kubernetes too, since it assumes interfaces are always called 'eth0' which may not always be the case, like with the Calico CNI plugin.

dcbw:
@steveeJ fixed the issue, PTAL

caseydavenport:
Looks useful :)

Wanted to chime in with a few thoughts:

- This is quite nice for CNI plugins which don't use multiple interfaces to implement multiple CNI networks (e.g Calico).  We should make sure the spec includes that orchestrators should allow for multiple CNI networks to return the same `interface` fields .

- This does put a bit more work on the backs of network plugins in terms of determining what interface they should use, but it isn't a whole lot of work and gives the plugin more flexibility which is nice.

@dcbw - Also wanted to correct the point above - the Calico CNI plugin **does** respect the CNI_IFNAME variable.  I believe you're thinking of Mike Spreitzer's experimentation with mapping his own CNI plugin for Kubernetes to the Calico libnetwork plugin (which doesn't respect CNI_IFNAME for obvious reasons).

dcbw:
@caseydavenport thanks for the clarification, and sorry for blaming the Calico plugin where I shouldn't have :)

zachgersh:
hey @dcbw would it be possible to include some tests of this functionality? We've started backfilling tests and it would be great to merge new code with tests.

If you need examples or want to chat about how to test this I'd be happy to help.

SPEC.md
```diff
@@ -100,11 +107,10 @@ Success is indicated by a return code of zero and the following JSON printed to
 }
 ```
 
-`cniVersion` specifies a [Semantic Version 2.0](http://semver.org) of CNI specification used by the plugin.
-`dns` field contains a dictionary consisting of common DNS information that this network is aware of.
-The result is returned in the same format as specified in the [configuration](#network-configuration).
-The specification does not declare how this information must be processed by CNI consumers.
-Examples include generating an `/etc/resolv.conf` file to be injected into the container filesystem or running a DNS forwarder on the host.
+- `cniVersion` specifies a [Semantic Version 2.0](http://semver.org) of CNI specification used by the plugin.
+- `dns` field contains a dictionary consisting of common DNS information that this network is aware of. The result is returned in the same format as specified in the [configuration](#network-configuration). The specification does not declare how this information must be processed by CNI consumers. Examples include generating an `/etc/resolv.conf` file to be injected into the container filesystem or running a DNS forwarder on the host.
```
steveeJ: Did you reformat this section intentionally? We've been writing one sentence (separated by `.`) per line in markdown files.

dcbw:
@steveeJ unintentional reformatting, I'll fix and repush with some testcases.

steveeJ:
> @steveeJ unintentional reformatting, I'll fix and repush with some testcases.

Also, please make sure the commit message conforms to https://github.com/appc/cni/blob/master/CONTRIBUTING.md#format-of-the-commit-message. We go back to insisting on this, as `git log` is starting to get out of shape :-| I wonder if I can get some bot to do this for us.

dcbw:
@steveeJ repushed with the SPEC.md reformat fix.  Also, instead of hardcoding "eth0" I have updated all plugins to use a prefix+index method and retry with an incremented index if the interface creation fails with EEXIST.  This should make all plugins handle multiple interfaces in the same container as long as the runtime doesn't pass CNI_IFNAME.

I have *not* added testcases yet, so this isn't ready to merge.  But I wanted to get your opinion on this updated approach.

dcbw:
> Also, please make sure the commit message conforms to https://github.com/appc/cni/blob/master/CONTRIBUTING.md#format-of-the-commit-message. We should go back to insisting on this, as git log is starting to get out of shape :-| I wonder if I can get some bot to do this for us.

Also updated the git commit message, is that better?

dcbw:
Rebased on top of https://github.com/appc/cni/pull/167 with additional tests.

dcbw:
@steveeJ PTAL, thanks!

steveeJ:
Thanks @dcbw, and sorry for the inconvenience but it needs another rebase for the tests, the travis configuration has changed.

dcbw:
@steveeJ rebased, PTAL thanks!

dcbw:
@steveeJ also could we remove the needs-tests label?  I've added those a while back.  Thanks!

xiangpengzhao:
This is absolutely useful for me:)
In my env, I do this in my CNI plugin: move an interface on the host to a pod namespace(also rename this interface to "eth0" in pod namespace). After the pod being deleted, its interface returns to host namespace, with the name "eth0". Then this "eth0" interface conflicts with the former "eth0" interface on the host.

xiangpengzhao:
@dcbw , have you already had any plan about this through kubernetes ? Thanks!
One idea:
Pass a new flag `default-interface-name` to kubelet. Kubelet will pass this flag value as `CNI_IFNAME ` to CNI plugins. If the flag is not set, plugins will determine that themsevles, such as default to `eth0`, or any other name. 

But this has one problem, how to get POD IP when flag is not set? Currently, kubernetes get the POD IP in func [GetPodNetworkStatus](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/network/cni/cni.go#L140) by calling `runtime.GetContainerIP(id.ID, network.DefaultInterfaceName)`. In this case, `DefaultInterfaceName` is empty.

dcbw:
@xiangpengzhao yes, I've got a plan for kubernetes.  Ideally, the plugin itself would know what its preferred interface name would be and would be able to retrieve that interface name.  Perhaps it could use a combination of it's own plugin name and the container ID instead of something entirely generic and potentially conflicting (like "eth0").

In any case, there are also discussions going on about whether kubelet should save returned plugin state, including the IP address and probably interface name, so that kubelet would pass it back into the plugin for the DEL action.

xiangpengzhao:
I read through this issue again and find that I had a bit misunderstanding in my last comment. Sorry about that. :sweat_smile: 

From kubernetes perspective, is our final goal that kubelet does not need to pass CNI_IFNAME to the plugin ? If so, I think kubelet should only save the IP address to determine POD IP. It don't need to care about what the interface name is. It may be the responsibility of plugin to save enough information (in etcd or else) for the DEL action.

May I know when will you implement you plan for kubernetes? Thanks! @dcbw

aanm:
Kindly related to https://github.com/containernetworking/cni/issues/112

pkg/ip/link.go
```diff
@@ -66,27 +73,76 @@ func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err err
...
 	}
 
-	// NetworkManager (recent versions) will ignore veth devices that start with "veth"
-	return fmt.Sprintf("veth%x", entropy), nil
+	// NetworkManager (1.0+) will ignore externally created virtual interfaces
+	return fmt.Sprintf("%s%x", prefix, entropy), nil
+}
+
+func RenameLink(curName, newName, newPrefix string) (string, error) {
+	if len(newName) == 0 && len(newPrefix) == 0 {
```
steveeJ: Why use `len()` instead of comparing to `""`?

pkg/ip/link.go
```diff
@@ -66,27 +73,76 @@ func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err err
...
+		return "", fmt.Errorf("one of new name or new name prefix required")
+	}
+
+	link, err := netlink.LinkByName(curName)
+	if err != nil {
+		return "", err
+	}
+
+	if len(newName) > 0 {
```
steveeJ: `newName != ""` ?

pkg/ip/link.go
```diff
@@ -66,27 +73,76 @@ func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err err
...
+		if err := netlink.LinkSetName(link, newName); err != nil {
+			return "", err
+		}
+		return newName, nil
+	}
+
+	for idx := 0; idx < 100; idx++ {
```
steveeJ: Is there a limit of interfaces? If it happens to be 100 we should still put it in a constant. Otherwise, we should use the real limit.

pkg/ip/link.go
```diff
@@ -66,27 +73,76 @@ func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err err
...
 
-// RandomVethName returns string "veth" with random prefix (hashed from entropy)
-func RandomVethName() (string, error) {
-	entropy := make([]byte, 4)
+const ENTROPY_LEN = 4
+
+// RandomIfaceName returns string @prefix with random suffix (hashed from entropy)
+func RandomIfaceName(prefix string) (string, error) {
+	if len(prefix)+ENTROPY_LEN > syscall.IFNAMSIZ {
+		return "", fmt.Errorf("prefix %s too long", len(prefix))
```
steveeJ: if we can't predict the maximum prefix length statically and document it, we should at least give it in the error message.

pkg/ip/link.go
```diff
@@ -66,27 +73,76 @@ func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err err
...
+	for idx := 0; idx < 100; idx++ {
+		newName = fmt.Sprintf(newPrefix, idx)
+		err = netlink.LinkSetName(link, newName)
+		if err == nil {
+			return newName, nil
+		} else if !os.IsExist(err) {
+			return "", err
+		}
+	}
+	return "", err
```
steveeJ: Am I correct that this will be reached if we had 100 conflicts? `err` will still be `nil` here, right?

steveeJ: Correction, it will be an `IsExist()`-type error. But the earlier mentioned magic 100 makes this weird. However, the error should indicate that the netns ran out of interfaces with an explicit message.

pkg/ip/link.go
```diff
@@ -66,27 +73,76 @@ func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err err
...
 }
 
-// RandomVethName returns string "veth" with random prefix (hashed from entropy)
-func RandomVethName() (string, error) {
-	entropy := make([]byte, 4)
+const ENTROPY_LEN = 4
+
+// RandomIfaceName returns string @prefix with random suffix (hashed from entropy)
+func RandomIfaceName(prefix string) (string, error) {
+	if len(prefix)+ENTROPY_LEN > syscall.IFNAMSIZ {
```
steveeJ: I'd like to discuss a counter-based naming scheme. Persisting the counter per-prefix should make this deterministic. We could calculate the maximum number of interfaces with `syscall.IFNAMSIZ - len(prefix)`.

pkg/ip/link.go
dcbw: It's an arbitrary limit and there could be more than 100.  I'll bump the limit and make it a constant.

dcbw:
@steveeJ fixed up and repushed, taking into account our conversation about interface rename loops.

steveeJ:
I want to push this back and follow-up an OOB discussion I hat with @tomdee that was initiated by [this post on the cni-dev ML](https://groups.google.com/forum/#!topic/cni-dev/d2h2fsdz_Qw).

steveeJ:
In my last comment I didn't make it clear that I'm assuming that all the newly added results fields should also be available as runtime configuration parameters. Also, clearly the environment variable `CNI_IFNAME` is not destined to carry all this information. IMHO, adding more environment variables or including this in `CNI_ARGS` is impractical as parsing is quite painful and not namespace-able. 

TLDR
@dcbw, I would prefer to solve #75 and use this issue's matter as first use-case.

dcbw:
@steveeJ sure, #75 is fine.  However, I think we also need to define standard k/v pairs along with #75 so that runtimes can depend on values plugins return.  That would take the place of the spec change in this PR.  Does that sound acceptable?

steveeJ:
@dcbw that sounds good assuming you meant _out of spec_.

pkg/ip/link.go
```diff
@@ -40,9 +41,15 @@ func makeVethPair(name, peer string, mtu int) (netlink.Link, error) {
 	return veth, nil
 }
 
-func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err error) {
+func makeVeth(namePrefix string, mtu int) (peerName string, veth netlink.Link, err error) {
 	for i := 0; i < 10; i++ {
-		peerName, err = RandomVethName()
+		peerName, err = RandomIfaceName(namePrefix)
```
steveeJ: Could we get rid of RandomIfaceName and replace it with a refactored version of https://github.com/containernetworking/cni/blob/f6135c7ba38fe82545322407a629d671e90b3ecc/pkg/utils/utils.go#L31?

SPEC.md
```diff
@@ -67,7 +68,7 @@ It will then look for this executable in a list of predefined directories. Once
 - `CNI_COMMAND`: indicates the desired operation; either `ADD` or `DEL`
 - `CNI_CONTAINERID`: Container ID
 - `CNI_NETNS`: Path to network namespace file
-- `CNI_IFNAME`: Interface name to set up
+- `CNI_IFNAME`: Interface name to set up; plugin must honor this interface name or return an error
```
steveeJ: The issue I'm about to raise is not caused by your change, but it's not clear which of the interfaces this applies to. This confusion bugs me and I'm tempted to suggest to remove the CNI_IFNAME variable completely.

steveeJ:
What do you think about listing the interfaces and information with it's containing network namespace?

SPEC.md
dcbw: > The issue I'm about to raise is not caused by your change, but it's not clear which of the interfaces this applies to. This confusion bugs me and I'm tempted to suggest to remove the CNI_IFNAME variable completely.

It has always applied to the interface name in the container's namespace, right?  Making that clearer in the docs would solve the confusion here.  I don't personally mind if we got rid of it as long as we pass the interface info back to the runtime though.

pkg/ip/link.go
dcbw: That would mean its no longer a random name though, since FormatChainName() just hashes the input.  We still need some kind of entropy.  If you want to call the same CNI plugin twice on the same container but get different interfaces, then without entropy you'd end up with the same interface name since there's nothing unique about multiple invocations of the plugin.

dcbw:
> What do you think about listing the interfaces and information with it's containing network namespace?

Wouldn't that be somewhat redundant?  The runtime already passes the namespace to the CNI plugin, and then the plugin would just be passing it back to the runtime.

steveeJ:
> Wouldn't that be somewhat redundant

In most cases where we act on the current (host) and the given netns you're definitely right. I've been tinkering around with ideas where we could create network namespaces from within the plugins and pass it back. I just want to make sure that we don't rush spec changes and try to be anticipatory.

pkg/ip/link.go
steveeJ: I see, have we agreed/disagreed on the counter-based indexing for interfaces? Could we just use the interfaces kernel ifindex? They should always be unique per netns.

dcbw:
> In most cases where we act on the current (host) and the given netns you're definitely right. I've been tinkering around with ideas where we could create network namespaces from within the plugins and pass it back. I just want to make sure that we don't rush spec changes and try to be anticipatory.

@steveeJ ah, good point.  It's intriguing, though unfortunately docker wouldn't support this case since it doesn't have the ability to run a container in an arbitrary network namespace.  I guess the container runtime that calls the CNI plugin would have to indicate whether the plugin should create the namespace or not (probably though CNI_NETNS="" or some other means), but I'd like to see how we can keep the runtime somewhat agnostic.  At least in Kubernetes-land, I don't want to have to special-case stuff in kubenet for docker vs. rkt.

steveeJ:
@dcbw Definitely this feature should be optional to remain compatible with environments where this is unsupported. For compatible environments this would definitely be a win as they wouldn't even have to use the `pkg/ns` functions anymore.

SPEC.md
```diff
@@ -106,6 +127,12 @@ Success is indicated by a return code of zero and the following JSON printed to
...
 The specification does not declare how this information must be processed by CNI consumers.
 Examples include generating an `/etc/resolv.conf` file to be injected into the container filesystem or running a DNS forwarder on the host.
+The `interfaces` describes specific network interfaces the plugin created and their IP configuration.
+If the plugin does not use L2 addressing and thus MAC addresses are not meaningful, then an interface element's "mac" key may be omitted.
+The `sandbox` key must be provided for interfaces created or moved into a sandbox like a network namespace or a hypervisor/VM.  For container/namespace-based environments the `sandbox` key should be the full filesystem path to the network namespace of that sandbox. Hypervisor/VM-based plugins should return an ID unique to the virtualized sandbox the interface was created in.
+If the `CNI_IFNAME` variable is passed to the plugin, it must honor that name or return an error if it cannot.
+Each interface in the `interfaces` list should also return IP addressing information if present.
+Plugins must also return duplicate IP address information for the sandboxed interface in the top-level `ip4` and `ip6` keys.
```
steveeJ: It needs to be specified which interfaces `ip4` and `ip6` belong to in regard to the ips lists embedded in the interfaces list.

SPEC.md
```diff
@@ -106,6 +127,12 @@ Success is indicated by a return code of zero and the following JSON printed to
 The result is returned in the same format as specified in the [configuration](#network-configuration).
 The specification does not declare how this information must be processed by CNI consumers.
 Examples include generating an `/etc/resolv.conf` file to be injected into the container filesystem or running a DNS forwarder on the host.
+The `interfaces` describes specific network interfaces the plugin created and their IP configuration.
+If the plugin does not use L2 addressing and thus MAC addresses are not meaningful, then an interface element's "mac" key may be omitted.
+The `sandbox` key must be provided for interfaces created or moved into a sandbox like a network namespace or a hypervisor/VM.  For container/namespace-based environments the `sandbox` key should be the full filesystem path to the network namespace of that sandbox. Hypervisor/VM-based plugins should return an ID unique to the virtualized sandbox the interface was created in.
```
steveeJ: The last word should probably be "for" instead of "in" since CNI will probably not actually act inside a VM but just prepare interfaces for it.

dcbw:
@steveeJ @tomdee @rosenhouse mega-update...  we kinda decided to do IPs-in-the-Interfaces-list in the maintainer's meeting today.  Unfortunately that makes life harder for the IPAM plugins.  Currently the just return the top-level 'ip4' field in the Result structure that is L3-only.

But IP-in-Interfaces-list removes the top-level 'ip4' and 'ip6' fields, and thus to get the IPAM info back they would have to create an Interfaces list that they don't (and shouldn't!) care about and don't have information for and wouldn't validate (since they don't know the interface name or anything like that, which is required).

So @steveeJ and I decided to jump back to an IPAM list with indexes into the Interfaces list. It's not the greatest solution, but it does appear to work and keeps the IPAM plugins fairly simple.

This PR has some downsides still:
(a) because IPAM is now list we need to deal with multiple addresses, so I had to change some code in other places to handle that
(b) those changes attempt to genericize IP address handling and thus this PR steps on some of the IPv6 PRs already outstanding
(c) no attempt is made to validate each IPConfig item to ensure the Version corresponds to the actual format of the IP address and gateway and stuff
(d) it passes the unit/integration tests, but some plugins don't have tests yet (PTP) and I haven't run this in production yet.

But in any case, I hope it pushes the Interfaces list and the IPs list discussion forward.

plugins/main/ptp/ptp.go
```diff
@@ -57,93 +57,114 @@ func setupContainerVeth(netns, ifName string, mtu int, pr *types.Result) (string
 	// "192.168.3.1/32 dev $ifName" and "192.168.3.0/24 via 192.168.3.1 dev $ifName".
 	// In other words we force all traffic to ARP via the gateway except for GW itself.
 
-	var hostVethName string
+	var hostInterface, containerInterface *types.Interface
+
 	err := ns.WithNetNSPath(netns, func(hostNS ns.NetNS) error {
-		hostVeth, _, err := ip.SetupVeth(ifName, mtu, hostNS)
+		hostVeth, contVeth, err := ip.SetupVeth(ifName, "eth%d", mtu, hostNS)
```
lukasredynk: Shouldn't it be: "eth" instead of "eth%d"?

plugins/main/ptp/ptp.go
```diff
@@ -57,93 +57,114 @@ func setupContainerVeth(netns, ifName string, mtu int, pr *types.Result) (string
 	// "192.168.3.1/32 dev $ifName" and "192.168.3.0/24 via 192.168.3.1 dev $ifName".
 	// In other words we force all traffic to ARP via the gateway except for GW itself.
 
-	var hostVethName string
+	var hostInterface, containerInterface *types.Interface
```
lukasredynk: hostInterface := &types.Interface{}
containerInterface := &types.Interface{}

plugins/main/ptp/ptp.go
jellonek: This is used as kernel template name so in this place this looks ok, but after adding random suffix it will look quite bad.
IMO there should be "eth" or "eth%d-".

dcbw: You're right; will fix.

plugins/main/ptp/ptp.go
dcbw: Fixed.

SPEC.md
```diff
@@ -77,21 +77,34 @@ Network configuration in JSON format is streamed to the plugin through stdin. Th
 
 ### Result
 
-Success is indicated by a return code of zero and the following JSON printed to stdout in the case of the ADD command. This should be the same output as was returned by the IPAM plugin (see [IP Allocation](#ip-allocation) for details).
+Success is indicated by a return code of zero and the following JSON printed to stdout in the case of the ADD command. The `ip4`, `ip6`, and `dns` items should be the same output as was returned by the IPAM plugin (see [IP Allocation](#ip-allocation) for details).
```
steveeJ: I assume it's a WIP but still noting that this is currently not consistent with the "ips" result list

SPEC.md
```diff
@@ -77,21 +77,34 @@ Network configuration in JSON format is streamed to the plugin through stdin. Th
...
+      {
+          "name": "<name>",
+          "mac": "<MAC address>",  (required if L2 addresses are meaningful)
+          "sandbox": "<netns path or hypervisor identifier>"   (required for container/hypervisor interfaces)
+      }
+  ],
+  "ips": [
+      {
+          "version": "<4-or-6>",
+          "ip": "<ip-and-prefix-in-CIDR>",
```
steveeJ: s/ip/address ?

SPEC.md
```diff
@@ -77,21 +77,34 @@ Network configuration in JSON format is streamed to the plugin through stdin. Th
...
-    "routes": <list-of-ipv6-routes>    (optional)
-  },
+  "interfaces": [
+      {
+          "name": "<name>",
+          "mac": "<MAC address>",  (required if L2 addresses are meaningful)
+          "sandbox": "<netns path or hypervisor identifier>"   (required for container/hypervisor interfaces)
+      }
+  ],
+  "ips": [
```
steveeJ: s/ips/ip

SPEC.md
steveeJ: s/ip/address

SPEC.md
dcbw: Fixed.

SPEC.md
dcbw: Fixed.

SPEC.md
dcbw: Fixed.

dcbw:
@steveeJ most recent push removed the new code that allowed runtimes to pass an empty CNI_IFNAME, just for simplicity and ease of review on this one.  I'll do a separate PR for that after this one lands.

dcbw:
@containernetworking/cni-maintainers review appreciated, thanks!

SPEC.md
```diff
@@ -50,7 +50,8 @@ The operations that the CNI plugin needs to support are:
     - **Extra arguments**. This provides an alternative mechanism to allow simple configuration of CNI plugins on a per-container basis.
     - **Name of the interface inside the container**. This is the name that should be assigned to the interface created inside the container (network namespace); consequently it must comply with the standard Linux restrictions on interface names.
   - Result:
-    - **IPs assigned to the interface**. This is either an IPv4 address, an IPv6 address, or both.
+    - **Interfaces list**. Depending on the plugin, this can include the sandbox (eg, container or hypervisor) interface name and/or the host interface name, the hardware addresses of each interface, and details about the sandbox (if any) the interface is in.
+    - **IPs assigned to each interface**. The IPv4 and/or IPv6 addresses, gateways, and routes assigned to sandbox and/or host interfaces.
```
steveeJ: s/IPs/IP configuration/

SPEC.md
```diff
@@ -102,7 +116,12 @@ Success is indicated by a return code of zero and the following JSON printed to
 ```
 
 `cniVersion` specifies a [Semantic Version 2.0](http://semver.org) of CNI specification used by the plugin.
-`dns` field contains a dictionary consisting of common DNS information that this network is aware of.
+`interfaces` describes specific network interfaces the plugin created.
+If L2 addresses are not meaningful for the plugin then an interface element's "mac" key may be omitted.
+The `sandbox` key must be provided for interfaces created or moved into a sandbox like a network namespace or a hypervisor/VM.  For container/namespace-based environments the `sandbox` key should be the full filesystem path to the network namespace of that sandbox. Hypervisor/VM-based plugins should return an ID unique to the virtualized sandbox the interface was created in.
+`ips` contains a list of IP address information for interfaces in the `interfaces` list. At least the `version` and `ip` items are required but all items should be provided if they are known to the plugin.
```
steveeJ: s/ips/ip

SPEC.md
```diff
@@ -217,16 +236,21 @@ Success is indicated by a zero return code and the following JSON being printed
...
-    "ip": <ipv4-and-subnet-in-CIDR>,
-    "gateway": <ipv4-of-the-gateway>,  (optional)
-    "routes": <list-of-ipv4-routes>    (optional)
-  },
-  "ip6": {
-    "ip": <ipv6-and-subnet-in-CIDR>,
-    "gateway": <ipv6-of-the-gateway>,  (optional)
-    "routes": <list-of-ipv6-routes>    (optional)
-  },
+  "ips": [
```
steveeJ: s/ips/ip

SPEC.md
```diff
@@ -217,16 +236,21 @@ Success is indicated by a zero return code and the following JSON being printed
...
+      {
+          "version": "<4-or-6>",
+          "ip": "<ip-and-prefix-in-CIDR>",
```
steveeJ: s/ip/address

SPEC.md
```diff
@@ -237,11 +261,14 @@ Success is indicated by a zero return code and the following JSON being printed
 ```
 
 `cniVersion` specifies a [Semantic Version 2.0](http://semver.org) of CNI specification used by the plugin.
+`ips` contains a list of IP address information determined by the plugin. At least the `version` and `ip` items are required but all items should be provided if they are known to the plugin.
```
steveeJ: s/ips/ip
s/IP address/IP configuration

SPEC.md
```diff
@@ -237,11 +261,14 @@ Success is indicated by a zero return code and the following JSON being printed
...
+`version` is either "4" or "6" and corresponds to the IP version of the addresses and routes in the entry. All addresses and routes provided must match the given `version`.
+`ip` is an IP address in CIDR notation (eg "192.168.1.3/24").
```
steveeJ: s/ip/address

dcbw:
@steveeJ reworked the SPEC, please double-check.  Thanks!

SPEC.md
```diff
@@ -217,40 +243,37 @@ Success is indicated by a zero return code and the following JSON being printed
 ```
 {
   "cniVersion": "0.1.0",
```
steveeJ: What did we agree on the CNI version bump? Can we implement this backwards compatible?

plugins/main/bridge/bridge.go
```diff
@@ -211,38 +216,48 @@ func cmdAdd(args *skel.CmdArgs) error {
...
-				result.IP4.Routes,
-				types.Route{Dst: *defaultNet, GW: result.IP4.Gateway},
-			)
+				ipc.Routes = append(
+					ipc.Routes,
+					types.Route{Dst: *defaultNet, GW: ipc.Gateway},
+				)
+			}
 
 			// TODO: IPV6
```
steveeJ: is something missing for IPv6 in this case?

plugins/main/bridge/bridge.go
```diff
@@ -211,38 +216,48 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 				}
-			}
 
-			result.IP4.Routes = append(
-				result.IP4.Routes,
-				types.Route{Dst: *defaultNet, GW: result.IP4.Gateway},
-			)
+				ipc.Routes = append(
+					ipc.Routes,
+					types.Route{Dst: *defaultNet, GW: ipc.Gateway},
```
steveeJ: Do we need to add an increasing metric here to make the default route deterministic?

plugins/main/bridge/bridge.go
```diff
@@ -211,38 +216,48 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
-	// TODO: make this optional when IPv6 is supported
-	if result.IP4 == nil {
-		return errors.New("IPAM plugin returned missing IPv4 config")
+	if len(result.IP) == 0 {
+		return errors.New("IPAM plugin returned missing IP config")
 	}
+	result.Interfaces = []*types.Interface{hostInterface, containerInterface}
```
steveeJ: IMO we should also add information about the bridge here

plugins/main/bridge/bridge.go
```diff
@@ -211,38 +216,48 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 
 	if err := netns.Do(func(_ ns.NetNS) error {
 		// set the default gateway if requested
 		if n.IsDefaultGW {
-			_, defaultNet, err := net.ParseCIDR("0.0.0.0/0")
-			if err != nil {
-				return err
-			}
+			for _, ipc := range result.IP {
+				defaultNet := &net.IPNet{
```
steveeJ: Readability and safety can be a little improved here:

```Go
defaultNet := &net.IPNet{}
switch {
    case ipc.Address.IP.To4() != nil && ipc.Address.IP.To6() == nil:
	    defaultNet.IP = net.IPv4zero
	    defaultNet.Mask = net.IPMask(net.IPv4zero)
    case ipc.Address.IP.To6() != nil && ipc.Address.IP.To4() == nil:
	    defaultNet.IP = net.IPv6zero
	    defaultNet.Mask = net.IPMask(net.IPv6zero)
    default:
        return fmt.Errorf("Unknown IP object: %v: ipc")
}
```

plugins/main/bridge/bridge_test.go
```diff
@@ -146,13 +146,16 @@ var _ = Describe("bridge Operations", func() {
...
+		var result *types.Result
 		err = originalNS.Do(func(ns.NetNS) error {
 			defer GinkgoRecover()
 
-			_, err := testutils.CmdAddWithResult(targetNs.Path(), IFNAME, func() error {
+			result, err = testutils.CmdAddWithResult(targetNs.Path(), IFNAME, func() error {
 				return cmdAdd(args)
 			})
 			Expect(err).NotTo(HaveOccurred())
+			Expect(len(result.Interfaces)).To(Equal(2))
```
steveeJ: If we add the bridge interface this might be 3

plugins/main/bridge/bridge_test.go
```diff
@@ -178,13 +181,10 @@ var _ = Describe("bridge Operations", func() {
...
 			Expect(len(links)).To(Equal(3)) // Bridge, veth, and loopback
-			for _, l := range links {
-				if l.Attrs().Name != BRNAME && l.Attrs().Name != "lo" {
-					_, isVeth := l.(*netlink.Veth)
-					Expect(isVeth).To(Equal(true))
-				}
-			}
+			link, err = netlink.LinkByName(result.Interfaces[0].Name)
 			Expect(err).NotTo(HaveOccurred())
+			_, isVeth := link.(*netlink.Veth)
```
steveeJ: gomega's matcher _BeAssignableToTypeOf(expected interface)_ can be used here. Roughly

```Go
Expect(link).To(BeAssignableToTypeOf(netlink.Veth))
```

I'm not sure about the pointer aspect here.

plugins/main/ptp/ptp.go
```diff
@@ -57,93 +57,115 @@ func setupContainerVeth(netns, ifName string, mtu int, pr *types.Result) (string
...
 				},
-				Scope: netlink.SCOPE_LINK,
-				Src:   pr.IP4.IP.IP,
-			},
-			netlink.Route{
-				LinkIndex: contVeth.Attrs().Index,
-				Dst: &net.IPNet{
-					IP:   pr.IP4.IP.IP.Mask(pr.IP4.IP.Mask),
-					Mask: pr.IP4.IP.Mask,
+				Scope: netlink.SCOPE_NOWHERE,
```
steveeJ: can you explain the scope change?

steveeJ:
Full review complete. This looks very good and it'll be exciting to merge these changes. One bigger question to discuss at this point is how do we handle the CNI version bump?

rosenhouse:
@steveeJ yes I'm still waiting for your review of #267.  Once that is merged I can resume work on the following steps in #266

steveeJ:
@rosenhouse are they dependent so our merge chain should be #267 -> #266 -> #145?

SPEC.md
```diff
@@ -77,33 +78,58 @@ Network configuration in JSON format is streamed to the plugin through stdin. Th
...
-    "ip": <ipv4-and-subnet-in-CIDR>,
-    "gateway": <ipv4-of-the-gateway>,  (optional)
-    "routes": <list-of-ipv4-routes>    (optional)
-  },
-  "ip6": {
-    "ip": <ipv6-and-subnet-in-CIDR>,
-    "gateway": <ipv6-of-the-gateway>,  (optional)
-    "routes": <list-of-ipv6-routes>    (optional)
-  },
+  "interfaces": [
```
eyakubovich: I'm unclear how this is going to be used by the container engine (the thing invoking CNI). The interface that is inside the container (one with `sandbox=<netns>`) is quite well-defined. But what about the rest of them? Sounds like another one would be the "host end". But there's no way to tell what this "host end" is for, its role or how it's connected to other components (e.g. `linux-bridge` or `ovs`). Or what happens in case of macvlan, ipvlan, ovs internal port, etc -- these things don't really have a host end.

So if I'm writing a container engine and invoking CNI plugin, what can I do with this information, provided that I don't know if it'll be there or not? Or if it is there, what can be done with that end. I guess it sounds like this is special casing `veth` interfaces and thus leaks the abstraction.

rosenhouse: One use could be to pass the results to subsequent CNI plugins that are specialized enough to need certain bits of this information.  The runtime could be agnostic to it.

This relates to the "chaining" that has been discussed recently in #46 .  On the other hand, I'm hard-pressed to identify a chaining use case that doesn't conflict with your comment at https://github.com/containernetworking/cni/issues/151#issuecomment-237362660.

dcbw: @eyakubovich @rosenhouse the use-cases that I currently have in mind are:

1. host port<->container port mapping which, if you're using Docker networking you get for free but of course this is CNI so you have to invent it yourself.  We've done that in Kubernetes but the iptables rules normally used for this require the host-side bridge interface name.  I suppose we could write a CNI port forwarding plugin and chain it but that would still require the host bridge interface name.
2. traffic shaping which typically requires the host bridge interface name and the host-side veth interface name.  Could be done as a CNI plugin or utility library too, but still needs chaining.
3. future VM/hypervisor based runtimes, where the networking setup needs to be done before the VM actually starts, so the thing starting the VM needs to know about the host-side resources the CNI plugin set up.  For example, after the CNI plugin creates a macvtap interface the VM-based runtime would then find the associated character device and pass that to qemu.

Maybe @steveeJ can comment more too, since he and I worked together to develop this specific API proposal a few weeks ago.

SPEC.md
dcbw: @steveeJ I since the new fields don't overwrite the old ones, we could make it backwards compatible by specifying the mapping between old and new (eg, first IP address of sandbox interface goes into the old 'ipv4' and 'ipv6' blocks).  But I think in the last meeting we discussed this and decided it wasn't necessarily worth it?  Need clarification.

plugins/main/bridge/bridge.go
dcbw: Nope, comment removed.

plugins/main/bridge/bridge.go
dcbw: Done.

plugins/main/bridge/bridge.go
dcbw: Done.

plugins/main/bridge/bridge_test.go
dcbw: Done, you pass in an empty object so it's (&netlink.Bridge{})

SPEC.md
eyakubovich: @dcbw While I don't view that all of container networking functionality needs to be in CNI (only parts that are meant to be pluggable), your use cases are certainly valid (in a sense that they need to be implemented somehow).

But the part that I'm missing is how will you set up traffic shaping if the plugin can return whatever it wants in the `interfaces` section. Will you try to introspect the "host" interfaces to see what they are and where they might be connected to?

Specific comments:
> host port<->container port mapping which, if you're using Docker networking you get for free but of course this is CNI so you have to invent it yourself. We've done that in Kubernetes but the iptables rules normally used for this require the host-side bridge interface name. I suppose we could write a CNI port forwarding plugin and chain it but that would still require the host bridge interface name.

rkt implements port mapping via DNAT to container-ip, assuming that it's reachable from the host.

> traffic shaping which typically requires the host bridge interface name and the host-side veth interface name. Could be done as a CNI plugin or utility library too, but still needs chaining.

I know very little about traffic shaping but I would think you'd want to do that to the interface inside the container. For example, I doubt you can set tc on a slave of macvlan from the host side.

plugins/main/ptp/ptp.go
dcbw: The diff just makes it look odd.  We're indenting the route-delete+route-add one level and moving them into a for() loop for all addresses, so you're looking at the route-delete block.  That scope hasn't changed.  A full diff would be clearer.

plugins/main/bridge/bridge.go
dcbw: @steveeJ I see what you mean now.  But I don't see the ability to set route metrics in the code, and the spec says "Routes are expected to be added with a 0 metric." so I think we should punt that to another PR.

SPEC.md
dcbw: > But the part that I'm missing is how will you set up traffic shaping if the plugin can return whatever it wants in the interfaces section. Will you try to introspect the "host" interfaces to see what they are and where they might be connected to?

Yes, that's what I would do in Kubernetes for now, for the simple cases.  If one of the returned host interfaces is a bridge, and the plugin has requested shaping, then that's the simplest way.  But this may actually be best handled by a CNI plugin, since then you don't need hints back to Kubernetes that shaping should be enabled.

> rkt implements port mapping via DNAT to container-ip, assuming that it's reachable from the host.

Good to know, but is that still true if rkt isn't setting up the container network, instead if that is being set up from Kubernetes via CNI plugins before the container is run with rkt?  Or is the DNAT stuff completely separate from any of the rkt network code?

> I know very little about traffic shaping but I would think you'd want to do that to the interface inside the container. For example, I doubt you can set tc on a slave of macvlan from the host side.

Current shaping using TC in Kubernetes uses the main host bridge interface (to set up the qdisc on the parent interface).  Shaping with OVS uses the host-side veth.

asridharan: @dcbw 
>host port<->container port mapping which, if you're using Docker networking you get for free but of course this is CNI so you have to invent it yourself. We've done that in Kubernetes but the iptables rules normally used for this require the host-side bridge interface name. I suppose we could write a CNI port forwarding plugin and chain it but that would still require the host bridge interface name.
I guess you can always pass the `host bridge` device name to the port forwarding plugin in the CNI config for the port forwarding plugin as an "exclude_list"?

dcbw:
@steveeJ @rosenhouse reworked for comments from Stefan, and added a new commit bumping the CNI version and adding some version checking.

rosenhouse:
@dcbw this version checking looks pretty different from what I put together in #267.  But I'm ok with that.  I rather like the idea of adding the version checking as part of actual feature development, where there's a clear motivating use case.  👍

dcbw:
@rosenhouse ah right, I'd forgotten about #267.  I like parts of it and I'm happy to drop the versioning patch from this PR for now and work with you to add versioning in #267.  One thing that's not addressed in #267 is how wrapped plugins (like IPAM) would work; it would require a VERSION call before doing the IPAM delegate call.  But maybe that's fine, and it's certainly simpler than what I have here.

dcbw:
@steveeJ thoughts?  Should I drop the version patch for now?

dcbw:
@containernetworking/cni-maintainers any further comments on the proposed API changes?

dcbw:
Rebased now that @rosenhouse versioning changes have landed...

dcbw:
@squeed care to take a look at this one too?  Comments appreciated!

pkg/ipam/ipam.go
```diff
@@ -45,21 +49,27 @@ func ConfigureIface(ifName string, res *types.Result) error {
...
-			// we skip over duplicate routes as we assume the first one wins
-			if !os.IsExist(err) {
-				return fmt.Errorf("failed to add route '%v via %v dev %v': %v", r.Dst, gw, ifName, err)
+
+		for _, r := range ipc.Routes {
+			gw := r.GW
+			if gw == nil {
+				gw = ipc.Gateway
+			}
+			if err = ip.AddRoute(&r.Dst, gw, link); err != nil {
```
squeed: Is it possible for ipc.Gateway to be nil here?

SPEC.md
```diff
@@ -87,33 +88,58 @@ Network configuration in JSON format is streamed to the plugin through stdin. Th
...
+      {
+          "name": "<name>",
+          "mac": "<MAC address>",                            (required if L2 addresses are meaningful)
+          "sandbox": "<netns path or hypervisor identifier>" (required for container/hypervisor interfaces)
```
squeed: What is this value when the interface is on the host side? "host"?

libcni/api_test.go
```diff
@@ -84,10 +84,13 @@ var _ = Describe("Invoking the plugin", func() {
 			Expect(err).NotTo(HaveOccurred())
 
 			Expect(result).To(Equal(&types.Result{
-				IP4: &types.IPConfig{
-					IP: net.IPNet{
-						IP:   net.ParseIP("10.1.2.3"),
-						Mask: net.IPv4Mask(255, 255, 255, 0),
+				IP: []*types.IPConfig{
+					{
+						Version: "4",
```
squeed: Does this need an `Interface` field?

pkg/types/types.go
```diff
@@ -69,34 +69,36 @@ type NetConf struct {
 
 // Result is what gets returned from the plugin (via stdout) to the caller
 type Result struct {
-	IP4 *IPConfig `json:"ip4,omitempty"`
-	IP6 *IPConfig `json:"ip6,omitempty"`
-	DNS DNS       `json:"dns,omitempty"`
+	Interfaces []*Interface `json:"interfaces,omitempty"`
+	IP         []*IPConfig  `json:"ip,omitempty"`
```
squeed: minor nit here: `Interfaces` is plural, but `IP` isn't.

pkg/ipam/ipam.go
dcbw: Yeah, but that's OK since it's assigned to the netlink route object later, and expected to be nil if no gateway is used.

SPEC.md
dcbw: empty or omittted from the json.  Clarified.

libcni/api_test.go
dcbw: No, it's just the Go struct version of the ReportResult in BeforeEach().  Since that doesn't report an Interface, we don't need one here.

pkg/types/types.go
dcbw: @steveeJ requested "IP" in https://github.com/containernetworking/cni/pull/145#discussion-diff-71392017

dcbw:
@containernetworking/cni-maintainers any more comments?

pkg/types/types.go
squeed: heh, OK :-)

rosenhouse:
@dcbw : I finally got around to looking at this.  Looks good, except of course for versioning 😢 

Two scenarios need consideration:

- Runtimes built with this code, invoking older plugins.   I opened #301 to provide a high-level test.  I **think** we should merge that PR first, then rebase this and do the work here to get the tests green.  But I'm happy to consider alternatives.

- Older runtimes invoking plugins built with this code.  For this, it would be nice to also have some kind of high-level test, where a "fake" runtime using a legacy version of CNI invokes a current plugin and parses the response.

@containernetworking/cni-maintainers & @squeed : I don't have a lot of bandwidth on this right now.  It would be great if someone could review #301 and to throw together a similar high-level test of the second bullet above.  Then we'd have some failing tests to guide us here.

SPEC.md
```diff
@@ -78,7 +79,7 @@ It will then look for this executable in a list of predefined directories. Once
 - `CNI_COMMAND`: indicates the desired operation; `ADD`, `DEL` or `VERSION`.
 - `CNI_CONTAINERID`: Container ID
 - `CNI_NETNS`: Path to network namespace file
-- `CNI_IFNAME`: Interface name to set up
+- `CNI_IFNAME`: Interface name to set up; plugin must honor this interface name or return an error
```
tomdee: Is CNI_IFNAME still optional? If so, that should be made clear in the spec.

dcbw: No, I dropped that part to reduce the changeset and controversy.

dcbw: @tomdee Also, this change clarifies what's currently the de-facto requirement for internal plugins at least; CNI_IFNAME is required for both add/del in skel.go (which all internal plugins use) and they will respect it.  I guess the spec is somewhat ambiguous, a 3rd party plugin can name the interface whatever it wants, regardless of CNI_IFNAME, but that violates some assumptions on the part of runtimes.

dcbw:
> Specifically, it adds the ability to return multiple interfaces. Can you remind me why that's required and provide a statement on whether any of the built in plugins will ever return more than one interface.

@tomdee this was mostly to facilitate both sandbox and host-side interfaces. If we completely drop host-side interfaces (and figure out the issues with things like HostPorts and shaping which require the host bridge name) then we don't need more than one interface to be reported.

Some history... the original PR didn't just passed back the single sandbox interface.  Then I wanted to facilitate the HostPorts/shaping type stuff that Kubernetes does, and added the host interface parts.  Originally it was simply a boolean or something, but then @steveeJ and I got into a discussion about VM/hypervisor use-cases and decided to add the 'sandbox' key.  Maybe that's a bit too far for now though.

squeed:
@dcbw I still think the host interfaces would be useful - especially for tc and port forwarding. If it needs to be split in a separate PR, so be it - but it sure feels like the right approach to me.

pkg/types/types.go
leodotcloud: @dcbw Shouldn't IP information be embedded inside the interface? If there is a need to created 2 interfaces, how to associate which IP needs to be configured on a particular interface?

tomdee:
LGTM - I think this is ready to merge right?

mpaluru:
@tomdee @dcbw pointed out that there is another discussion to address this, hence closed this PR.

rosenhouse:
@tomdee we're still waiting on versioning fixes to this PR, right?

I expect that a rebase on master would show red tests of the backwards compatibility

tomdee:
Ah, good point, thanks @rosenhouse

steveeJ:
The last CI run here was _about a month ago_ according to travis, so this still needs a rebase, right?

rosenhouse:
Now that #301 and #307 are both merged, this PR needs to be rebased.  I expect the backward compatibility tests will fail after the rebase, because this PR makes backwards incompatible changes.    The fix will be non-trivial -- by somehow supporting the new and the old stuff in a single codebase.

dcbw:
@containernetworking/cni-maintainers repushed with a bunch of versioning rework.

Theory is as follows... plugins now pull the cniVersion out of the net config and a new method types.PrintResult() handles printing the result in the given version.  I'd actually like to move this to the Result interface, but not this push.

When parent plugins (eg, bridge, ipvlan, etc) receive a result from a delegate, they always convert that result into the most current result version so that IPAM code can be simpler.  Supporting multiple result versions in each plugin would be madness.

Only the most current version (pkg/types/current) supports transformation to/from older formats.  I didn't think it was worthwhile to have every version support transformation to every other version, and doing thing would also cause import loops.

SPEC.md
```diff
@@ -87,33 +88,58 @@ Network configuration in JSON format is streamed to the plugin through stdin. Th
...
+          "mac": "<MAC address>",                            (required if L2 addresses are meaningful)
+          "sandbox": "<netns path or hypervisor identifier>" (required for container/hypervisor interfaces, empty/omitted for host interfaces)
+      }
+  ],
+  "ip": [
+      {
+          "version": "<4-or-6>",
+          "address": "<ip-and-prefix-in-CIDR>",
+          "gateway": "<ip-address-of-the-gateway>",      (optional)
+          "interface": "<index into 'interfaces' list>",
```
tgraf: Gow about adding `"ip"` to the particular interface it should be assigned to. This makes it crystal clear and avoids the user to having to reference via index to the `"interfaces"` array?

SPEC.md
```diff
@@ -87,33 +88,58 @@ Network configuration in JSON format is streamed to the plugin through stdin. Th
...
+          "routes": [                                    (optional)
```
tgraf: It seems that the only reason to associate routes with an IP is to derive the default gateway. What is the use case for defining a default gateway per address? I think `"routes"` should live independently outside of `"ip"` and `"gw"` should be made mandatory. That makes the API a lot easier.

SPEC.md
```diff
@@ -87,33 +88,58 @@ Network configuration in JSON format is streamed to the plugin through stdin. Th
...
+              {
+                  "dst": "<ip-and-prefix-in-cidr>",
+                  "gw": "<ip-of-next-hop>"               (optional)
```
tgraf: Can we rename this to `"nexthops"` and make it a list of nexthop addresses? It's fine to only accept a single nexthop at first but multiple nexthops will definitely be useful in the future and we shouldn't be required to change the spec again.

pkg/ipam/ipam.go
```diff
@@ -46,21 +50,27 @@ func ConfigureIface(ifName string, res *current.Result) error {
...
 
-	// TODO(eyakubovich): IPv6
-	addr := &netlink.Addr{IPNet: &res.IP4.IP, Label: ""}
-	if err = netlink.AddrAdd(link, addr); err != nil {
-		return fmt.Errorf("failed to add IP addr to %q: %v", ifName, err)
-	}
+	for _, ipc := range res.IP {
+		if int(ipc.Interface) >= len(res.Interfaces) || res.Interfaces[ipc.Interface].Name != ifName {
+			// IP address is for a different interface
+			continue
```
tgraf: This should probably not be a silent error.

SPEC.md
dcbw: @tgraf could you clarify?  One reason to have the index thing is that IPAM plugins are largely independent of any L2 configuration, eg host-local or static config shouldn't care about what interface they are acquiring an IP for. So they don't pass back an 'interfaces' field in which the IP could be included, or an interface index, because they don't have any of that information.  

Assuming that's what you mean here, which could be completely wrong...

tgraf: @dcbw I thought the `"interface"` member was required as there was no `(optional)` in the spec itself, hence I didn't see why not to move it into the interface and omit the "interface" altogether. The description later on declares it as optional for IPAM plugins and in that context, your statement makes sense.

What I wonder though is if the IPAM plugin cannot not indicate the interface how is this API supposed to work if multiple interfaces belong to multiple networks and thus require multiple addresses. Is the network plugin expected to resolve this complexity independently of this API?

rosenhouse:
@dcbw The versioning work in here looks great, I think that part is good to go.

dcbw:
@rosenhouse would it help to split the first few commits that make Result an interface into a separate PR that could be reviewed and committed separately?  Happy to do that.

tgraf:
Does anyone have doubts that this is the wrong direction in general? This would be great to have and I see little value for the compatibility layer if we later don't merge the new model.

SPEC.md
dcbw: @tgraf I should clarify that.  The existing relevant text is in the "Result" section:

 The `ip` and `dns` items should be the same output as was returned by the IPAM plugin (see [IP Allocation](#ip-allocation) for details) except that the plugin should fill in the `interface` indexes appropriately, which are missing from IPAM plugin output since IPAM plugins should be unaware of interfaces.

rosenhouse:
@dcbw @tgraf IMHO, this is 👍  good to merge.

I couldn't tell if you intended to (or maybe already had?) address @tgraf 's comments before the merge.  I'm ok with either.

rosenhouse:
FWIW, I think we could merge this now and get a lot of value from it.

Supporting more flexible use cases for routes etc would be nice, but we can make that a later PR for a later release.  This isn't the last spec change that we'll want to make that causes a version bump and requires compatibility work.

tgraf:
I think we should address the comments that were brought on the spec part, namely:
 - Move `[]route` out of addresses. it will be confusing to have the same route attached to multiple addresses.
 - Turn `gw` into a array of `[]nexthop` even if it currently only ever holds a single nexthop to support multiple nexthops in the future.

SPEC.md
dcbw: I've updated the spec to clarify what IPAM plugins return.

SPEC.md
dcbw: Per discussion with CNI maintainers, we agreed to move Routes out of the IPs structure for now, but leave GW in each address since that's harder to extract at this time due to the bridge plugin.  We also have not made Routes.Gw a list yet, but only to get this large PR merged.

Following this PR I will do another PR (or if I don't, certainly entertain one!) to do all or some of:
1) rename Routes.Gw -> Routes.NextHops
2) make Routes.NextHops an array
3) make Routes.NextHops mandatory
4) remove the "Gateway" field from the IPConfig structure

All these things would require copying types.Route out of pkg/types/types.go into *both* pkg/types/current and pkg/types/020, then modifying the pkg/types/current/types.go version with the required changes.

SPEC.md
dcbw: Not quite yet, see https://github.com/containernetworking/cni/pull/145#discussion_r96880817

pkg/ipam/ipam.go
dcbw: done

dcbw:
@tgraf I just pushed an update that move Routes out of the IPs structure and updates the spec to clarify what IPAM plugins return.  See https://github.com/containernetworking/cni/pull/145#discussion_r96880817 for more details on immediate plans.

@containernetworking/cni-maintainers this push has these changes:
1) moves Routes out of the IPs structure to a top-level alongside IPs and DNS
2) renames the "IP" list in Result to "IPs" to be consistent with Routes and Interfaces
3) adds a testcase to pkg/ipam to ensure we don't break it's Gateway handling functionality in the future
4) updates the Spec with some forgotten things and to clarify what IPAM plugins should returnd

PTAL thanks!

tomdee:
The spec changes look OK to me, with just one simple fix:
* You should bump the spec version at the top of the SPEC.md (might need a rebase)

That being said, I'm a little fuzzy on what the routes (or gateway) are actually used for or needed for. Is that information consumed by any runtimes or is it just just used for passing info between IPAM and network plugins?

tgraf:
@dcbw Given that `[]nexthops` can be added later on while staying compatible with the non-array `gw` I believe what is being proposed a good course of action.

dcbw:
> That being said, I'm a little fuzzy on what the routes (or gateway) are actually used for or needed for. Is that information consumed by any runtimes or is it just just used for passing info between IPAM and network plugins?

@tomdee mostly passed between plugins.  I can't see why anything except the thing that configures the container's networking (eg, the IPAM-wrapping plugin) would use it.

dcbw:
> You should bump the spec version at the top of the SPEC.md (might need a rebase)

@tomdee if we're landing both this and #145 at/around the same time, both change the spec, so I'd prefer to either (a) get #145 committed and then commit this and leave the spec version at 0.3.0 (which #145 sets it to) or (b) commit this first and then commit #145 immediately after, which still bumps the spec.

My motivation is that changing the spec version actually causes a lot of churn in the plugins and testcases now that we have real versioning :(  Thoughts?

dcbw:
@tomdee updated the spec version, I don't know why I hadn't done that before...

dcbw:
@tomdee @rosenhouse @squeed rebased on top of ordered chaining.

SPEC.md
```diff
@@ -92,33 +93,60 @@ Network configuration in JSON format is streamed to the plugin through stdin. Th
...
+          "mac": "<MAC address>",                            (required if L2 addresses are meaningful)
+          "sandbox": "<netns path or hypervisor identifier>" (required for container/hypervisor interfaces, empty/omitted for host interfaces)
+      }
+  ],
+  "ip": [
+      {
+          "version": "<4-or-6>",
+          "address": "<ip-and-prefix-in-CIDR>",
+          "gateway": "<ip-address-of-the-gateway>",          (optional)
+          "interface": "<index into 'interfaces' list>"
```
rosenhouse: Totally minor nit-pick, but the quotes probably don't belong around the numeric index value.

SPEC.md
```diff
@@ -438,6 +465,68 @@ IPAM plugin examples:
...
+
+#### IP
+
+```
+  "ips": [
+      {
+          "version": "<4-or-6>",
+          "address": "<ip-and-prefix-in-CIDR>",
+          "gateway": "<ip-address-of-the-gateway>",      (optional)
+          "interface": "<index into 'interfaces' list>" (not required for IPAM plugins)
```
rosenhouse: Again, I think this numeric index for `interface` shouldn't have quotes around the value.

dcbw: @rosenhouse done

SPEC.md
dcbw: @rosenhouse done

dcbw:
@rosenhouse fix up the spec; if this looks good mind hitting the merge button?  Thanks!

rosenhouse:
It would be an honor...

plugins/ipam/dhcp/daemon.go
```diff
@@ -70,11 +71,12 @@ func (d *DHCP) Allocate(args *skel.CmdArgs, result *types.Result) error {
 
 	d.setLease(args.ContainerID, conf.Name, l)
 
-	result.IP4 = &types.IPConfig{
-		IP:      *ipn,
+	result.IPs = []*current.IPConfig{{
+		Version: "4",
```
dcbw: @AlexeyKasatkin nothing about the Allocate() call and the lease acquisition process changes.  The DHCP plugin does not currently support DHCPv6 at all, and thus it's still limited to DHCPv4.  Any invocation of the DHCP plugin will inherently try only IPv4.

But IPv6 patches would be gratefully accepted!

PR #146, opened 2016-03-03
Use Ubuntu Trusty in travis

achanda:
This boots faster

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: go
 sudo: required
+dist: trusty
 
 go:
   - 1.4


steveeJ:
What's the default? I'd prefer the one with the most recent kernel over bootup time.

achanda:
The default is precise. The highest kernel version that Ubuntu repos have is 4.2 for trusty and 3.8 for precise. So, I think trusty has a newer kernel. 

http://packages.ubuntu.com/trusty/kernel/
http://packages.ubuntu.com/precise/kernel/

achanda:
The kernel version is shown in the build jobs. Precise is `3.13.0-63-generic` while trusty is `3.19.0-28-generic`

zachgersh:
@steveeJ I didn't buy the faster boot time as the primary argument either but since this is the newer kernel I am going to go ahead and merge.

We can always make a minimal change to the `.travis.yml` to revert this if any craziness crops up.

Thanks for the PR @achanda

PR #148, opened 2016-03-04
travis & README: add coveralls coverage report

### LINKED ISSUE #66: missing functional tests in CNI
We should add functional for all combinations of plugins and their configuration.
rkt has partial [functional tests for networking](https://github.com/coreos/rkt/issues/1010), some of which probably fit better in here.

This issue is for overview and tracking purposes.

steveeJ:
Fixes #66.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,13 +7,18 @@ go:
   - 1.6
 
 env:
-  - TOOLS_CMD=golang.org/x/tools/cmd
+  global:
+    - TOOLS_CMD=golang.org/x/tools/cmd
+    - PATH=$GOROOT/bin:$PATH
 
 install:
  - go get ${TOOLS_CMD}/vet
+ - go get ${TOOLS_CMD}/cover
+ - go get github.com/modocache/gover
+ - go get github.com/mattn/goveralls
 
 script:
- - sudo -E /bin/bash -c 'PATH=$GOROOT/bin:$PATH ./test'
+ - ./test
 
 notifications:
   email: false
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 [![Build Status](https://travis-ci.org/appc/cni.svg?branch=master)](https://travis-ci.org/appc/cni)
+[![Coverage Status](https://coveralls.io/repos/github/appc/cni/badge.svg?branch=master)](https://coveralls.io/github/appc/cni?branch=master)
 
 # cni - the Container Network Interface
 
--- a/test
+++ b/test
@@ -33,8 +33,23 @@ fi
 split=(${TEST// / })
 TEST=${split[@]/#/${REPO_PATH}/}
 
-echo "Running tests..."
-go test ${COVER} $@ ${TEST}
+echo -n "Running tests "
+function testrun {
+    sudo -E bash -c "umask 0; PATH=\$GOROOT/bin:\$PATH go test -covermode set $@"
+}
+if [ ! -z "${COVERALLS}" ]; then
+    echo "with coverage profile generation..."
+    i=0
+    for t in ${TEST}; do 
+        testrun "-coverprofile ${i}.coverprofile ${t}"
+        i=$((i+1))
+    done 
+    gover
+    goveralls -service=travis-ci -coverprofile=gover.coverprofile -repotoken=$COVERALLS_TOKEN
+else 
+    echo "without coverage profile generation..."
+    testrun "${TEST}"
+fi
 
 echo "Checking gofmt..."
 fmtRes=$(gofmt -l $FMT)


steveeJ:
@zachgersh do you know of a better way to add coverage reports for our environment? I've tried to use *ginkgo* to run the tests the coverage reports contained paths that weren't compatible with the `goveralls` CLI tool.

zachgersh:
@steveeJ sorry for the delayed reply :D I am pretty sure what you've got is the correct approach (at least from the code that I see).

instead of using go test were you running `ginkgo` with the `-cover` flag as your command or just ginkgo by itself?

As a different thing to try - what happens if you use the `goverall` command to run the tests and output the coverage report (it is supposed to do both).

Would it be possible to put the output you were getting here on the issue?

Also, I see a note that goveralls is wants you to use [github.com/axw/gocov/gocov](github.com/axw/gocov/gocov). Is that a requirement? I need to read a little bit more as to how that fits in.

steveeJ:
@zachgersh no rush, we're on a weekend ;-)

> instead of using go test were you running ginkgo with the -cover flag as your command or just ginkgo by itself?
> (...)
> Would it be possible to put the output you were getting here on the issue?

Applying the following patch to my current PR:
```diff
diff --git a/test b/test
index 3811771..0143d81 100755
--- a/test
+++ b/test
@@ -38,12 +38,8 @@ function testrun {
     sudo -E bash -c "umask 0; PATH=\$GOROOT/bin:\$PATH go test -covermode set $@"
 }
 if [ ! -z "${COVERALLS}" ]; then
+    sudo -E bash -c "ginkgo -v -cover ${TESTABLE}"
     echo "with coverage profile generation..."
-    i=0
-    for t in ${TEST}; do 
-        testrun "-coverprofile ${i}.coverprofile ${t}"
-        i=$((i+1))
-    done 
     gover
     goveralls -service=travis-ci -coverprofile=gover.coverprofile -repotoken=$COVERALLS_TOKEN
 else    
```
yields the result
```
21:23:56 gocover.go:41: Can't find can't find "args.go": cannot find package "_/home/steveej/src/github/appc/cni/pkg/invoke/" in any of:
	/nix/store/zfjn6abbd49iyiai3fnvdnxjkqjqwxxi-go-1.5.3/share/go/src/_/home/steveej/src/github/appc/cni/pkg/invoke (from $GOROOT)
	/home/steveej/src/github/appc/cni/gopath/src/_/home/steveej/src/github/appc/cni/pkg/invoke (from $GOPATH)
	/home/steveej/src/github/appc/cni/Godeps/_workspace/src/_/home/steveej/src/github/appc/cni/pkg/invoke
```
It seems that the coverprofiles produced by `ginkgo` are different from what's produced by `go test` and can't be processed by `goveralls`, and it's author [is going to remove `gocov`](https://github.com/mattn/goveralls/pull/64), so I'm guessing that can't be required.

> As a different thing to try - what happens if you use the goverall command to run the tests and output the coverage report (it is supposed to do both).

In case `goveralls` runs the test, it's merely running `go test -coverprofile...` and therefore does not support multiple package names. For the same reason I've written the loop that runs `go test` per package.

zachgersh:
@steveeJ - thanks for the awesome response! I took a look, and both ginkgo and go test are generating the exact same type of coverage report (which I had hoped they would). I am going to dig into this a little bit more as there seems to be a lot of complexity/tooling around this one area (who know uploading a coverage report was so difficult).

.travis.yml
```diff
@@ -7,13 +7,18 @@ go:
...
+    - PATH=$GOROOT/bin:$PATH
 
 install:
  - go get ${TOOLS_CMD}/vet
+ - go get ${TOOLS_CMD}/cover
+ - go get github.com/modocache/gover
+ - go get github.com/mattn/goveralls
 
 script:
- - sudo -E /bin/bash -c 'PATH=$GOROOT/bin:$PATH ./test'
```
zachgersh: It dawned on me that removing this line was probably the root cause of the issues you were seeing (as we saw similar path related issues when setting these tests up).

Due to the way that Travis sets up your `GOPATH` / `GOROOT` we came up with that sudo line so that we could hand all of the `ENV` vars through to the test script.

Bringing this line back should allow you to run the `ginkgo` command in the test script.

steveeJ: When not using `sudo` there's not need to hand through explicitly.

zachgersh: @steveeJ - I agree. The issue we seem to be having is due to the files not being found which makes me think something is up with the `GOPATH`. Can we try putting it back and see if it works properly.

steveeJ:
Actually the errors I showed in my comments on the PR are not from travis but were produced on my laptop. 

> both ginkgo and go test are generating the exact same type of coverage report

I'm afraid I can't reproduce this. The biggest difference, which can be seen in the patch above, seems that `ginkgo` takes the path to a package as a opposed to a package name in case of `go test`. This apparently leads to different paths in the report, although they are theoretically the same results when it comes to coverage.

steveeJ:
ping @zachgersh. sorry I'm getting a little impatient over here :-) I would really like to place that batch on the README. I think this is in a mergable state and we can figure out how/if we can use `ginkgo` in a follow-up.

/cc @jonboulle

zachgersh:
@steveeJ - I say go for it. I am going to compare the two coverage files as I want to see the discrepancy you pointed out. Probably going to end up patching ginkgo if it is different so we have that compatibility.

PR #154, opened 2016-03-15
WithNetNS restores original namespace when callback errors

rosenhouse:
Without this, an error in the callback function would leave the caller of WithNetNS in the wrong namespace.


steveeJ:
Very nice, thanks! Would you mind adding a line to the docstring of the `WithNetNS` function?

zachgersh:
With the tests passing and the docstring updated - this lgtm

PR #156, opened 2016-03-16
Moves to go team vendoring solution

zachgersh:
Now that the go team has decided that `vendor` is the way to do dependencies - i'd like to propose we move as well. Note, this vendor folder will become the default in 1.7 and it would be good to stay ahead of the curve.

Only real impact we see - now go 1.5 is required to build the project (which is now noted in the readme).

cc @steveeJ


steveeJ:
I had to double-check if the CoreOS build system has moved away from Go 1.4, and it has. We should be fine in that regard.

zachgersh:
@steveeJ and I already know why the build is red for 1.5 :/

alban:
It still use `Godeps/_workspace` on master:
https://github.com/appc/cni/blob/master/build#L13
```
export GOPATH=${PWD}/gopath:$(pwd)/Godeps/_workspace
```

PR #158, opened 2016-03-17
Add generic IgnoreUnknown to valid CNI_ARGS

### LINKED ISSUE #126: host-local IPAM plugin doesn't work with kubernetes

steveeJ:
Plugin developers that use CNI_ARGS now need to embed `types.CommonArgs` in their structs.
Then, passing `CNI_ARGS=IgnoreUnknown=1;...` will cause extraneous arguments to be ignored.

Fixes #126
Closes #109 #127

/cc @aanm @tomdee @dcbw @yifan-gu

### DIFF
--- a/pkg/types/args.go
+++ b/pkg/types/args.go
@@ -47,9 +47,9 @@ type CommonArgs struct {
 	IgnoreUnknown UnmarshallableBool `json:"ignoreunknown,omitempty"`
 }
 
-// getKeyField is a helper function to receive Values
+// GetKeyField is a helper function to receive Values
 // Values that represent a pointer to a struct
-func getKeyField(keyString string, v reflect.Value) reflect.Value {
+func GetKeyField(keyString string, v reflect.Value) reflect.Value {
 	return v.Elem().FieldByName(keyString)
 }
 
--- a/pkg/types/args.go
+++ b/pkg/types/args.go
@@ -21,6 +21,39 @@ import (
 	"strings"
 )
 
+// UnmarshallableBool typedef for builtin bool
+// because builtin type's methods can't be declared
+type UnmarshallableBool bool
+
+// UnmarshalText implements the encoding.TextUnmarshaler interface.
+// Returns boolean true if the string is "1" or "[Tt]rue"
+// Returns boolean false if the string is "0" or "[Ff]alse"
+func (b *UnmarshallableBool) UnmarshalText(data []byte) error {
+	s := strings.ToLower(string(data))
+	switch s {
+	case "1", "true":
+		*b = true
+	case "0", "false":
+		*b = false
+	default:
+		return fmt.Errorf("Boolean unmarshal error: invalid input %s", s)
+	}
+	return nil
+}
+
+// CommonArgs contains the IgnoreUnknown argument
+// and must be embedded by all Arg structs
+type CommonArgs struct {
+	IgnoreUnknown UnmarshallableBool `json:"ignoreunknown,omitempty"`
+}
+
+// getKeyField is a helper function to receive Values
+// Values that represent a pointer to a struct
+func getKeyField(keyString string, v reflect.Value) reflect.Value {
+	return v.Elem().FieldByName(keyString)
+}
+
+// LoadArgs parses args from a string in the form "K=V;K2=V2;..."
 func LoadArgs(args string, container interface{}) error {
 	if args == "" {
 		return nil
@@ -29,6 +62,7 @@ func LoadArgs(args string, container interface{}) error {
 	containerValue := reflect.ValueOf(container)
 
 	pairs := strings.Split(args, ";")
+	unknownArgs := []string{}
 	for _, pair := range pairs {
 		kv := strings.Split(pair, "=")
 		if len(kv) != 2 {
@@ -36,15 +70,22 @@ func LoadArgs(args string, container interface{}) error {
 		}
 		keyString := kv[0]
 		valueString := kv[1]
-		keyField := containerValue.Elem().FieldByName(keyString)
+		keyField := GetKeyField(keyString, containerValue)
 		if !keyField.IsValid() {
-			return fmt.Errorf("ARGS: invalid key %q", keyString)
+			unknownArgs = append(unknownArgs, pair)
+			continue
 		}
+
 		u := keyField.Addr().Interface().(encoding.TextUnmarshaler)
 		err := u.UnmarshalText([]byte(valueString))
 		if err != nil {
 			return fmt.Errorf("ARGS: error parsing value of pair %q: %v)", pair, err)
 		}
+	}
+
+	isIgnoreUnknown := GetKeyField("IgnoreUnknown", containerValue).Bool()
+	if len(unknownArgs) > 0 && !isIgnoreUnknown {
+		return fmt.Errorf("ARGS: unknown args %q", unknownArgs)
 	}
 	return nil
 }
--- a/pkg/types/args_test.go
+++ b/pkg/types/args_test.go
@@ -1,0 +1,92 @@
+package types_test
+
+import (
+	"reflect"
+
+	. "github.com/appc/cni/pkg/types"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("UnmarshallableBool UnmarshalText", func() {
+	DescribeTable("string to bool detection should succeed in all cases",
+		func(inputs []string, expected bool) {
+			for _, s := range inputs {
+				var ub UnmarshallableBool
+				err := ub.UnmarshalText([]byte(s))
+				Expect(err).ToNot(HaveOccurred())
+				Expect(ub).To(Equal(UnmarshallableBool(expected)))
+			}
+		},
+		Entry("parse to true", []string{"True", "true", "1"}, true),
+		Entry("parse to false", []string{"False", "false", "0"}, false),
+	)
+
+	Context("When passed an invalid value", func() {
+		It("should result in an error", func() {
+			var ub UnmarshallableBool
+			err := ub.UnmarshalText([]byte("invalid"))
+			Expect(err).To(HaveOccurred())
+		})
+	})
+})
+
+var _ = Describe("GetKeyField", func() {
+	type testcontainer struct {
+		Valid string `json:"valid,omitempty"`
+	}
+	var (
+		container          = testcontainer{Valid: "valid"}
+		containerInterface = func(i interface{}) interface{} { return i }(&container)
+		containerValue     = reflect.ValueOf(containerInterface)
+	)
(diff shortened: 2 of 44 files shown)


pkg/types/args.go
```diff
@@ -21,6 +21,39 @@ import (
...
+// UnmarshallableBool typedef for builtin bool
+// because builtin type's methods can't be declared
+type UnmarshallableBool bool
+
+// UnmarshalText implements the encoding.TextUnmarshaler interface.
+// Returns boolean true if the string is "1" or "[Tt]rue"
+// Returns boolean false if the string is "0" or "[Ff]alse"
+func (b *UnmarshallableBool) UnmarshalText(data []byte) error {
+	s := strings.ToLower(string(data))
+	switch {
```
jonboulle: switch s
But doesn't golang built in json do this for you?

steveeJ: `*bool` doesn't have this functionality. When using the builtin bool type for the _IgnoreExtra_ field, unmarshaling on the reflection variable will throw a not implemented error.

pkg/types/args_test.go
```diff
@@ -0,0 +1,113 @@
...
+	Context("When a valid field is provided", func() {
+		It("should return the correct field", func() {
+			field := types.GetKeyField("Valid", containerValue)
+			Expect(field.String()).To(Equal("valid"))
+		})
+	})
+})
+
+var _ = Describe("LoadArgs", func() {
+
```
zachgersh: nitpick but no space here please :D

pkg/types/args_test.go
```diff
@@ -0,0 +1,113 @@
...
+	. "github.com/onsi/gomega"
+)
+
+func TestArgs(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "pkg/ns Suite")
+}
+
+var _ = Describe("UnmarshallableBool UnmarshalText", func() {
+	positive := []string{"True", "true", "1"}
```
zachgersh: Could you wrap this and the `negative` before it in a `BeforeEach`. Easier to spot. Variable declarations for both would still be at the top of the `Describe` just before the beforeEach

pkg/types/args_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"reflect"
+	"testing"
+
+	"github.com/appc/cni/pkg/types"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestArgs(t *testing.T) {
```
zachgersh: This can be removed from this file. ginkgo will programmatically generate this for you when you run `ginkgo bootstrap` from within the directory you are writing your tests. It makes a `PACKAGE_suite_test.go` file to hold this.

pkg/types/args_test.go
```diff
@@ -0,0 +1,113 @@
...
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "pkg/ns Suite")
+}
+
+var _ = Describe("UnmarshallableBool UnmarshalText", func() {
+	positive := []string{"True", "true", "1"}
+	Context(fmt.Sprintf("When passed one of %q", positive), func() {
+		It("should result in boolean true", func() {
+			ubexp := types.UnmarshallableBool(true)
+			for _, s := range positive {
```
zachgersh: It looks like you want these to be table based tests (which is awesome). Ginkgo has a built in to do this.  You can check it out [here](https://onsi.github.io/ginkgo/#extensions)

zachgersh:
Are these specific tests being run? I didn't see it added to the test script.

aanm:
Can you update the SPEC.md?

steveeJ:
@zachgersh 
Thanks for the review. I'll address your comments shortly. Usually we add the "needs rework" label if appropriate, then again I didn't have the "needs review" label attached ;-)

@aanm 
> Can you update the SPEC.md?
What would require a change to the spec? It's supposed to be an implementation detail.

pkg/types/args_test.go
steveeJ: How exactly does that work? I've been manually running the tests using plain `go test github.com/appc/cni/pkg/types` so far.

tomdee:
I don't understand how this will work, could you add a little more to the description? Specifically, does this approach mean that there will need to be a Kubernetes updates to set `IgnoreExtra=1`?

steveeJ:
> does this approach mean that there will need to be a Kubernetes updates to set IgnoreExtra=1

That's correct. I'm about to restructure the documentation, as of now there is no real place for such information.

pkg/types/args.go
```diff
@@ -21,6 +21,39 @@ import (
...
+	default:
+		return fmt.Errorf("Boolean unmarshal error: invalid input %s", s)
+	}
+	return nil
+}
+
+// CommonArgs contains the IgnoreExtra argument
+// and must be embedded by all Arg structs
+type CommonArgs struct {
+	IgnoreExtra UnmarshallableBool `json:"ignoreextra,omitempty"`
```
yifan-gu: Does IgnoreUnknown sound better to you?

aanm:
>What would require a change to the spec? It's supposed to be an implementation detail.

>I don't understand how this will work

That's what I meant by updating the `SPEC.md` ;-)

pkg/types/args.go
steveeJ: It does, it's also more clear on the intention :+1:

pkg/types/args_test.go
```diff
@@ -0,0 +1,98 @@
...
+	"testing"
+
+	"github.com/appc/cni/pkg/types"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+)
+
+func TestArgs(t *testing.T) {
```
zachgersh: This func should still be removed.

You want to end up with this directory structure:
https://github.com/appc/cni/tree/master/pkg/ns

You'll see the suite test file in the above directory holds the `t.testing`

here is what the file looks like in question:
https://github.com/appc/cni/blob/master/pkg/ns/ns_suite_test.go

steveeJ: See https://github.com/appc/cni/pull/158#discussion_r56532709.
What's the proper solution for generating this file and running the selected test manually?

steveeJ: I think I got it. I ran `ginkgo bootstrap` in the `pkgs/types` directory. Please let me know if we're good now!

pkg/types/args_test.go
```diff
@@ -0,0 +1,92 @@
...
+	"reflect"
+
+	"github.com/appc/cni/pkg/types"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("UnmarshallableBool UnmarshalText", func() {
```
zachgersh: Last thing I see - wrap all of the tests in some sort of top level Describe so you can avoid having this multiple types

```go
var _ = Describe(
```
Top level describe could just name the package you are testing.

zachgersh:
@steveeJ Looks good - just had one more comment. Apologies for all the back and forth :D

I'll merge after that change.

pkg/types/args_test.go
steveeJ: This is a pure matter of preference, right? I actually like the current structure, they're all testing different areas, but I'm not insisting.

dcbw:
@steveeJ I'll handle the Kubernetes PR if nobody else has gotten there yet.

steveeJ:
> Looks good - just had one more comment. Apologies for all the back and forth :D
> 
> I'll merge after that change.

No need to apologize, this is what reviews are good for, so I'd rather have to thank you! ;-)
Please note that my last change turned a function unexported and put the tests and the implementation the the same package for that reason.

zachgersh:
> Please note that my last change turned a function unexported and put the tests and the implementation the the same package for that reason.
 Show all checks

I hadn't noticed this. Now that you pointed it out, is there a different way to do this? Jamming the tests into the same package as the implementation is general a smell.

pkg/types/args_test.go
zachgersh: Mostly a matter of style but we should be consistent with our test files across the board.

steveeJ:
> I hadn't noticed this. Now that you pointed it out, is there a different way to do this? Jamming the tests into the same package as the implementation is general a smell.

According to https://golang.org/ref/spec#Exported_identifiers it is the only way, but I would rather go this route than exporting identifiers that I don't intend to for other purposes than testing.

zachgersh:
@steveeJ I would prefer to not go this way and instead we should test things via their public interfaces. If you want something to be only internal to CNI and not imported by the rest of the world we should be using:

https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit

This would actually get rid of the code smell.

steveeJ:
> option 1:
> If you want something to be only internal to CNI and not imported by the rest of the world we should be using:

I haven't seen this before and it would work but I don't think it's an style improvement or for anything else.

> option 2:
> make the test package but dot import the code
> 
> either gets rid of the smell

I would not mind using the package name, so it's not a question of importing or not. The import doesn't give access to unexported, so option 2 doesn't help.

---

Not sure why we shouldn't just use the same package for the tests and be happy. My impression is that tnis would also be Go idiomatic. It works and it does exactly what we want which is to test private functions without exporting them. Do you see any technical disadvantages?

zachgersh:
My recommendation here would be put them into the separate `_test` package, make the methods private that you feel should be, and then ONLY test the public method(s) we would use out of the package (maybe this is LoadArgs). 

The above is how most of the go stdlib is structured test wise - [net/http for example](https://github.com/golang/go/tree/master/src/net/http)

If we choose not to go the the `_test` package route then we are also inflating the size of any built go code that might use this package (it will pull in both the tests and implementation when compiled).

steveeJ:
> If we choose not to go the the _test package route then we are also inflating the size of any built go code that might use this package (it will pull in both the tests and implementation when compiled).

This is something I was concerned about too but haven't investigated if it is actually the case. You seem to already know (I'll just trust you with this, but a link for me read about this would be appreciated), and even though our tests suites are not huge, I think this aspect is enough to decide for separate test packages.

steveeJ:
@zachgersh this should be good for now, no more smell ;-)

PR #164, opened 2016-03-23
pkg: add a function to generate chain names

achanda:
Adds a function to generate chain names for use in iptables and
ports all drivers to use that function. Also adds tests for the
said function.
Fixes https://github.com/coreos/rkt/issues/2314

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -1,0 +1,20 @@
+package utils
+
+import (
+	"crypto/sha512"
+	"fmt"
+)
+
+// 29 - len('CNI') - 2*len('-')
+const maxNameLen = 16
+
+// Generates a chain name to be used with iptables.
+// Ensures that the generated name is less than
+// 29 chars in length
+func FormatChainName(name string, id string) string {
+	h := sha512.Sum512([]byte(id))
+	if len(name) > maxNameLen {
+		return fmt.Sprintf("CNI-%s-%x", name[:len(name)-maxNameLen], h[:8])
+	}
+	return fmt.Sprintf("CNI-%s-%x", name, h[:8])
+}
--- a/pkg/utils/utils_suite_test.go
+++ b/pkg/utils/utils_suite_test.go
@@ -1,0 +1,13 @@
+package utils_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"testing"
+)
+
+func TestUtils(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Utils Suite")
+}
--- a/pkg/utils/utils_test.go
+++ b/pkg/utils/utils_test.go
@@ -1,0 +1,18 @@
+package utils
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Utils", func() {
+	It("should format a short name", func() {
+		chain := FormatChainName("test", "1234")
+		Expect(chain).To(Equal("CNI-test-d404559f602eab6f"))
+	})
+
+	It("should truncate a long name", func() {
+		chain := FormatChainName("testalongnamethatdoesnotmakesense", "1234")
+		Expect(chain).To(Equal("CNI-testalongnamethat-d404559f602eab6f"))
+	})
+})
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -28,6 +28,7 @@ import (
 	"github.com/appc/cni/pkg/ns"
 	"github.com/appc/cni/pkg/skel"
 	"github.com/appc/cni/pkg/types"
+	"github.com/appc/cni/pkg/utils"
 	"github.com/vishvananda/netlink"
 )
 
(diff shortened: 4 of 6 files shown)


pkg/utils/utils.go
```diff
@@ -0,0 +1,17 @@
+package utils
+
+import (
+        "crypto/sha512"
+        "fmt"
+)
+
+const maxChainName = 29
+
+func FormatChainName(name string, id string) string {
```
steveeJ: missing docstring

steveeJ:
Thanks!

Please adjust your commit messages according to the [contribution guidelines about commit messages](https://github.com/appc/cni/blob/master/CONTRIBUTING.md#format-of-the-commit-message).

achanda:
Thanks @steveeJ . I updated the commit.

steveeJ:
I'm currently not up to thinking this through myself, so I'd rather ask this than forget about it.
Does this increase the chance of a chain name clash? If so, what happens in that case?

achanda:
@steveeJ I do think this increases collision probability. `iptables` will error out in that case. Alternatives:
1. Disallow long network names
2. Format current system time as a 8 byte string to generate the name
Thoughts?

pkg/utils/utils_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package utils_test
+
```
alban: The [TESTABLE](https://github.com/appc/cni/blob/master/test#L14) list must be updated.

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Utils", func() {
+	It("should format a short name", func() {
+		chain := formatChainName("test", "1234")
```
alban: `formatChainName` with a uppercase "F".

plugins/main/bridge/bridge.go
```diff
@@ -220,7 +221,7 @@ func cmdAdd(args *skel.CmdArgs) error {
 	}
 
 	if n.IPMasq {
-		chain := "CNI-" + n.Name
+		chain := utils.FormatChainName(n.Name, n.Name)
```
alban: Why not using `args.ContainerID` as in the plugin ptp?

pkg/utils/utils.go
```diff
@@ -0,0 +1,20 @@
...
+)
+
+const maxChainName = 29
+
+// Generates a chain name to be used with iptables.
+// Ensures that the generated name is less than
+// maxChainName chars in length
+func FormatChainName(name string, id string) string {
+        h := sha512.Sum512([]byte(id))
+        chain := fmt.Sprintf("CNI-%s-%x", name, h[:8])
```
alban: What about using `name[:16]` to ensure that `h` will not be completely trimmed if `name` is very long?

pkg/utils/utils_test.go
```diff
@@ -0,0 +1,19 @@
...
+		Expect(chain).To(Equal("CNI-test-d404559f602eab6f"))
+	})
+
+	It("should truncate a long name", func() {
+		chain := formatChainName("testtesttesttest", "1234")
+		Expect(chain).To(Equal("CNI-testtesttesttest-d404559f"))
+	})
+})
```
alban: you could also add another test for names longer than 29

achanda:
@steveeJ the earlier implementation could cause collisions more frequently when network names are long. The new implementation does not change collision probability. Thoughts on this?

pkg/utils/utils_suite_test.go
achanda: Done now, thanks.

pkg/utils/utils.go
```diff
@@ -0,0 +1,20 @@
...
+// 29 - len('CNI') - 2*len('-')
+const maxNameLen = 16
+
+// Generates a chain name to be used with iptables.
+// Ensures that the generated name is less than
+// 29 chars in length
+func FormatChainName(name string, id string) string {
+	h := sha512.Sum512([]byte(id))
+	if len(name) > maxNameLen {
+		return fmt.Sprintf("CNI-%s-%x", name[:len(name)-maxNameLen], h[:8])
```
steveeJ: @achanda 
This looks reasonable to me, please take a second look @alban

alban: Not sure to understand... shouldn't it be `name[:maxNameLen]`?

Also, to distinguish "verylongnameofthenetwork01" and "verylongnameofthenetwork02" (both truncated to "verylongnameofth"), could the hash `h` be feed with both `name` and `id`?

steveeJ: Looking at this closer I don't like the math, disregarding of the variable length, truncating a hash value will always increase the risk of collisions significantly. We need a hash function that produces an string output of 25 characters.

steveeJ: @achanda @alban PTAL at https://github.com/appc/cni/pull/168/files#diff-fb81b0a96eed5b010cfb081b3aef4431R15. It's a step closer but theoretically definitely not unique.

achanda: That's what I was trying to arrive at. Should we close this one then?

steveeJ: > Should we close this one then?

I think we can by now, the other PR is ready.

PR #165, opened 2016-03-24
scripts: fix build with go-1.5 after switch to vendor/

dcbw:
go-1.6 enables vendor by default, but go-1.5 needs an environment variable.


dcbw:
@steveeJ looks like .travis.yaml got updated, but that doesn't help those of us with go-1.5...

zachgersh:
@dcbw - whoops, thank you for adding this :+1:

PR #168, opened 2016-03-30
attempt to ensure chain name is unique and at most 29 chars long

steveeJ:
Duplicate of and closes #164.

### DIFF
--- a/pkg/ip/ipmasq.go
+++ b/pkg/ip/ipmasq.go
@@ -23,7 +23,7 @@ import (
 
 // SetupIPMasq installs iptables rules to masquerade traffic
 // coming from ipn and going outside of it
-func SetupIPMasq(ipn *net.IPNet, chain string) error {
+func SetupIPMasq(ipn *net.IPNet, chain string, comment string) error {
 	ipt, err := iptables.New()
 	if err != nil {
 		return fmt.Errorf("failed to locate iptables: %v", err)
@@ -36,25 +36,25 @@ func SetupIPMasq(ipn *net.IPNet, chain string) error {
 		}
 	}
 
-	if err = ipt.AppendUnique("nat", chain, "-d", ipn.String(), "-j", "ACCEPT"); err != nil {
+	if err = ipt.AppendUnique("nat", chain, "-d", ipn.String(), "-j", "ACCEPT", "-m", "comment", "--comment", comment); err != nil {
 		return err
 	}
 
-	if err = ipt.AppendUnique("nat", chain, "!", "-d", "224.0.0.0/4", "-j", "MASQUERADE"); err != nil {
+	if err = ipt.AppendUnique("nat", chain, "!", "-d", "224.0.0.0/4", "-j", "MASQUERADE", "-m", "comment", "--comment", comment); err != nil {
 		return err
 	}
 
-	return ipt.AppendUnique("nat", "POSTROUTING", "-s", ipn.String(), "-j", chain)
+	return ipt.AppendUnique("nat", "POSTROUTING", "-s", ipn.String(), "-j", chain, "-m", "comment", "--comment", comment)
 }
 
 // TeardownIPMasq undoes the effects of SetupIPMasq
-func TeardownIPMasq(ipn *net.IPNet, chain string) error {
+func TeardownIPMasq(ipn *net.IPNet, chain string, comment string) error {
 	ipt, err := iptables.New()
 	if err != nil {
 		return fmt.Errorf("failed to locate iptables: %v", err)
 	}
 
-	if err = ipt.Delete("nat", "POSTROUTING", "-s", ipn.String(), "-j", chain); err != nil {
+	if err = ipt.Delete("nat", "POSTROUTING", "-s", ipn.String(), "-j", chain, "-m", "comment", "--comment", comment); err != nil {
 		return err
 	}
 
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -19,3 +19,9 @@ func FormatChainName(name string, id string) string {
 	chain := fmt.Sprintf("%s%x", chainPrefix, chainBytes)
 	return chain[:maxChainLength]
 }
+
+// FormatComment returns a comment used for easier
+// rule identification within iptables.
+func FormatComment(name string, id string) string {
+	return fmt.Sprintf("name: %q id: %q", name, id)
+}
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -12,7 +12,7 @@ const (
 )
 
 // Generates a chain name to be used with iptables.
-// Ensures that the generated chain name is less than
+// Ensures that the generated chain name is exactly
 // maxChainLength chars in length
 func FormatChainName(name string, id string) string {
 	chainBytes := sha512.Sum512([]byte(name + id))
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -5,16 +5,17 @@ import (
 	"fmt"
 )
 
-const ChainCNIPrefix = ("CNI-")
-const MaxChainLength = 29 - len(ChainCNIPrefix)
+const (
+	maxChainLength = 28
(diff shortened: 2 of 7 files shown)


pkg/utils/utils.go
```diff
@@ -0,0 +1,19 @@
+package utils
+
+import (
+	"crypto/sha512"
+	"fmt"
+)
+
+const MaxChainLength = 29 - len("CNI-")
```
achanda: This is max name of name, not the whole chain. The var name is misleading. Also, shouldn't we take into account `-` twice?

pkg/utils/utils.go
```diff
@@ -0,0 +1,19 @@
...
+
+// Generates a chain name to be used with iptables.
+// Ensures that the generated name is less than
```
achanda: s/generated name/generated chain name/g

pkg/utils/utils.go
```diff
@@ -0,0 +1,20 @@
+package utils
+
+import (
+	"crypto/sha512"
+	"fmt"
+)
+
+const ChainCNIPrefix = ("CNI-")
+const MaxChainLength = 29 - len(ChainCNIPrefix)
```
achanda: Why do we need to export these two?

pkg/utils/utils.go
```diff
@@ -0,0 +1,20 @@
+package utils
+
+import (
+	"crypto/sha512"
+	"fmt"
+)
+
+const ChainCNIPrefix = ("CNI-")
```
achanda: Just `CNI-` should work, no?

pkg/utils/utils.go
steveeJ: We don't, good point! I'll also rename it as I don't like the name anymore today.

achanda:
LGTM

pkg/utils/utils.go
```diff
@@ -0,0 +1,21 @@
...
+	maxChainLength = 29
+	chainPrefix    = "CNI-"
+	prefixLength   = len(chainPrefix)
+)
+
+// Generates a chain name to be used with iptables.
+// Ensures that the generated chain name is less than
+// maxChainLength chars in length
+func FormatChainName(name string, id string) string {
+	chainBytes := sha512.Sum512([]byte(name + id))
```
alban: It might not be a problem in this specific case, but usually I think it is better to add a separator between fields being hashed to ensure collisions cannot be forged. If "_" is a character that is not allowed in name or id:
```
chainBytes := sha512.Sum512([]byte(name + "_" + id))
```

steveeJ: Can you explain further? AFAIU adding a constant to the input of a hash function doesn't change anything.

pkg/utils/utils_test.go
```diff
@@ -6,13 +6,34 @@ import (
...
+	})
+
+	It("must change when a character changes", func() {
+		chain1 := FormatChainName("testalongnamethatdoesnotmakesense", "1234")
+		chain2 := FormatChainName("testalongnamethatdoesnotmakesense", "1235")
+		Expect(len(chain1) == 29).To(Equal(true))
+		Expect(len(chain2) == 29).To(Equal(true))
+		Expect(chain1).To(Equal("CNI-374f33fe84ab0ed84dcdebe38"))
+		Expect(chain2).NotTo(Equal("CNI-374f33fe84ab0ed84dcdebe38"))
+		Expect(chain1).NotTo(Equal(chain2))
```
alban: Not sure it worth adding the hashes in the tests. Tesing `Expect(chain1).NotTo(Equal(chain2))` is good but the tests on the exact hash means that  whenever we change the hash function, we would need to update the tests...

pkg/utils/utils.go
```diff
@@ -5,7 +5,8 @@ import (
 	"fmt"
 )
 
-const MaxChainLength = 29 - len("CNI-")
+const ChainCNIPrefix = ("CNI-")
```
alban: no need for parenthesis

pkg/utils/utils.go
```diff
@@ -5,16 +5,17 @@ import (
...
+const (
+	maxChainLength = 29
+	chainPrefix    = "CNI-"
+	prefixLength   = len(chainPrefix)
+)
 
 // Generates a chain name to be used with iptables.
 // Ensures that the generated chain name is less than
-// 29 chars in length
+// maxChainLength chars in length
```
alban: less or equal?

pkg/utils/utils.go
alban: https://play.golang.org/p/HeGo6Ig66D

With your function,
```
	name := "Hello "
	id := "world!"
```
will have the same hash as:
```
	name := "Hello worl"
	id := "d!"
```

pkg/utils/utils_test.go
steveeJ: > whenever we change the hash function, we would need to update the tests...

I think we don't have much choice here, somehow you have to have at least one example tested. For changes to the hash functions that are not supposed to alter the result we need this.

alban:
https://github.com/appc/cni/pull/164#discussion_r57926235
> truncating a hash value will always increase the risk of collisions significantly. We need a hash function that produces a string output of 25 characters.

I don't think that's true: if it is a good hash function, truncating the hash to n characters should give the same risk of collisions as another hash function that produces directly n characters.

I would still prefer to have a part of the human readable name in the chain. We don't really need to have so many characters for the hash. But we can do that in another PR.

pkg/utils/utils.go
steveeJ: equal! thanks for catching

steveeJ:
> I would still prefer to have a part of the human readable name in the chain. We don't really need to have so many characters for the hash. But we can do that in another PR.

I'd prefer that too, a follow-up could use iptables comments module for that purpose.

steveeJ:
@achanda how did you evaluate the maximum chain length of 29 characters? My system insists of having 28 chars max for iptables chain names:

```
$ sudo -E $CNI_PATH/cnitool add mynet /proc/30674/ns/net
exit status 2: iptables v1.4.21: chain name `CNI-be1eedf7a76853f303ebd985b' too long (must be under 29 chars)
Try `iptables -h' or 'iptables --help' for more information.
```

pkg/ip/ipmasq.go
```diff
@@ -36,25 +36,25 @@ func SetupIPMasq(ipn *net.IPNet, chain string) error {
...
 // TeardownIPMasq undoes the effects of SetupIPMasq
-func TeardownIPMasq(ipn *net.IPNet, chain string) error {
+func TeardownIPMasq(ipn *net.IPNet, chain string, comment string) error {
 	ipt, err := iptables.New()
 	if err != nil {
 		return fmt.Errorf("failed to locate iptables: %v", err)
 	}
 
-	if err = ipt.Delete("nat", "POSTROUTING", "-s", ipn.String(), "-j", chain); err != nil {
+	if err = ipt.Delete("nat", "POSTROUTING", "-s", ipn.String(), "-j", chain, "-m", "comment", "--comment", comment); err != nil {
```
alban: Do you have to give the `--comment` when deleting too?

steveeJ: Yes, it's part of the rule spec and won't match without.

alban:
If you figure out what should really be the maximum chain length, LGTM.

Some say the [max is 28](https://review.openstack.org/#/c/23216/) but I don't see the source...

steveeJ:
@alban

> If you figure out what should really be the maximum chain length, LGTM.
> 
> Some say the max is 28 but I don't see the source...

The check [occurs within iptables](http://git.netfilter.org/iptables/tree/iptables.c?h=v1.4.21&id=5429b41c2bb4ac8fe672a1513a041c0ed0c241f6#n1879) and throws an error if the chain name length is `>=` [XT_EXTENSION_MAXNAMELEN](http://lxr.free-electrons.com/source/include/uapi/linux/netfilter/x_tables.h#L7) which equals 29.

PR #175, opened 2016-04-05
add support to set hairpin mode for bridge plugin

freehan:
Added `hairpin_mode` option for bridge plugin. After veth setup, `hairpin_mode` will be enable if `hairpin_mode` is set to true. We think it is natural for CNI to handle this.

### DIFF
--- a/Documentation/bridge.md
+++ b/Documentation/bridge.md
@@ -19,6 +19,7 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
 	"bridge": "mynet0",
 	"isGateway": true,
 	"ipMasq": true,
+	"hairpinMode": true,
 	"ipam": {
 		"type": "host-local",
 		"subnet": "10.10.0.0/16"
@@ -34,4 +35,5 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
 * `isGateway` (boolean, optional): assign an IP address to the bridge. Defaults to false.
 * `ipMasq` (boolean, optional): set up IP Masquerade on the host for traffic originating from this network and destined outside of it. Defaults to false.
 * `mtu` (integer, optional): explicitly set MTU to the specified value. Defaults to the value chosen by the kernel.
+* `hairpinMode` (boolean, optional): set hairpin mode for interfaces on the bridge. Defaults to false.
 * `ipam` (dictionary, required): IPAM configuration to be used for this network.
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -36,10 +36,11 @@ const defaultBrName = "cni0"
 
 type NetConf struct {
 	types.NetConf
-	BrName string `json:"bridge"`
-	IsGW   bool   `json:"isGateway"`
-	IPMasq bool   `json:"ipMasq"`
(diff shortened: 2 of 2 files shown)


philips:
This makes sense to me. Do you think there will ever be something other than true/false for this?

cc @yifan-gu @sjpotter

Documentation/bridge.md
```diff
@@ -34,4 +35,5 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
 * `isGateway` (boolean, optional): assign an IP address to the bridge. Defaults to false.
 * `ipMasq` (boolean, optional): set up IP Masquerade on the host for traffic originating from this network and destined outside of it. Defaults to false.
 * `mtu` (integer, optional): explicitly set MTU to the specified value. Defaults to the value chosen by the kernel.
+* `hairpinMode` (boolean, optional): set hairpin mode for interfaces on the bridge. Defaults to false.
```
philips: It feels like people would expect hairpin, can you rationalize in this doc why it defaults to false?

freehan: I just want to preserve the current behavior.

plugins/main/bridge/bridge.go
```diff
@@ -151,9 +160,23 @@ func setupVeth(netns string, br *netlink.Bridge, ifName string, mtu int) error {
 		return fmt.Errorf("failed to connect %q to bridge %v: %v", hostVethName, br.Attrs().Name, err)
 	}
 
+	// set hairpin mode if enabled
+	if hairpinMode {
```
achanda: Can't we just call `netlink.LinkSetHairpin(hostveth)` here?
https://godoc.org/github.com/vishvananda/netlink#LinkSetHairpin

freehan: Good Call. Will submit a patch.

Documentation/bridge.md
steveeJ: I'd also keep the current default. IMHO the developers who use CNI for their projects will make sure their users are not surprised by setting this accordingly.

plugins/main/bridge/bridge.go
```diff
@@ -32,14 +32,21 @@ import (
 	"github.com/vishvananda/netlink"
 )
 
-const defaultBrName = "cni0"
+const (
+	defaultBrName           = "cni0"
```
steveeJ: AFAICT the constants below are not used anymore. Please revert changes to this section to the previous state.

dcbw:
Latest LGTM, we'll use this for Kubernetes' kubenet module once it lands.  We'll want another CNI release soon though :)

steveeJ:
@freehan would you feel comfortable writing a functional test for this use-case? No worries if not, then someone else (probably me in this case) will add them.

freehan:
@steveeJ I could not find any existing test for bridge plugin though. I think it is better for you guys to lay the groundwork.

dcbw:
@freehan I've got a bunch in https://github.com/appc/cni/pull/167.  There's a lot of PRs touching the same code now, which is making rebasing due to conflicts a pain...

steveeJ:
LGTM

/cc @tomdee

steveeJ:
This was merged, thanks @freehan!

PR #176, opened 2016-04-06
ns: fix reading net namespace in multi-threaded processes

dcbw:
/proc/self/ns/net gives the main thread's namespace, not necessarily
the namespace of the thread that's running the testcases.  This causes
sporadic failures of the tests.  For example, with a testcase reading
inodes after switching netns:

/proc/27686/task/27689/ns/net 4026532565
/proc/self/ns/net 4026531969
/proc/27686/task/27689/ns/net 4026532565

See also:
https://github.com/vishvananda/netns/commit/008d17ae001344769b031375bdb38a86219154c6

```
Running Suite: pkg/ns Suite
===========================
Random Seed: 1459953577
Will run 6 of 6 specs

• Failure [0.028 seconds]
Linux namespace operations
/cni/gopath/src/github.com/appc/cni/pkg/ns/ns_test.go:167
  WithNetNS
  /cni/gopath/src/github.com/appc/cni/pkg/ns/ns_test.go:166
    executes the callback within the target network namespace [It]
    /cni/gopath/src/github.com/appc/cni/pkg/ns/ns_test.go:97

    Expected
        <uint64>: 4026531969
    to equal
        <uint64>: 4026532565

    /cni/gopath/src/github.com/appc/cni/pkg/ns/ns_test.go:96
------------------------------
•••••

Summarizing 1 Failure:

[Fail] Linux namespace operations WithNetNS [It] executes the callback within the target network namespace
/cni/gopath/src/github.com/appc/cni/pkg/ns/ns_test.go:96

Ran 6 of 6 Specs in 0.564 seconds
FAIL! -- 5 Passed | 1 Failed | 0 Pending | 0 Skipped --- FAIL: TestNs (0.56s)
FAIL
```

### DIFF
--- a/pkg/ns/ns_test.go
+++ b/pkg/ns/ns_test.go
@@ -28,13 +28,9 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-const CurrentNetNS = "/proc/self/ns/net"
-
 var _ = Describe("Linux namespace operations", func() {
 	Describe("WithNetNS", func() {
 		var (
-			originalNetNS *os.File
-
 			targetNetNSName string
 			targetNetNSPath string
 			targetNetNS     *os.File
@@ -42,8 +38,6 @@ var _ = Describe("Linux namespace operations", func() {
 
 		BeforeEach(func() {
 			var err error
-			originalNetNS, err = os.Open(CurrentNetNS)
-			Expect(err).NotTo(HaveOccurred())
 
 			targetNetNSName = fmt.Sprintf("test-netns-%d", rand.Int())
 
@@ -60,8 +54,6 @@ var _ = Describe("Linux namespace operations", func() {
 
 			err := exec.Command("ip", "netns", "del", targetNetNSName).Run()
 			Expect(err).NotTo(HaveOccurred())
-
-			Expect(originalNetNS.Close()).To(Succeed())
 		})
 
 		It("executes the callback within the target network namespace", func() {
@@ -71,7 +63,7 @@ var _ = Describe("Linux namespace operations", func() {
 			var actualInode uint64
 			var innerErr error
 			err = ns.WithNetNS(targetNetNS, false, func(*os.File) error {
-				actualInode, innerErr = testhelpers.GetInode(CurrentNetNS)
+				actualInode, innerErr = testhelpers.GetInodeCurNetNS()
 				return nil
 			})
 			Expect(err).NotTo(HaveOccurred())
@@ -81,7 +73,7 @@ var _ = Describe("Linux namespace operations", func() {
 		})
 
 		It("provides the original namespace as the argument to the callback", func() {
-			hostNSInode, err := testhelpers.GetInode(CurrentNetNS)
+			hostNSInode, err := testhelpers.GetInodeCurNetNS()
 			Expect(err).NotTo(HaveOccurred())
 
 			var inputNSInode uint64
@@ -97,7 +89,7 @@ var _ = Describe("Linux namespace operations", func() {
 		})
 
 		It("restores the calling thread to the original network namespace", func() {
-			preTestInode, err := testhelpers.GetInode(CurrentNetNS)
+			preTestInode, err := testhelpers.GetInodeCurNetNS()
 			Expect(err).NotTo(HaveOccurred())
 
 			err = ns.WithNetNS(targetNetNS, false, func(*os.File) error {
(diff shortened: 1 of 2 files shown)


dcbw:
@steveeJ @jonboulle @rosenhouse took a while to debug these failures, but I think this is the fix.

zachgersh:
@dcbw thanks for this. One thing of note now, I think this means that this set of tests now can no longer be run in parallel due to locking to a particular thread.

dcbw:
@zachgersh yeah, we've got to take that hit unfortunately :(

steveeJ:
Naiv question, why isn't it enough to put `runtime.LockOSThread()` in the preamble of the the testsuite? I imagine the suite will be run by the same goroutine, and if we lock that to a thread it should be safe, or not? Happy to receive a patient explanation :-)

steveeJ:
Also, could we just use https://github.com/vishvananda/netns instead of maintaining something similar here?

dcbw:
> Naiv question, why isn't it enough to put runtime.LockOSThread() in the preamble of the the testsuite? I imagine the suite will be run by the same goroutine, and if we lock that to a thread it should be safe, or not? Happy to receive a patient explanation :-)

Locking to an OS thread doesn't guarantee *which* thread you're locked to.  /proc/self/ns/net only tells you what the main thread's namespace is, but the current thread may not be the main thread, depending on where Go schedules you and whatever Ginkgo does.

steveeJ:
@dcbw 

>> Naiv question, why isn't it enough to put runtime.LockOSThread() in the preamble of the the testsuite? I imagine the suite will be run by the same goroutine, and if we lock that to a thread it should be safe, or not? Happy to receive a patient explanation :-)
> 
> Locking to an OS thread doesn't guarantee which thread you're locked to. /proc/self/ns/net only tells you what the main thread's namespace is, but the current thread may not be the main thread, depending on where Go schedules you and whatever Ginkgo does.

Thanks, sounds reasonable to assume Ginkgo runs things in parallel from the beginning. :+1: 
Any thoughts on using vishvanananda's netns package?

pkg/ns/ns_test.go
```diff
@@ -57,8 +64,12 @@ var _ = Describe("Linux namespace operations", func() {
 		)
 
 		BeforeEach(func() {
+			runtime.LockOSThread()
+			defer runtime.UnlockOSThread()
+
 			var err error
-			originalNetNS, err = os.Open(CurrentNetNS)
+			name := fmt.Sprintf("/proc/%d/task/%d/ns/net", os.Getpid(), syscall.Gettid())
```
steveeJ: would you mind using a constant for this format string?

pkg/ns/ns_test.go
```diff
@@ -29,6 +31,13 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+func getCurNSInode() (uint64, error) {
+	// /proc/self/ns/net returns the namespace of the main thread, not
+	// of whatever thread this goroutine is running on.  Make sure we
+	// use the thread's net namespace since the thread is switching around
+	return getInode(fmt.Sprintf("/proc/%d/task/%d/ns/net", os.Getpid(), syscall.Gettid()))
```
steveeJ: would you mind using a constant for this format string?

steveeJ: alternatively a function, as we're using it twice within the test suite

dcbw:
> Any thoughts on using vishvanananda's netns package?

@steveeJ I looked into it yesterday and it is possible, but it makes a few things harder because it's not quite a match for the testing infrastructure PRs I've submitted.  I'd like to explore doing that after this PR and the testing ones are merged if that's OK?

dcbw:
@steveeJ pulled out the function and repushed, PTAL thanks!

dcbw:
Another test flake?

steveeJ:
> Another test flake?
> 
> Bad response status from coveralls: 422 - {"message":"Couldn't find a repository matching this job.","error":true}

Yeah, we've been getting those but I haven't found the reason. Pressing rebuild on travis has always lead to a successful build.

steveeJ:
LGTM

@rosenhouse would you please take a look at this too? Since you wrote the initial test.

rosenhouse:
@steveeJ Just seeing this now.  Will take a look soon!

zachgersh:
hey @dcbw - you may want to have a look at the test-helpers I just merged from @rosenhouse.

rosenhouse:
@dcbw @steveeJ 👍  sorry about the delay.

dcbw:
@rosenhouse @steveeJ @zachgersh rebased on top of master and moved a few things around. PTAL, thanks!

PR #180, opened 2016-04-11
README: add some more introductory material

philips:
Changes:

4c778a6 (Brandon Philips, 51 seconds ago)
   README: add some material on what CNI isn't

   We get some questions about this stuff from time to time and I want to
   acknowledge we have thought about them but aren't currently tackling them.

d5a8ec9 (Brandon Philips, 79 seconds ago)
   README: add some more details on what CNI is

   We needed a bit more intro material to scope what CNI does. Here is my
   attempt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,9 @@
 CNI, the _Container Network Interface_, is a proposed standard for configuring network interfaces for Linux application containers.
 The standard consists of a simple specification for how executable plugins can be used to configure network namespaces; this repository also contains a go library implementing that specification.
 
+CNI concerns itself with setting up network connectivity of containers and removing allocated resources when the container is deleted.
+Because of this focus CNI has a wide range of support and the specification is simple to implement.
+
 The specification itself is contained in [SPEC.md](SPEC.md).
 
 ## Why develop CNI?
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -139,6 +142,16 @@ lo        Link encap:Local Loopback
...
 
+## What might CNI do in the future?
+
+CNI currently covers a wide range of needs for network configuration due to it simple model and API.
+However, in the future CNI might want to branch out into other directions:
+
+- Dynamic updates to existing network configuration
+- Dynamic policies for network bandwidth and firewall rules
+
+If these topics of are interest please contact the team via the mailing list or IRC and find some like minded people in the community to put a proposal together.
```
squaremo: Are these the only areas for which proposals will be accepted? If not, a "for example" may be in order.

squaremo:
Good idea, @philips. I like the statement of scope you've got here.

I am less sure about the future directions -- my inclination would be to indicate what the criteria would be (in particular what things would count _against_ proposals), rather than specific areas.

Since we don't have a firm idea of those criteria yet, giving some examples is certainly an improvement. So: LGTM.

PR #182, opened 2016-04-18
Document use of goroutine and LockOSThread in test helpers

rosenhouse:
Docs to accompany https://github.com/appc/cni/pull/181


rosenhouse:
cc: @zachgersh

PR #185, opened 2016-04-18
#issue-184

asridharan:
Modified flannel plugin to add default routes for `delegate` IPAM plugin.


jellonek:
I leaved my comment in original issue.

asridharan:
Closing this PR, since we need to come up with a better way to `delegate` routes from `flannel` to IPAM.

PR #186, opened 2016-04-20
README: clarify what CNI is

steveeJ:
The word "standard" does not represent what CNI is actually trying to
be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,13 +5,15 @@
 
 ## What is CNI?
 
-CNI, the _Container Network Interface_, is a proposed standard for configuring network interfaces for Linux application containers.
-The standard consists of a simple specification for how executable plugins can be used to configure network namespaces; this repository also contains a go library implementing that specification.
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,14 +5,17 @@
 
 ## What is CNI?
 
-CNI, the _Container Network Interface_, is a proposed standard for configuring network interfaces for Linux application containers.
-The standard consists of a simple specification for how executable plugins can be used to configure network namespaces; this repository also contains a go library implementing that specification.
-
+CNI, the _Container Network Interface_, is a project for configuring network interfaces for Linux application containers.
```
squaremo: How about something like "CNI [...] is a specification for container networking plugins. It defines how a plugin is invoked by an application to configure a network namespace. This repository also contains a Go library implementing the specification on both sides (the application, and the plugin)."

squaremo:
Yes, good to remove "standard", and emphasise "specification". Thanks @steveeJ

PR #189, opened 2016-04-22
Remove reference to policy on accepting PRs

tomdee:
There is to policy


squaremo:
Good point :) As I don't think we've developed such a policy, better to omit it. Thanks Tom!

PR #195, opened 2016-05-03
plugins/bridge: clean masquerading rules

iaguis:
In the Add command we set up masquerading rules that didn't have a
corresponding clean-up code in Del.

Add the clean-up code.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -238,14 +238,25 @@ func cmdDel(args *skel.CmdArgs) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


dcbw:
LGTM

yifan-gu:
cc @steveeJ @euank

squaremo:
(NB in the absence of tests here, I went through the set-up described in the rkt bug mentioned above using cnitool, and confirmed that the iptables-save output showed that rules were removed when the network was deleted).

steveeJ:
Thanks @iaguis!

PR #196, opened 2016-05-03
generate and configure veth mac address based on assigned ip

### LINKED ISSUE #194: host-local ipam should allocate round-robin ips
host-local ipam tends to allocate ips that are recently used. This sometimes cause unexpected behavior related to arp caching. 

If this sounds like a good idea, I will put together a PR. 

https://github.com/kubernetes/kubernetes/issues/24953

freehan:
plugins: generate and configure veth mac address based on assigned ip for bridge plugin

`host-local` ipam search for available ips from the start of subnet. Hence it tends to allocate ip addresses that are recently used. But each new veth gets a different mac address. This cause unexpected behavior related to arp caching. For instance, a new container with an ip address that was recently released may not be accessible due to the difference between the container's mac address and the requester's arp cache. 

This PR solves the above problem by generate and configure veth mac address based on assigned ip. This ensures the 1:1 mapping between ip and mac address.

Fixes #194

### DIFF
--- a/pkg/ipam/ipam.go
+++ b/pkg/ipam/ipam.go
@@ -23,6 +23,17 @@ import (
 	"github.com/containernetworking/cni/pkg/types"
 
 	"github.com/vishvananda/netlink"
+	"net"
+	"strconv"
+	"strings"
+)
+
+const (
+	// private mac prefix safe to use
+	privateMACPrefix = "0a:58"
+
+	// veth link dev type
+	vethLinkType = "veth"
 )
 
 func ExecAdd(plugin string, netconf []byte) (*types.Result, error) {
@@ -45,6 +56,17 @@ func ConfigureIface(ifName string, res *types.Result) error {
 		return fmt.Errorf("failed to set %q UP: %v", ifName, err)
 	}
 
+	// only set hardware address to veth when using ipv4
+	if link.Type() == vethLinkType && res.IP4 != nil {
+		hwAddr, err := generateHardwareAddr(res.IP4.IP.IP)
+		if err != nil {
+			return fmt.Errorf("failed to generate hardware addr: %v", err)
+		}
+		if err = netlink.LinkSetHardwareAddr(link, hwAddr); err != nil {
+			return fmt.Errorf("failed to add hardware addr to %q: %v", ifName, err)
+		}
+	}
+
 	// TODO(eyakubovich): IPv6
 	addr := &netlink.Addr{IPNet: &res.IP4.IP, Label: ""}
 	if err = netlink.AddrAdd(link, addr); err != nil {
@@ -65,4 +87,22 @@ func ConfigureIface(ifName string, res *types.Result) error {
 	}
 
 	return nil
+}
+
+// generateHardwareAddr generates 48 bit virtual mac addresses based on the IP input.
+func generateHardwareAddr(ip net.IP) (net.HardwareAddr, error) {
+	if ip.To4() == nil {
+		return nil, fmt.Errorf("generateHardwareAddr only support valid ipv4 address as input")
+	}
+	mac := privateMACPrefix
+	sections := strings.Split(ip.String(), ".")
+	for _, s := range sections {
+		i, _ := strconv.Atoi(s)
+		mac = mac + ":" + fmt.Sprintf("%02x", i)
+	}
+	hwAddr, err := net.ParseMAC(mac)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to parse mac address %s generated based on ip %s due to: %v", mac, ip, err)
+	}
+	return hwAddr, nil
 }
--- a/pkg/ipam/ipam_test.go
+++ b/pkg/ipam/ipam_test.go
@@ -1,0 +1,62 @@
+// Copyright 2016 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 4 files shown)


steveeJ:
Thanks for this @freehan. Can you add a short message to the commit that explains what we gain from this? I think it's also worth documenting the behavior.

steveeJ:
@freehan ping, just waiting on an updated commit message for this change.

jonboulle:
@steveej I suggest using the github squash feature to fix the commit
message as you merge

steveeJ:
@jonboulle thanks, for the conformance that will help speed things up in the future. Although in this case I actually don't know the whole story, so it'd be helpful if the author would add an explanatory message.

freehan:
@steveeJ Thanks for noticing. I was working on another PR aimming to solve the same problem with a better approach. I got distracted and will push it soon. 

I believe we may want both PRs.

freehan:
@steveeJ ready for review

pkg/ipam/ipam.go
```diff
@@ -45,6 +51,14 @@ func ConfigureIface(ifName string, res *types.Result) error {
 		return fmt.Errorf("failed to set %q UP: %v", ifName, err)
 	}
 
+	hwAddr, err := generateHardwareAddr(res.IP4.IP.IP)
```
steveeJ: This code should also handle an empty IP4 in case there's IP6 only.

freehan: What about returning an error on empty IP4?

pkg/ipam/ipam.go
```diff
@@ -45,6 +51,14 @@ func ConfigureIface(ifName string, res *types.Result) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to generate hardware addr: %v", err)
+	}
+	if err = netlink.LinkSetHardwareAddr(link, hwAddr); err != nil {
+		return fmt.Errorf("failed to add hardware addr to %q: %v", ifName, err)
+	}
+
 	// TODO(eyakubovich): IPv6
 	addr := &netlink.Addr{IPNet: &res.IP4.IP, Label: ""}
```
freehan: it looks like the code is assuming ipv4 only.

freehan:
@steveeJ PTAL
Thanks!

steveeJ:
@freehan in case you haven't noticed, tests are failing. Did you want me to take a look to find out why?

freehan:
@steveeJ Oh my bad. I leave the message before the tests finished. Will fix it and come back. Thanks

freehan:
@steveeJ Ready for review. Thanks!

pkg/ipam/ipam_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"net"
+)
+
+var _ = Describe("ipam utils", func() {
+	Context("Generate Hardware Addrress", func() {
+		It("generate hardward address based on ipv4 address", func() {
```
steveeJ: typos

pkg/ipam/ipam_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+		It("return error if input is not ipv4 address", func() {
+			testCases := []net.IP{
+				net.ParseIP(""),
+				net.ParseIP("2001:db8:0:1:1:1:1:1"),
+			}
+			for _, tc := range testCases {
+				_, err := generateHardwareAddr(tc)
+				Expect(err).To(HaveOccurred())
+			}
```
steveeJ: Is this a specific error we could match against?

pkg/ipam/ipam_test.go
freehan: fixed

pkg/ipam/ipam_test.go
freehan: fixed

pkg/ipam/ipam_test.go
```diff
@@ -0,0 +1,62 @@
...
+		})
+
+		It("return error if input is not ipv4 address", func() {
+			testCases := []net.IP{
+				net.ParseIP(""),
+				net.ParseIP("2001:db8:0:1:1:1:1:1"),
+			}
+			for _, tc := range testCases {
+				_, err := generateHardwareAddr(tc)
+				Expect(err.Error()).To(Equal("generateHardwareAddr only support valid ipv4 address as input"))
```
steveeJ: Instead of matching a string, I would suggest to introduce new error types. See https://github.com/containernetworking/cni/commit/35f3a090b2f7ebe32daf0c6a73eb060c669e2128 for an example. I'm not insisting, but IMHO this improves code quality a lot.

freehan:
@steveeJ  not sure why coverage decrease :(

pkg/ipam/ipam.go
```diff
@@ -45,6 +51,17 @@ func ConfigureIface(ifName string, res *types.Result) error {
 		return fmt.Errorf("failed to set %q UP: %v", ifName, err)
 	}
 
+	// only set hardware address for ipv4
+	if res.IP4 != nil {
+		hwAddr, err := generateHardwareAddr(res.IP4.IP.IP)
+		if err != nil {
+			return fmt.Errorf("failed to generate hardware addr: %v", err)
+		}
+		if err = netlink.LinkSetHardwareAddr(link, hwAddr); err != nil && err.Error() != "operation not supported" {
```
steveeJ: this isn't very clean, why are we ignoring "operation not supported"?

steveeJ:
> @steveeJ not sure why coverage decrease :(

That's easy to explain: you added code that is not covered by tests.

freehan:
Got caught up by something else. Ready for review. Thanks!

pkg/ipam/ipam_test.go
freehan: This error was kind of unique. So I did not introduce another error type to carry it.

PR #199, opened 2016-05-04
bridge: leave TX queue length as kernel default, not 0

dcbw:
Not using NewLinkAttrs() or not initializing TxQLen leaves
the value as 0, which tells the kernel to set a zero-length
tx_queue_len.  That messes up FIFO traffic shapers (like pfifo)
that use the device TX queue length as the default packet
limit.  This leads to a default packet limit of 0, which drops
all packets.

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -101,6 +101,11 @@ func ensureBridge(brName string, mtu int) (*netlink.Bridge, error) {
 		LinkAttrs: netlink.LinkAttrs{
 			Name: brName,
 			MTU:  mtu,
+			// Let kernel use default txqueuelen; leaving it unset
+			// means 0, and a zero-length TX queue messes up FIFO
+			// traffic shapers which use TX queue length as the
+			// default packet limit
+			TxQLen: -1,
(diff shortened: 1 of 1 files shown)


dcbw:
@steveeJ @jonboulle

dcbw:
Travis failure seems to be a test flake:

`Bad response status from coveralls: 422 - {"message":"Couldn't find a repository matching this job.","error":true}`

maybe an issue with the recent rename from appc/cni?

dcbw:
cc @rosenhouse

steveeJ:
LGTM,

with one note: there is an [example in the netlink README](https://github.com/vishvananda/netlink/blob/adb0f53af689dd38f1443eba79489feaacf0b22e/README.md#examples) about this. It is shown that the constructor `netlink.NewLinkAttrs()` would imply defaults. @dcbw we might switch to using that constructor in case any other defaults will be there in the future. OTOH I'm a friend of explicitness, so I'm fine with merging as is.

dcbw:
@steveeJ I'd rather keep it how we have now, and investigate moving to NewLinkAttrs() later.

PR #205, opened 2016-05-11
plugins/{ptp,bridge}: teardown first

### LINKED ISSUE #204: *: teardown for rebooted machines leaks IP addresses
Note without a deep investigation but this could be critical, so better safe than sorry.

The ptp and bridge (#18) plugins fail to delete the IPAM allocations in case the teardown happens after the machine has been rebooted. Interfaces and iptables rules will be gone, wheres the IPAM state will remain on disk and never be cleaned due to previous failures.

/cc @iaguis

steveeJ:
This will allow the IPAM allocations to be cleared in case the
interfaces and iptables rules are non-existent.

Fixes #204 

/cc @iaguis @tomdee

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -238,6 +238,10 @@ func cmdDel(args *skel.CmdArgs) error {
 		return err
 	}
 
+	if err := ipam.ExecDel(n.IPAM.Type, args.StdinData); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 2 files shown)


jonboulle:
Ripam?

tmrts:
It's missing import statements

steveeJ:
@jonboulle thanks, fixed.

tmrts:
LGTM

dcbw:
lgtm

PR #210, opened 2016-05-18
Send ClientID DHCP option in DISCOVER/ACK msgs v2

fnordahl:
This is a continuation of the work in #83 

Please note that it is a work in progress and I need to add tests to actually verify that my hypothesis for a fix holds water.

### DIFF
--- a/Documentation/ipvlan.md
+++ b/Documentation/ipvlan.md
@@ -38,3 +38,31 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
 Therefore the container will not be able to reach the host via `ipvlan` interface.
 Be sure to also have container join a network that provides connectivity to the host (e.g. `ptp`).
 * A single master interface can not be enslaved by both `macvlan` and `ipvlan`.
+
+### Using ipvlan and dhcp
+There are two important considerations when using the "ipvlan" plugin together with the "dhcp" IPAM plugin.
+
+First, the kernel ipvlan driver uses the IP packet's destination address when choosing the slave interface to which it should be forwarded.
+The driver cannot perform this forwarding if the container interface has not yet been assigned an IP address.
+Since the DHCP packets are exchanged to determine which IP should be used, the interface will not yet have an IP address.
+DHCP servers typically unicast replies back to the client.
+The ipvlan driver will drop these replies.
+To work around this, the DHCP client must instruct the DHCP server to use broadcast replies instead.
+This can be accomplished by setting the "broadcast" flag in the "ipam" section of the network configuration:
+
+```
+{
+	"name": "mynet",
+	"type": "ipvlan",
+	"master": "eth0",
+	"ipam": {
+		"type": "dhcp",
+		"broadcast": true
+	}
+}
+```
+
+Second, because all ipvlan interfaces share the MAC address of the master interface, DHCP servers will have trouble identifying and tracking individual container interfaces.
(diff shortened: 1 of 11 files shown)


Documentation/ipvlan.md
```diff
@@ -38,3 +38,31 @@ Because all ipvlan interfaces share the MAC address with the host interface, DHC
...
+This can be accomplished by setting the "broadcast" flag in the "ipam" section of the network configuration:
+
+```
+{
+	"name": "mynet",
+	"type": "ipvlan",
+	"master": "eth0",
+	"ipam": {
+		"type": "dhcp",
+		"broadcast": true
```
dcbw: Should ipvlan force broadcast to true?  Or should that be left up to the caller of the CNI plugin to handle?

dcbw:
LGTM

fnordahl:
Thank you for your feedback @dcbw 

Currently doing some r&d for how to best create tests for this here: fnordahl/flux-capacitor#1

I will be back with tests suitable for merging with this PR soon.

fnordahl:
@steveeJ I have added unit tests.

I have also added functional tests, but they need a few more adjustments before they work. I have to toss in the towel now. Getting back at it tomorrow.

fnordahl:
I think we need a newer or patched kernel/ipvlan driver to make the functional tests work. Specifically this patch: https://patchwork.ozlabs.org/patch/455225/

Travis-Ci image is currently running Ubuntu 14.04 with the 3.19.0-30 kernel from 15.04 (Vivid). I have the tests working on a newer system here.

Possible strategies going forward:
- [ ] Do we have access to resources elsewhere to run these tests (VM on GCE?)
- [ ] Backport the patch to 3.19.0 and compile the ipvlan module on Travis-Ci before running the tests
- [ ] Make Travis-Ci provide updated images

Documentation/ipvlan.md
fnordahl: Good point @dcbw . The current codebase does not seem to have relations to allow this to be deduced automatically, but we sure should aim at making that a reality.

dcbw:
@fnordahl yeah, you'll likely need some of those patches.  3.19 is so, so very old; in fact ipvlan was only added for the 3.19 kernel so it's entirely unsurprising that the driver would have bugs.

dcbw:
@fnordahl but instead of my patches, Mahesh solved the issue a different way.  You'd want:

ba35f8588f474d5bf8988615b04ee722a2684fd2
f631c44bbe1581d18d0aba628b3802d0eacb6373

fnordahl:
> @fnordahl but instead of my patches, Mahesh solved the issue a different way. You'd want:
> 
> ba35f8588f474d5bf8988615b04ee722a2684fd2
> f631c44bbe1581d18d0aba628b3802d0eacb6373

Thanks @dcbw ! f631c44bbe1581d18d0aba628b3802d0eacb6373 applies clean on the 3.19 sources and it suffices for the tests to work.

Patch updated, and now with consistently successful functional tests!

squeed:
Hm, it looks like this took a year-long nap. What is needed to bring it back to life?

fnordahl:
@squeed feedback. Previous feedback was missing tests, functional tests has since been added but the PR has been ignored nonetheless. Code is probably still good but needs a re-base.

Should I interpret your query as a sign of interest in having this merged?

squeed:
Indeed, I think a more well-rounded DHCP plugin would be a very good idea. The lack of DNS is also frustrating.

However, the repository has split, so please consider re-submitting for the Plugins repo.

Issue #183, opened 2016-04-18
Discussion: removing lockThread from WithNetNS?


rosenhouse:
## Summary
I believe that we can provide `WithNetNS` and `WithNetNSPath` functions that remove the `lockThread` argument and completely hide from the caller the complexity of goroutine thread affinity. This could let us simplify a lot of our plugin and test code.  I'm opening this issue to gather feedback before I assemble a PR. 

## Background
In our network namespace switching package, we provide [`WithNetNS`](https://github.com/appc/cni/blob/master/pkg/ns/ns.go#L71)
```go
func WithNetNS(ns *os.File, lockThread bool, f func(*os.File) error) error
```
which runs a function `f` inside a network namespace `ns`.

The middle argument, `lockThread` is easily misused. 

If the caller passes `false` then she must be sure that she has called `runtime.LockOSThread` beforehand.  Many of the CNI plugins use this approach, but it requires care, especially when creating goroutines.

Conversely, if the caller passes `true`, then that must be the *only* call using `true` within the call-stack.  This is because `Lock` and `UnlockOSThread` do not reference count, so a single `Unlock` will unlock **all** prior `Lock` calls.  That can leave a caller unlocked without knowing it.

Incorrect usage either way can lead to all sorts of hard-to-debug undefined behavior as goroutines end up in unexpected namespaces.

## Proposed change
Replace `WithNetNS` in `pkg/ns` with the following
```go
// WithNetNS synchronously executes toRun within the given network namespace.
//
// It does not affect the network namespace of the calling goroutine
// nor does it modify the thread-affinity of the calling goroutine.
// The caller does not need to call runtime.LockOSThread()
func WithNetNS(netNS *os.File, toRun func() error) error {
	containedCall := func() error {
		originalNetNS, err := os.Open(fmt.Sprintf("/proc/%d/task/%d/ns/net", os.Getpid(), syscall.Gettid()))
		if err != nil {
			return fmt.Errorf("opening current netns: %s", err)
		}
		defer originalNetNS.Close()

		if err = SetNS(netNS, syscall.CLONE_NEWNET); err != nil {
			return fmt.Errorf("switching to netns %v: %v", netNS.Name(), err)
		}
		defer SetNS(originalNetNS, syscall.CLONE_NEWNET) // switch back

		return toRun()
	}

	var wg sync.WaitGroup
	wg.Add(1)

	var innerError error
	go func() {
		defer wg.Done()
		runtime.LockOSThread()
[...]

PR #211, opened 2016-05-18
Fix namespace switch issues and add ipvlan, macvlan, and bridge e2e testing

### LINKED ISSUE #179: {mac,ip}vlan and bridge plugins need tests

dcbw:
A subset of https://github.com/containernetworking/cni/pull/167 while we discuss how the mocking should actually work.

Fixes #179.


dcbw:
cc @rosenhouse

dcbw:
cc @steveeJ

dcbw:
@rosenhouse one thing I'm not 100% sure about is the removal of LockOSThread() from the plugins init() functions themselves; that all works for the testcases because they run all the checks with Do(), and they run the plugin's cmdAdd/cmdDel from inside the Do() for the original namespace too.  But when the plugins are run standalone from kubernetes or rkt or whatever, should they get the current host namespace and wrap inital setup with Do()?

steveeJ:
> But when the plugins are run standalone from kubernetes or rkt or whatever, should they get the current host namespace and wrap inital setup with Do()?
AFAIU this shouldn't be a problem, reasoning below.

If the plugins are invoked, the given network namespace will be explicitly set whenever necessary using `NetNS.Do()`. AFAICT the assumption is that the rest of the code is supposed to run in the caller's netns, and unless the caller doesn't mess with setting the netns themselves, CNI shouldn't be messing with it either.

dcbw:
@steveeJ yeah, that should be true.  I guess the testcases have issues here (which were fixed by wrapping everything in Do()) because they create a completely new namespace to ensure they don't affect the host.  So I guess we're OK.

steveeJ:
> yeah, that should be true. I guess the testcases have issues here (which were fixed by wrapping everything in Do()) because they create a completely new namespace to ensure they don't affect the host. So I guess we're OK.

Agreed! At this point, thank you again for this excellent contribution. I lack the overview if all of @rosenhouse's issues related to the code are addressed, @dcbw can you help here?

dcbw:
rebased to resolve the merge conflict

dcbw:
@steveeJ @rosenhouse I believe they were all addressed in the previous mocking PR; the PR here doesn't change anything from that other PR except for dropping the 'netops' object.  Should get @rosenhouse to confirm though.

steveeJ:
As discussed in the CNI meeting, this will be merged after bringing the locking back to the plugins `init()` functions.

dcbw:
@steveeJ reverted the init() changes; PTAL.

rosenhouse:
So, like I said at the CNI meeting, I'm traveling on PTO, and then at a conference and therefore unable to look at this closely until late next week.

However, at first glance, I think I see instances of the issue I mentioned on the call: given how we're trying to isolate namespace switches to a goroutine, nested calls to `netns.Do` are problematic.  Specifically, you've got test code which calls `netns.Do` and then inside its callback, you directly invoke the plugin code.  And inside the plugin code (e.g. in the bridge plugin) there is another call to `netns.Do`.  That is likely the cause of the test flakes you're seeing.  Calling `LockOSThread` at the start of the test suite, as I suggested on the call, won't help here.

I think the easiest way to fix those flakes would be to invoke the plugins as a separate OS process, rather than running them in the same process as the tests.

I'll leave it up to the other maintainers to decide on merging.

steveeJ:
> reverted the init() changes; PTAL.

I've checked the test cases before the revert, so I trust the green lights with this now.

steveeJ:
@rosenhouse 
> Specifically, you've got test code which calls netns.Do and then inside its callback, you directly invoke the plugin code. And inside the plugin code (e.g. in the bridge plugin) there is another call to netns.Do. That is likely the cause of the test flakes you're seeing. Calling LockOSThread at the start of the test suite, as I suggested on the call, won't help here.

Let me try to repeat what you've explained to confirm and also lay the information open on the issue.
The first `Do()`'s goroutine is locked correctly, but in spite of this the nested goroutine could potentially be scheduled on a new OS thread because `GOMAXPROCS` has not been exhausted yet?

rosenhouse:
@steveeJ exactly -- that new goroutine may or may not end up on a new OS thread.  If a new OS thread is created, then it will inherit the namespaces of its parent.  So the new OS thread's "original" namespace is actually the "switched-to" special namespace of the first call.  From that point on, the newly created OS thread is "polluted".  

I've updated https://github.com/containernetworking/cni/issues/183 to try to describe this issue and recommend against nested calls to this type of namespace switching code.  It's still not clear to me how to best expose namespace switching at a higher level of abstraction in a safe way.

steveeJ:
> If a new OS thread is created, then it will inherit the namespaces of its parent. So the new OS thread's "original" namespace is actually the "switched-to" special namespace of the first call. From that point on, the newly created OS thread is "polluted".

This seems to go very deep down to Go's thread/routine handling mechanism and I would appreciate your source to your knowledge.

rosenhouse:
@steveeJ I don't have a written reference for the fact that threads inherit the namespace context of their parent thread.  However, another member of our CloudFoundry container networking team (@sykesm) has investigated this issue.  I can share the results of his work here.

He wrote code that switched the namespace of the main goroutine and spun up several more goroutines from there.  All the goroutines waited while the program reported the namespace associated with each OS thread.  ([code here](https://gist.github.com/sykesm/020a27341cca5169250990d250b25db4))  Here are the results:
```
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15832/ns/net -> net:[4026533719]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15833/ns/net -> net:[4026533719]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15834/ns/net -> net:[4026533719]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15835/ns/net -> net:[4026533719]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15836/ns/net -> net:[4026533719]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15838/ns/net -> net:[4026533719]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15839/ns/net -> net:[4026533719]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15840/ns/net -> net:[4026533719]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15841/ns/net -> net:[4026533940]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15842/ns/net -> net:[4026533940]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15843/ns/net -> net:[4026533940]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15844/ns/net -> net:[4026533940]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15845/ns/net -> net:[4026533940]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15846/ns/net -> net:[4026533940]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15848/ns/net -> net:[4026533719]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/task/15849/ns/net -> net:[4026533719]
lrwxrwxrwx 1 root root 0 May  9 16:12 /proc/15832/ns/net -> net:[4026533719]
```
As you can see, the first few threads all share the original netns of the process, but several others are on the new netns.

As I mentioned earlier, I don't have time to dig into this now.  But it might be interesting to perform a more detailed investigation to show how this state of the world evolves, perhaps reporting the full set of `thread:netns` pairings at each step of the loop.

pkg/ns/ns.go
```diff
@@ -15,64 +15,209 @@
...
 
-// SetNS sets the network namespace on a target file.
-func SetNS(f *os.File, flags uintptr) error {
-	_, _, err := syscall.RawSyscall(setNsNr, f.Fd(), flags, 0)
-	if err != 0 {
-		return err
-	}
+type NetNS interface {
+	// Executes the passed closure in this object's network namespace,
+	// restoring the original namespace afterwards. If the closure returns
```
steveeJ: as discussed OOB, we can't guarantee that the original namespace is restored. please add an explanation and maybe link to further docs, e.g. https://github.com/golang/go/wiki/LockOSThread

pkg/ns/README.md
```diff
@@ -0,0 +1,16 @@
+### Namespaces, Threads, and Go
+On Linux each OS thread can have a different network namespace.  Go's thread scheduling model switches goroutines between OS threads based on OS thread load and whether the goroutine would block other goroutines.  This can result in a goroutine switching network namespaces without notice and lead to errors in your code.
+
+### Namespace Switching
+Switching namespaces with the `ns.Set()` method is not recommended without additional strategies to prevent unexpected namespace changes when your goroutines switch OS threads.
+
+Go provides the `runtime.LockOSThread()` function to ensure a specific goroutine executes on its current OS thread and prevents any other goroutine from running in that thread until the locked one exits.  Careful usage of `LockOSThread()` and short goroutines can provide good control over which network namespace a given goroutine executes in.
```
steveeJ: "short goroutines" worries me a little, is that really necessary? A lock's purpose is to synchronize concurrent tasks and prevent any unwanted runtime behavior.

pkg/ns/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+For example, you cannot rely on the `ns.Set()` namespace being the current namespace after the `Set()` call unless the goroutine calling `Set()` has previously called `LockOSThread()`, and you ensure `runtime.UnlockOSThread()` is not called somewhere in-between.  You also cannot rely on the initial network namespace remaining the current network namespace of a goroutine if any other code in your program switches namespaces, unless you have already called `LockOSThread()` in that goroutine.  Note that `LockOSThread()` prevents the Go scheduler from optimally scheduling goroutines for best performance, so `LockOSThread()` should only be used in small, isolated goroutines that release the lock quickly.
```
steveeJ: second sentence is too long IMO, do you mind splitting it apart?

pkg/ns/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+### Do() The Recommended Thing
+The `ns.Do()` method provides control over network namespaces for you by implementing these strategies. All code dependent on a particular network namespace should be wrapped in the `ns.Do()` method to ensure the correct namespace is selected for the duration of your code.
```
steveeJ: shall we mention the tests as an example for usage? or should we give a short example here?

pkg/ns/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+For example, you cannot rely on the `ns.Set()` namespace being the current namespace after the `Set()` call unless you do two things.  First, the goroutine calling `Set()` must have previously called `LockOSThread()`.  Second, you must ensure `runtime.UnlockOSThread()` is not called somewhere in-between.  You also cannot rely on the initial network namespace remaining the current network namespace if any other code in your program switches namespaces, unless you have already called `LockOSThread()` in that goroutine.  Note that `LockOSThread()` prevents the Go scheduler from optimally scheduling goroutines for best performance, so `LockOSThread()` should only be used in small, isolated goroutines that release the lock quickly.
+
+### Do() The Recommended Thing
+The `ns.Do()` method provides control over network namespaces for you by implementing these strategies. All code dependent on a particular network namespace should be wrapped in the `ns.Do()` method to ensure the correct namespace is selected for the duration of your code.  For example:
+
```
steveeJ: please wrap the code snippet in ```

PR #216, opened 2016-05-20
plugins/ipam/dhcp/daemon: Remove Unix socket on exit

fnordahl:
Get rid of the need of removing it manually

### DIFF
--- a/plugins/ipam/dhcp/daemon.go
+++ b/plugins/ipam/dhcp/daemon.go
@@ -23,9 +23,11 @@ import (
 	"net/http"
 	"net/rpc"
 	"os"
+	"os/signal"
 	"path/filepath"
 	"runtime"
 	"sync"
+	"syscall"
 
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
@@ -126,7 +128,25 @@ func getListener() (net.Listener, error) {
 		if err := os.MkdirAll(filepath.Dir(socketPath), 0700); err != nil {
 			return nil, err
 		}
-		return net.Listen("unix", socketPath)
+		l, err := net.Listen("unix", socketPath)
+		if err == nil {
+			signalChan := make(chan os.Signal, 1)
+			signal.Notify(signalChan, os.Interrupt)
+			signal.Notify(signalChan, syscall.SIGTERM)
+			go func() {
+				signal := <-signalChan
+
+				signum := 0
+				switch signal := signal.(type) {
+				case syscall.Signal:
+					signum = int(signal)
+				}
+
+				l.Close()
+				os.Exit(128 + signum)
+			}()
+		}
+		return l, err
 
 	case len(l) == 1:
 		if l[0] == nil {
--- a/plugins/ipam/dhcp/daemon_test.go
+++ b/plugins/ipam/dhcp/daemon_test.go
@@ -1,0 +1,101 @@
+// Copyright 2015 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"os"
+	"path"
+	"syscall"
+	"testing"
+	"time"
+)
+
+func TestGetListener(t *testing.T) {
+	l, err := getListener()
+	if err != nil {
+		t.Errorf("Unable to get listener: %v", err)
+	}
+
+	// verify existence of dhcp socket
+	_, err = os.Stat(socketPath)
+	if err != nil {
+		t.Errorf("Unable to Stat socket %v: %v", socketPath, err)
+		return
+	}
+
+	// clean up
+	l.Close()
+}
+
+func TestSignalHandler(t *testing.T) {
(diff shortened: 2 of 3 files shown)


plugins/ipam/dhcp/daemon.go
```diff
@@ -126,6 +128,16 @@ func getListener() (net.Listener, error) {
 		if err := os.MkdirAll(filepath.Dir(socketPath), 0700); err != nil {
 			return nil, err
 		}
+		signalChan := make(chan os.Signal, 1)
+		signal.Notify(signalChan, os.Interrupt)
+		signal.Notify(signalChan, syscall.SIGTERM)
+		go func() {
+			select {
```
steveeJ: If you only have one case there is no need for select

fnordahl:
Updated version:
- [x] Removed superflous select {}
- [x] Added functional tests

fnordahl:
The coverage test still fails although I have added tests for the code I am adding. Am I being punished from already non-existing tests here? :-) ( @steveeJ )

fnordahl:
I guess coveralls does not detect that the code is tested since there are no relevant function calls in the test. I will update with code that simulates the process execution with fork + function call + sleep and then interrupt to see if that makes coveralls happy.

fnordahl:
Updated version:
- [x] Use net.Listener.Close() to remove the socket instead of os.Remove()
- [x] Exercise the getListener() function in a test prior to forking of a seperate process for testing the signal processing to make coveralls happy with the code addittion / test coverage ratio.

With this I believe I have addressed the issues that needed rework @steveeJ

plugins/ipam/dhcp/daemon.go
```diff
@@ -126,7 +128,18 @@ func getListener() (net.Listener, error) {
...
-		return net.Listen("unix", socketPath)
+		l, err := net.Listen("unix", socketPath)
+		if err == nil {
+			signalChan := make(chan os.Signal, 1)
+			signal.Notify(signalChan, os.Interrupt)
+			signal.Notify(signalChan, syscall.SIGTERM)
+			go func() {
+				<-signalChan
+				l.Close()
+				os.Exit(1)
```
jonboulle: better to do 128 + signal:
http://tldp.org/LDP/abs/html/exitcodes.html

plugins/ipam/dhcp/daemon_test.go
```diff
@@ -0,0 +1,80 @@
...
+		t.Errorf("Unable to start process: %v", err)
+		return
+	}
+
+	// wait for process initialization to finish
+	time.Sleep(1 * time.Second)
+
+	// verify existence of dhcp socket
+	_, err = os.Stat(socketPath)
+	if err != nil {
```
jonboulle: Better to be a little more discerning when checking stat errors, e.g. for this specific error message the condition should really be `os.IsNotExist(err)` - otherwise, for `err != nil`, a more general message would be appropriate ("error checking socket existence:" or so)

plugins/ipam/dhcp/daemon_test.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	var procAttr os.ProcAttr
+	proc, err := os.StartProcess(path.Join(cwd, "../../../bin/dhcp"), []string{"dhcp", "daemon"}, &procAttr)
+	if err != nil {
+		t.Errorf("Unable to start process: %v", err)
+		return
+	}
+
+	// wait for process initialization to finish
+	time.Sleep(1 * time.Second)
```
jonboulle: Sleeps are really bad in tests. Is there another way you can wait for the process to initialise? Something you could either block or poll on?

fnordahl:
Thanks @jonboulle 

Updated version:
- [x] use 128 + signal as exit code and verify in tests
- [x] make error messages more generic for check on err from os.Stat()
- [x] Don't sleep in tests! Continuously poll for existence of pid file (max 1 second) on start. Use proc.Wait() on quit.

plugins/ipam/dhcp/daemon_test.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+	var procAttr os.ProcAttr
+	proc, err := os.StartProcess(path.Join(cwd, "../../../bin/dhcp"), []string{"dhcp", "daemon"}, &procAttr)
+	if err != nil {
+		t.Errorf("Unable to start process: %v", err)
+		return
+	}
+
+	// verify existence of dhcp socket, keep retrying until 1 second has passed
+	for startTime := time.Now(); err != nil || time.Now().Before(startTime.Add(1*time.Second)); _, err = os.Stat(socketPath) {
```
steveeJ: This loop header looks very confusing. We don't require most compact version of the code, but prefer readability. Please rewrite this accordingly.

plugins/ipam/dhcp/daemon_test.go
```diff
@@ -0,0 +1,95 @@
...
+		// NOOP
+	}
+	if err != nil {
```
steveeJ: it should be possible to not have to check `err` again after the loop

plugins/ipam/dhcp/daemon_test.go
```diff
@@ -0,0 +1,95 @@
...
+	status, ok := procState.Sys().(syscall.WaitStatus)
+	if !ok {
+		t.Errorf("Unable to convert procState.Sys to syscall.WaitStatus. Unsupported platform?")
+	} else {
+		var signum int
+		switch signal := os.Interrupt.(type) {
+		case syscall.Signal:
+			signum = int(signal)
+		default:
+			signum = 0
```
steveeJ: use `signum := 0` instead of the default case?

fnordahl:
Thanks @steveeJ 

Updated version:
- [x] Initialize signum before select making the default case redundant
- [x] Re-write retryloop to be more readable, and utilize Go idioms

squeed:
The CNI repository has split; if you'd like to re-open this PR against the plugins repository, please do so.

PR #217, opened 2016-05-20
build: don't be verbose / test: fix bash variable escaping in sudo command 

steveeJ:
/cc @zachgersh


steveeJ:
Trivial changes, merging :-)

Issue #125, opened 2016-02-17
README: various improvements


jonboulle:
- [x] describe current state of project and roadmap/timeline 
- [x] link to implementations/integrations (e.g. rkt, kubernetes, calico, weave)
- [x] encourage people to get involved and describe how
- [x] Roadmap https://github.com/coreos/rkt/blob/master/ROADMAP.md #218

squaremo:
Also worth considering:

 * contribution guidelines (commit messages e.g.)
 * how planning/roadmaps works

jonboulle:
FWIW, there is a contribution guide [here](https://github.com/appc/cni/blob/master/CONTRIBUTING.md). That document is just a fork of our standard CoreOS one, feel free to tweak/simplify/fold into README as desired.

steveeJ:
@squaremo any updates on this one?

squaremo:
> any updates on this one?

Working on it in my spare time :)

jonboulle:
I suggest punting so we can release a 0.2.0 asap.

squaremo:
OK, let me put some uncontroversial changes into a PR quickly ..

steveeJ:
@jonboulle 

> encourage people to get involved and describe how

Is https://github.com/appc/cni#contributing-to-cni enough to check that box?

steveeJ:
Last bits fixed by #218.

PR #218, opened 2016-05-20
Add ROADMAP document

steveeJ:
Contributes to #125.

/cc @squaremo

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,33 @@
+# CNI Roadmap
+
+This document defines a high level roadmap for CNI development.
+The list below is not complete, and we advise to get the current project state from the [milestones defined in GitHub](https://github.com/containernetworking/cni/milestones).
+
+## CNI Milestones
+
+### [v0.2.0](https://github.com/containernetworking/cni/milestones/v0.2.0)
+
+* Signed release binaries
+* Introduction of a testing strategy/framework
+
+### [v0.3.0](https://github.com/containernetworking/cni/milestones/v0.3.0)
+
+* Further increase test coverage
+* Simpler default route handling in bridge plugin
+* Clarify project description, documentation and contribution guidelines
+
+### [v0.4.0](https://github.com/containernetworking/cni/milestones/v0.4.0)
+
+* Further increase test coverage
+* Simpler bridging of host interface
+* Improve IPAM allocator predictability
+* Allow in- and output of arbitrary K/V pairs for plugins
+
+### [v1.0.0](https://github.com/containernetworking/cni/milestones/v1.0.0)
+
+- Plugin composition functionality
+- IPv6 support
+- Stable SPEC
+- Strategy and tooling for backwards compatibility
+- Complete test coverage
+- Integrate build artefact generation with CI


ROADMAP.md
```diff
@@ -0,0 +1,19 @@
+# CNI Roadmap
+
+This document defines a high level roadmap for CNI development.
+The list below is not complete, and we advise to get the current project state from the [milestones defined in GitHub](https://github.com/containernetworking/cni/milestones).
+
+## CNI Milestones
+
+### [v0.3.0](https://github.com/containernetworking/cni/milestones/v0.3.0)
```
jonboulle: peanut gallery, this seems a bit uselessly vague. how about an estimated timeline and/or at least some high level goals for the milestone releases?

steveeJ: Yep, would be good to have some real content ;-) I was wondering if we shall do a mix of roadmap and changelog, opinions?

jonboulle: I think there's enough in the milestones to extract a few highlights to put here

ROADMAP.md
```diff
@@ -0,0 +1,38 @@
...
+* Add firewalld plugin
+
+### [v1.0.0](https://github.com/containernetworking/cni/milestones/v1.0.0)
+
+- Plugin composition functionality
+- IPv6 support
+- Stable SPEC
+- Strategy and tooling for backwards compatibility
+- Complete test coverage
+- Integrate build artefact generation with CI
```
fnordahl: s/artefact/artifact

jonboulle: http://grammarist.com/spelling/artefact-artifact/

jonboulle:
lgtm for now

ROADMAP.md
fnordahl: Ah, all good then. We learn American English here in Norway. Sorry about the noise. O:-)

tomdee:
I'm happy for this to be merged but can we agree that this roadmap hasn't had much discussion (that I'm aware of) so we should be afraid to change it future? I worry that once it's committed we're somehow committed to it.

jonboulle:
I for one certainly wouldn't consider it set in stone

steveeJ:
> I'm happy for this to be merged but can we agree that this roadmap hasn't had much discussion (that I'm aware of) so we should be afraid to change it future? I worry that once it's committed we're somehow committed to it.

It hasn't had much discussion. I assembled it from different conversations, including the recent meetings but also OOB discussions, mostly @squaremo. It's a first impression and I'm certainly not saying this is how it has to be. As always, it's a start.

ROADMAP.md
```diff
@@ -0,0 +1,38 @@
...
+## CNI Milestones
+
+### [v0.2.0](https://github.com/containernetworking/cni/milestones/v0.2.0)
+
+* Signed release binaries
+* Introduction of a testing strategy/framework
+
+### [v0.3.0](https://github.com/containernetworking/cni/milestones/v0.3.0)
+
+* Further Increase test coverage
```
squaremo: Nitpick: I don't think "Increase" needs a capital letter here.

squaremo:
I sympathise with Tom's concern -- an explicit roadmap represents a commitment (whatever one says in the front matter). However we do need to balance that with taking _some_ responsibility for making progress.

Perhaps we can compile the roadmap this way:

 * start with what happened in previous releases
 * include proposed (and ideally, discussed) headline issues for one minor version ahead
 * end with aspirations for the next major release

These are all present already -- nice! The difference is that 0.4 and 0.5 would be elided (but may be discoverable from github milestones, which are understood to be provisional). WDYT @steveeJ @tomdee ?

steveeJ:
> The difference is that 0.4 and 0.5 would be elided

I have been trying to keep issues/PRs to an undefined but limited count to allow a more frequent release cycle, represented in the ROADMAP as well as the issue management. 
With your suggestion, would we put these planned changes to the 1.0 milestone? Once released, then create the previous milestones and move those items back?

ROADMAP.md
```diff
@@ -0,0 +1,33 @@
...
+
+### [v0.3.0](https://github.com/containernetworking/cni/milestones/v0.3.0)
+
+* Further increase test coverage
+* Simpler default route handling in bridge plugin
+* Clarify project description, documentation and contribution guidelines
+
+### [v0.4.0](https://github.com/containernetworking/cni/milestones/v0.4.0)
+
+* Further Increase test coverage
```
squaremo: No capital needed for "increase"

squaremo:
Thanks @steveeJ and commenters!

PR #222, opened 2016-05-24
pkg/ns: verify netns when initialized with GetNS

### LINKED ISSUE #221: pkg/ns: doesn't ensure that the stored fd is a valid network namespace
My concern is if we initialize a `ns.NetNS` object using `ns.GetNS(path)` and _path_ points to just an empty file without any network namespace mounted on top, we don't detect that. 
We'll eventually throw an error, but that is simply from the attempt to execute `ns.Do()` and `ns.Set()` which will try to switch the network namespace, which will fail.

I suggest that we introduce validation of the file and set the `mounted` flag respectively. Furthermore, if `mounted == false`, we shouldn't attempt to switch the network namespace at all.

steveeJ:
Fixes #221.

### DIFF
--- a/pkg/ns/ns.go
+++ b/pkg/ns/ns.go
@@ -21,6 +21,7 @@ import (
 	"path"
 	"runtime"
 	"sync"
+	"syscall"
 
 	"golang.org/x/sys/unix"
 )
@@ -77,9 +78,32 @@ func GetCurrentNS() (NetNS, error) {
 func GetNS(nspath string) (NetNS, error) {
 	fd, err := os.Open(nspath)
 	if err != nil {
+		return nil, fmt.Errorf("Failed to open %v: %v", nspath, err)
+	}
+
+	isNSFS, err := IsNSFS(nspath)
+	if err != nil {
+		fd.Close()
 		return nil, err
 	}
+	if !isNSFS {
+		fd.Close()
+		return nil, fmt.Errorf("%v is not of type NSFS", nspath)
+	}
+
 	return &netNS{file: fd}, nil
+}
+
+// Returns whether or not the nspath argument points to a network namespace
+func IsNSFS(nspath string) (bool, error) {
+	const NSFS_MAGIC = 0x6e736673
+
+	stat := syscall.Statfs_t{}
+	if err := syscall.Statfs(nspath, &stat); err != nil {
+		return false, fmt.Errorf("failed to Statfs %q: %v", nspath, err)
+	}
+
+	return stat.Type == NSFS_MAGIC, nil
 }
 
 // Creates a new persistent network namespace and returns an object
@@ -255,7 +279,7 @@ func (ns *netNS) Set() error {
 func WithNetNSPath(nspath string, toRun func(NetNS) error) error {
 	ns, err := GetNS(nspath)
 	if err != nil {
-		return fmt.Errorf("Failed to open %v: %v", nspath, err)
+		return err
 	}
 	defer ns.Close()
 	return ns.Do(toRun)
--- a/pkg/ns/ns_test.go
+++ b/pkg/ns/ns_test.go
@@ -211,6 +211,28 @@ var _ = Describe("Linux namespace operations", func() {
 			})
 		})
 	})
+
+	Describe("IsNSFS", func() {
+		It("should detect a namespace", func() {
+			createdNetNS, err := ns.NewNS()
+			isNSFS, err := ns.IsNSFS(createdNetNS.Path())
+			Expect(err).NotTo(HaveOccurred())
+			Expect(isNSFS).To(Equal(true))
+		})
+
+		It("should refuse other paths", func() {
+			tempFile, err := ioutil.TempFile("", "nstest")
+			Expect(err).NotTo(HaveOccurred())
+			defer tempFile.Close()
+
+			nspath := tempFile.Name()
+			defer os.Remove(nspath)
+
+			isNSFS, err := ns.IsNSFS(nspath)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(isNSFS).To(Equal(false))
+		})
+	})
 })
 
 func allNetNSInCurrentProcess() []string {
--- a/pkg/ns/ns_test.go
+++ b/pkg/ns/ns_test.go
@@ -17,6 +17,7 @@ package ns_test
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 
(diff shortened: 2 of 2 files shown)


pkg/ns/ns.go
```diff
@@ -74,11 +75,22 @@ func GetCurrentNS() (NetNS, error) {
...
+
+	stat := syscall.Statfs_t{}
+	if err = syscall.Statfs(nspath, &stat); err != nil {
+		return nil, fmt.Errorf("Failed to Statfs %v: %v", nspath, err)
 	}
-	return &netNS{file: fd}, nil
+
+	return &netNS{
+		file:    fd,
+		mounted: stat.Type == NSFS_MAGIC,
```
dcbw: I wouldn't set mounted:true for plain GetNS() calls.  The  reason is because only the creator of the netns should be disposing of it via ns.Close(), and things that call GetNS() are the creator.  I think all you want to do here is return an error if the mountpoint is not an NSFS mount point.

pkg/ns/ns.go
```diff
@@ -179,7 +191,28 @@ func (ns *netNS) Close() error {
...
 }
 
+func (ns *netNS) validate(action string) error {
+	switch {
+	case ns.file == nil:
+		return fmt.Errorf(
+			"%q failed: no filedescriptor associated",
+			action,
+		)
+	case ns.mounted == false:
```
dcbw: We don't care if some paths aren't mounted; we only care that ones created through NewNS() are mounted.  For example, a CNI plugin that is passed a namespace path needs to close the file handle with Close(), and if mounted:true then pkg/ns will also try to unmount it and remove the mountpoint when Close() is called.  We only want that to happen for things that pkg/ns creates with NewNS(), not for any random namespace passed into the process.

dcbw: Possibly we should rename 'mounted' to 'cni_created' or something like that?

pkg/ns/ns.go
```diff
@@ -179,7 +191,28 @@ func (ns *netNS) Close() error {
...
+			"%q failed: no network namespace mounted on file %q",
+			action, ns.file.Name,
+		)
+	}
+
+	return nil
+}
+
 func (ns *netNS) Do(toRun func(NetNS) error) error {
+	if err := ns.validate("Do"); err != nil {
```
dcbw: Why bother calling validate() here?  Shouldn't ensuring validity on GetNS() be sufficient?  If we hold an open file handle to the namespace, it can't have been closed yet and it can't have been unmounted, so it should be sufficient to check that it's NSFS_MAGIC on GetNS() and then we're good, I'm pretty sure.

pkg/ns/ns.go
```diff
@@ -218,6 +251,10 @@ func (ns *netNS) Do(toRun func(NetNS) error) error {
 }
 
 func (ns *netNS) Set() error {
+	if err := ns.validate("Set"); err != nil {
```
dcbw: Same comment as Do()...

pkg/ns/ns.go
```diff
@@ -74,11 +75,22 @@ func GetCurrentNS() (NetNS, error) {
...
+		return nil, fmt.Errorf("Failed to open %v: %v", nspath, err)
+	}
+
+	stat := syscall.Statfs_t{}
+	if err = syscall.Statfs(nspath, &stat); err != nil {
+		return nil, fmt.Errorf("Failed to Statfs %v: %v", nspath, err)
 	}
-	return &netNS{file: fd}, nil
+	if stat.Type != NSFS_MAGIC {
+		return nil, fmt.Errorf("no network namespace mounted on file %q", nspath)
```
dcbw: Have to close the fd here on error.

pkg/ns/ns.go
```diff
@@ -74,11 +75,22 @@ func GetCurrentNS() (NetNS, error) {
...
+	const NSFS_MAGIC = 0x6e736673
+
 	fd, err := os.Open(nspath)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Failed to open %v: %v", nspath, err)
+	}
+
+	stat := syscall.Statfs_t{}
+	if err = syscall.Statfs(nspath, &stat); err != nil {
```
dcbw: have to close the fd here on error.

pkg/ns/ns.go
```diff
@@ -74,11 +75,22 @@ func GetCurrentNS() (NetNS, error) {
...
+		return nil, fmt.Errorf("Failed to Statfs %v: %v", nspath, err)
 	}
-	return &netNS{file: fd}, nil
+	if stat.Type != NSFS_MAGIC {
+		return nil, fmt.Errorf("no network namespace mounted on file %q", nspath)
+	}
+
+	return &netNS{file: fd, mounted: true}, nil
```
dcbw: The Travis tests are failing because of this mounted:true.  Per my other comments, I don't think we should set mounted:true for namespaces that pkg/ns doesn't itself create, since that will attempt to unmount the namespace which is going to fail if other holders have it open.  Only the creator of the namespace should try to unmount it on Close().

dcbw:
My take on this would be https://github.com/dcbw/cni/commits/namespace-validate ...

pkg/ns/ns.go
```diff
@@ -76,11 +77,31 @@ func GetCurrentNS() (NetNS, error) {
 func GetNS(nspath string) (NetNS, error) {
 	fd, err := os.Open(nspath)
 	if err != nil {
+		return nil, fmt.Errorf("Failed to open %v: %v", nspath, err)
+	}
+
+	isNSFS, err := IsNSFS(nspath)
+	if err != nil {
```
dcbw: Still need to close the FD here on error return.

pkg/ns/ns.go
```diff
@@ -76,11 +77,31 @@ func GetCurrentNS() (NetNS, error) {
...
 		return nil, err
 	}
+	if !isNSFS {
```
dcbw: Still need to close the FD here on error return.

pkg/ns/ns.go
steveeJ: Thanks!

dcbw:
LGTM now, but how about some tests like:

```golang
		It("fails when the path is not a namespace", func() {
			tempFile, err := ioutil.TempFile("", "nstest")
			defer tempFile.Close()
			defer os.Remove(tempFile.Name())

			_, err = ns.GetNS(tempFile.Name())
			Expect(err).To(MatchError(fmt.Sprintf("no network namespace mounted on file %q", tempFile.Name())))
		})
```

steveeJ:
> LGTM now, but how about some tests like:

Right, always good to have more tests! I fleshed your draft out and added a test case.

dcbw:
> Right, always good to have more tests! I fleshed your draft out and added a test case.

Thanks, test LGTM.

Issue #116, opened 2016-02-08
IPAM allocator and backend should be decoupled


steveeJ:
We could reorganize the current implementation which could be done right now without losing compatibility with existing configurations.
- ipam
  - host-local.go (implements combination of disk and sequential)
  - store
    - store.go (package, interface)
    - disk/...
  - allocator
    - allocator.go (package, interface)
    - sequential/...

However, the next step could be to configure the store and allocator in the configuration file, so that any combination could be chosen. We can probably still allow predefined combinations for known types as _host-local_.

/cc @aanm

aanm:
Answering to the #115 question, I'm currently using the `IPAMConfig struct` and the allocator on a different IPAM.

jonboulle:
Splitting this apart SGTM

PR #224, opened 2016-05-24
Add consul backend for host-local plugin

murat1985:
Hi guys,

I thought it would be amazing if host-local ipam plugin would support a distributed backend, such as KV store. 
I propose to use Consul as another alternative for disk store with host-local plugin. 
I've implemented storing ip-allocations and network definitions in Consul, as well as locking with Consul session/lock.

To work with this backend a configuration file should contain something like this:

```
{
    "name": "mynet",
    ...
    "ipam": {
    "type": "host-local",
    "backend": "consul",
        "subnet": "10.22.0.0/16",
    "consul_addr": "127.0.0.1",
    "consul_port": "8500",
    "dc": "dc1",
        "routes": [
            { "dst": "0.0.0.0/0" }
        ]
    }
}
```

That would be really great to have such backend to experiment with CNI in distributed environments.


philips:
This should be a separate plugin as the point of host-local is to be local to a host.

It feels like this should be a different external plugin as this starts to be a big distributed system and to test correctly will require complex functional tests.

jonboulle:
+1 on keeping host-local as host-local, and that consul support shouldn't be in this repo; perhaps CNI can add a store backend interface that others can more easily plug in to etcd, consul, etc.

philips:
@jonboulle Yea, I wouldn't be opposed to putting a Go interface in here for storage to make creating these external plugins possible.

@murat1985 would you be interested in doing that work?

murat1985:
@philips yeah it makes a lot of sense and it is quiet interesting as well, will start to work on in it then. thanks

steveeJ:
Thanks for your proposal @murat1985, in general this kind of change is very welcome.

I wrote down my thoughts about decoupling the IPAM allocator and storage backend in #116. 
After we figure that out you it is possible to implement this as a storage backend and pair it with other allocators. Let me know if this sounds reasonable to you and if you're interested in helping out!

murat1985:
@steveeJ sounds reasonable and I am ready to help. Having some universal mechanism to attach different backends would be great. I suspect that besides allocator and backend, also configuration should be decoupled as well if we are going to pass some structures inside a backend. I will look into #116 carefully during this weekend.

murat1985:
@steveeJ I tried to decouple backend and allocator, not sure that it is right way. I had to move `config` into allocator.

Also to be more generic if some of the backends want to use arguments which are passed into plugin, I suggest to send full `ipamConf` struct to the backend, e.g.:
```
store, err := disk.New(ipamConf)
```

For sequential plugin I suggest to add additional Args options: `StoreAddr`, `StorePort` and `StoreNS`.

Which could be used as follows:
```
CNI_ARGS="StoreAddr=127.0.0.1;StorePort=8500;StoreNs=dc1;IgnoreUnknown=1"
```
I am happy to write tests for host-local plugin to fix coverage.

squeed:
Closing this; @murat1985, feel free to submit against the plugins repostory (though I suspect it should be its own plugin).

PR #230, opened 2016-05-27
plugins: don't require CNI_NETNS for DEL command

### LINKED ISSUE #220: plugins: teardown doesn't handle empty netns
When a plugin is called with a non-existing network namespace, it should still be able to cleanup as much as possible despite throwing an error.

steveeJ:
This will allow to free up the IPAM allocations when the caller doesn't
have access to the network namespace anymore, e.g. due to a reboot.

Fixes #220.

### DIFF
--- a/pkg/skel/skel.go
+++ b/pkg/skel/skel.go
@@ -36,28 +36,72 @@ type CmdArgs struct {
 	StdinData   []byte
 }
 
+type reqForCmdEntry map[string]bool
+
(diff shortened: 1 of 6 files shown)


jonboulle:
nice, LGTM

iaguis:
LGTM

steveeJ:
@jonboulle @iaguis I added a test case

PR #231, opened 2016-05-27
pkg/ns: introduce error types indicate NS verification

steveeJ:
Depends on #230.

### DIFF
--- a/pkg/ns/ns.go
+++ b/pkg/ns/ns.go
@@ -81,11 +81,23 @@ const (
 	PROCFS_MAGIC = 0x9fa0
 )
 
-func IsNS(nspath string) (isNS bool, msg string, err error) {
+type NSPathNotExistErr struct{ msg string }
+
+func (e NSPathNotExistErr) Error() string { return e.msg }
+
+type NSPathNotNSErr struct{ msg string }
+
+func (e NSPathNotNSErr) Error() string { return e.msg }
+
+func IsNSorErr(nspath string) error {
 	stat := syscall.Statfs_t{}
-	if err = syscall.Statfs(nspath, &stat); err != nil {
-		err = fmt.Errorf("failed to Statfs %s: %v", nspath, err)
-		return
+	if err := syscall.Statfs(nspath, &stat); err != nil {
+		if os.IsNotExist(err) {
+			err = NSPathNotExistErr{msg: fmt.Sprintf("failed to Statfs %q: %v", nspath, err)}
+		} else {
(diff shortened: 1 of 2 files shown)


pkg/ns/ns.go
```diff
@@ -81,11 +81,23 @@ const (
...
+func (e NSPathNotNSErr) Error() string { return e.msg }
+
+func IsNSorErr(nspath string) error {
 	stat := syscall.Statfs_t{}
-	if err = syscall.Statfs(nspath, &stat); err != nil {
-		err = fmt.Errorf("failed to Statfs %s: %v", nspath, err)
-		return
+	if err := syscall.Statfs(nspath, &stat); err != nil {
+		if os.IsNotExist(err) {
+			err = NSPathNotExistErr{msg: fmt.Sprintf("failed to Statfs %s: %v", nspath, err)}
```
iaguis: `%q` on nspath?

pkg/ns/ns.go
```diff
@@ -81,11 +81,23 @@ const (
...
+		} else {
+			err = fmt.Errorf("failed to Statfs %s: %v", nspath, err)
```
iaguis: ditto

pkg/ns/ns.go
```diff
@@ -95,33 +107,29 @@ func IsNS(nspath string) (isNS bool, msg string, err error) {
...
 		return nil, err
 	}
-	if !isNS {
-		return nil, fmt.Errorf("no network namespace detected on %s: %s", nspath, msg)
-	}
 
 	fd, err := os.Open(nspath)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to open %v: %v", nspath, err)
+		return nil, err
```
dcbw: for consistency, should this check os.IsNotExist() and return NSPathNotExistErr?  The path already got verified by IsNSorErr() but there's a very small window between that and the Open() where if could get deleted.

steveeJ: Alternatively I could move the call to `Open()` to the beginning of the function. Little chance it gets created and unmounted in the meantime ;-)

steveeJ: On second thought, this is not really something I'd like to consider. There is also a small window between the call to `IsNSorErr()` and the actual return of that error value, but yet it doesn't make sense to check it again. We're talking about very short periods of time, when the error we're probably trying to catch is in time length of a reboot or similar.

iaguis:
LGTM

jonboulle:
lgtm

Issue #31, opened 2015-07-09
Add IPv6 support


eyakubovich:
The spec does allow for IPv6 but the plugins only support IPv4 (with the exception of host-local).

Tracking:
- [ ] bridge
- [ ] ptp
- [ ] macvlan
- [ ] ipvlan
- [ ] host-local #131

NeilW:
I'm particularly interested in adding IPv6 support to macvlan since plugging the containers straight into the upstream ethernet network allows stateless address auto configuration to work - where a Host virtual machine is allocated a /64 on its network.

bboreham:
* `host-local` in https://github.com/containernetworking/plugins/pull/12
* `bridge` was done in https://github.com/containernetworking/plugins/pull/10
* `ptp` in https://github.com/containernetworking/plugins/pull/25
* `macvlan` and `ipvlan` picked the support up as part of the above

PR #233, opened 2016-05-30
ipam/host-local: Return correct IP version for allocation

fnordahl:
The logic in the host-local IPAM plugin does support IPv6 allocation.
However, the result is given with types.IP4 regardless of IP address
type. This patch addresses this by checking for what IP version is
used before filling the result structure.

Fixes part of #31

### DIFF
--- a/plugins/ipam/host-local/main.go
+++ b/plugins/ipam/host-local/main.go
@@ -47,8 +47,11 @@ func cmdAdd(args *skel.CmdArgs) error {
 		return err
 	}
 
-	r := &types.Result{
-		IP4: ipConf,
+	var r types.Result
+	if ipConf.IP.IP.To4() == nil {
+		r.IP6 = ipConf
+	} else {
+		r.IP4 = ipConf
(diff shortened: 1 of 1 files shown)


fnordahl:
FYI; There will be tests coming along soon.

dcbw:
lgtm, assuming the tests land

steveeJ:
@fnordahl any time for adding tests here?

dcbw:
Per https://github.com/containernetworking/cni/issues/245 I suggested dropping this PR since it doesn't support dual-stack config and #137 does...  Need consensus on my suggestions in #245 first though.

fnordahl:
I have no problems with going with #137 over this one. I probably should have done my research on on-going PRs before proposing this anyway. It does require a update of the specification, but that is something we want anyway. :+1:

rosenhouse:
Plugins have moved to a [separate repo](https://github.com/containernetworking/plugins).  Also, I think this is already fixed.

PR #235, opened 2016-05-30
[WIP] plugins/main/ptp: Add IPv6 support

fnordahl:
Fixes part of #31

### DIFF
--- a/plugins/main/ptp/ptp.go
+++ b/plugins/main/ptp/ptp.go
@@ -73,12 +73,21 @@ func setupContainerVeth(netns, ifName string, mtu int, pr *types.Result) (string
 			return fmt.Errorf("failed to look up %q: %v", ifName, err)
 		}
 
+		ipConfig := pr.IP6
+		cidrOnes := 128
+		cidrBits := 128
+		if ipConfig == nil {
+			ipConfig = pr.IP4
+			cidrOnes = 32
+			cidrBits = 32
+		}
+
 		// Delete the route that was automatically added
 		route := netlink.Route{
 			LinkIndex: contVeth.Attrs().Index,
 			Dst: &net.IPNet{
-				IP:   pr.IP4.IP.IP.Mask(pr.IP4.IP.Mask),
-				Mask: pr.IP4.IP.Mask,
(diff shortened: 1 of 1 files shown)


dcbw:
It seems like Go should have a better mechanism to create a bitmask, but lgtm.

plugins/main/ptp/ptp.go
```diff
@@ -120,26 +127,33 @@ func setupContainerVeth(netns, ifName string, mtu int, pr *types.Result) (string
...
+func setupHostVeth(vethName string, ipamResult *types.Result) error {
 	// hostVeth moved namespaces and may have a new ifindex
 	veth, err := netlink.LinkByName(vethName)
 	if err != nil {
 		return fmt.Errorf("failed to lookup %q: %v", vethName, err)
 	}
 
-	// TODO(eyakubovich): IPv6
+	ipConfig := ipamResult.IP6
+	cidrOnes := 128
```
steveeJ: please declare these magic numbers as const, like so:
```
const (
    CIDROnesIPv6 = 128
    CIDROnesIPv4 = 32
)
```
Of course make sure I understood the names correctly ;-)

plugins/main/ptp/ptp.go
```diff
@@ -120,26 +127,33 @@ func setupContainerVeth(netns, ifName string, mtu int, pr *types.Result) (string
...
+	cidrBits := 128
+	if ipConfig == nil {
```
steveeJ: This behavior will prefer IPv6 over IPv4. Ideally we'd support both simultaneously, but we have discussed in other issues that it probably requires a spec revision. But can be done in this PR already is document the new behavior and then follow-up with a spec change.

dcbw: Maybe even a helper function like:

```golang
func getIpConfig(result types.Result) (*types.IPConfig, uint, uint) {
    <stuff>
}
```

which might be a bit clearer in the code.

rosenhouse:
Plugins have moved to [their own repository](https://github.com/containernetworking/plugins).  If you'd like to continue development, open a new PR over there.

PR #238, opened 2016-06-01
pkg/types: Add UnmarshallableString type

tomdee:
Allow strings to be unmarshalled for CNI_ARGS

CNI_ARGS uses types.LoadArgs to populate a struct.
The fields in the struct must meet the TextUnmarshaler interface.

This code adds a UnmarshallableString type to assist with this.

PR #240, opened 2016-06-06
pkg/types: unmarshal string tests

steveeJ:
Follows up #238.

/cc @tomdee


tomdee:
Tests LGTM - thanks for writing them!

PR #241, opened 2016-06-08
Add arm64 and ppc64le as new platforms

luxas:
Made it possible to build for other arches using `GOARCH=something ./build`
Made the CI spot compilation errors
Updated the `golang.org/x/sys/unix/` dependency
travis.yml now has the same layout as etcd: https://github.com/coreos/etcd/pull/5450

@steveeJ


dcbw:
I'm probably ignorant of the travis operation here, but is the point of the test -> build change for !amd64 because CNI doesn't currently pass tests on !amd64?  Why wouldn't we want to run the tests for arm/ppc?

luxas:
It isn't possible to run arm, arm64 or ppc64le go binaries on an amd64 host, unfortunately, so we can't run the tests unless we're on a native machine

But catching compilation errors is a good start

steveeJ:
Please split out the travis change from the vendoring bump. We'll deal with the latter first to see if that causes the tests to fail. Thanks!

luxas:
Tests passing, the only problem was that I replaced GOBIN because it can't be used when cross-compiling, but the test script used the variable. I fixed that now.

Also, I added the arches to your releasing script. I haven't tested it, so please give feedback on it.
It would be really awesome if you may upload the cross-compiled binaries at the same time you're uploading the "normal" amd64 ones. Everything can be done from an amd64 host, so it shouldn't be any problems with it.

We're gonna use cni now in docker-multinode, and it would help if binaries for all arches are easily available.

dcbw:
Seems OK to me, but somebody who does actual releases should take a look too.

luxas:
@steveeJ Please take a look. If you want me to split the releasing code out from this patch, I'll do that and send a follow-up.
For us, it would be great to have released binaries for all arches already in v0.4.0, and that shouldn't require more code than this.

luxas:
Should I split it up now, so you may merge the code that makes it possible to compile for arm64 and ppc64le, and then we'll deal with the releasing later on?

steveeJ:
@luxas No need to split out the release script change IMO. I'll try the release script and let you know if it works out. 

On another note, have you tried running the cross-compiled binaries using qemu-user? I'm not sure if the netlink syscalls are supported but it'd be very useful for CI.

steveeJ:
Please adjust the commit message(s) to conform to https://github.com/containernetworking/cni/blob/master/CONTRIBUTING.md#format-of-the-commit-message, thanks a lot!

bboreham:
The change from `go install` to `go build` means that dependent packages are no longer installed by the build; they are built in a temporary directory and then thrown away.

If you didn't mean to have that effect, `go build -i` is a closer match.

luxas:
@bboreham Yes, agree. That was unintentional, but by utilizing `go build`, it's possible to specify output path, that's why I changed it.

But is it possible to merge this now (@steveeJ)? 
I can make a small follow up for that after this PR is merged (adding `-i` is not critical)

luxas:
ping @steveeJ

luxas:
Are there any blockers for this PR or haven't you just had time for it...?
This would be useful for Kubernetes.

Side question: When is v0.4.0 going to be released?

cc @dcbw @rosenhouse

tomdee:
I'd be happy to merge this - just needs a rebase

luxas:
@tomdee Rebased. PTAL and merge.
When will v0.4.0 be out btw?

PR #247, opened 2016-06-13
SPEC: introduce "args" field and new error code

tomdee:
Based on previous discussions on the CNI maintainers calls, the spec is
unclear on 1) when CNI_ARGS should be used and 2) the fact the dynamic
config can be passed in through the network JSON.

This PR makes it clear that per-container config can be passed in
through the network JSON, adding a top level `args` field into
which orchestrators can add additional metadata without worrying that
plugins might reject the additional data. It also allows for plugins to
reject unknown fields passed in at the top level.

Using JSON is preferable to CNI_ARGS since it allows namespaced and
structured data. CNI_ARGS is a flat list of KV pairs which has reserved
characters with no escaping rules defined.
CNI_ARGS may still be used by orchestrators that want the simplicity of
passing the network config JSON as specified by the user, unchanged
through to the CNI plugin. But for any kind of structured data, it's
recommended that the `args` field in the JSON is used instead.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -23,6 +23,7 @@ The intention is for the container runtime to first create a new network namespa
 It then determines which networks this container should belong to and for each network, which plugin must be executed.
 The network configuration is in JSON format and can easily be stored in a file.
 The network configuration includes mandatory fields such as "name" and "type" as well as plugin (type) specific ones.
+The network configuration allows for fields to change values between invocations. For this purpose there is an optional field "args" which should contain the varying information.
 The container runtime sequentially sets up the networks by executing the corresponding plugin for each network.
 Upon completion of the container lifecycle, the runtime executes the plugins in reverse order (relative to the order in which they were added) to disconnect them from the networks.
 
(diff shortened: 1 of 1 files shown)


rosenhouse:
Good idea, 👍

SPEC.md
```diff
@@ -255,3 +264,4 @@ IPAM plugin examples:
 
 ## Well-known Error Codes
 - `1` - Incompatible CNI version
+- `2` - Unknown network configuration passed to plugin.
```
philips: It would be good to qualify "unknown" a bit more. An example might be helpful somewhere.

steveeJ: Better:
Unsupported field in network configuration.

SPEC.md
```diff
@@ -71,7 +72,7 @@ It will then look for this executable in a list of predefined directories. Once
 - `CNI_ARGS`: Extra arguments passed in by the user at invocation time. Alphanumeric key-value pairs separated by semicolons; for example, "FOO=BAR;ABC=123"
 - `CNI_PATH`: Colon-separated list of paths to search for CNI plugin executables
 
-Network configuration in JSON format is streamed through stdin.
+Network configuration in JSON format is streamed through stdin. This means it is not tied to a particular file on disk and can contain information which changes between invocations.
```
bboreham: "streamed to the plugin through stdin" ?  Just to make it a bit clearer what we are talking about.

SPEC.md
```diff
@@ -128,6 +129,7 @@ The network configuration is described in JSON form. The configuration can be st
 - `cniVersion` (string): [Semantic Version 2.0](http://semver.org) of CNI specification to which this configuration conforms.
 - `name` (string): Network name. This should be unique across all containers on the host (or other administrative domain).
 - `type` (string): Refers to the filename of the CNI plugin executable.
+- `args` (dictionary): Optional additional arguments provided by the container runtime. For example a dictionary of labels could be passed to CNI plugins by adding them to a labels field under `args`.
```
bboreham: Is there an expectation that `args` will have any correspondence with `CNI_ARGS` in the environment?

Do we need to say which takes precedence if there are matching keys with conflicting values?

steveeJ: I would say that this is completely up to the plugin developers and should go in the respective documentation. I don't have an opinion if we should suggest or even dictate a precedence order, FWIW the spec already says on a different line `args` is to be preferred over `CNI_ARGS`

PR #254, opened 2016-06-21
pkg/ipam: IP based macaddr

steveeJ:
Closes #196.

### DIFF
--- a/pkg/ip/link_test.go
+++ b/pkg/ip/link_test.go
@@ -1,0 +1,132 @@
+// Copyright 2016 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ip_test
+
+import (
+	"fmt"
+	"net"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/containernetworking/cni/pkg/ip"
+	"github.com/containernetworking/cni/pkg/ns"
+
+	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netlink/nl"
+)
+
+var _ = Describe("Link", func() {
+	const (
+		ifaceFormatString string = "i%d"
+		mtu               int    = 1400
+	)
+	var (
+		hostNetNS         ns.NetNS
+		containerNetNS    ns.NetNS
+		ifaceCounter      int = 0
+		hostVethName      string
+		containerVethName string
+	)
+
+	BeforeEach(func() {
+		var err error
+
+		hostNetNS, err = ns.NewNS()
+		Expect(err).NotTo(HaveOccurred())
+
+		containerNetNS, err = ns.NewNS()
+		Expect(err).NotTo(HaveOccurred())
+
+		_ = containerNetNS.Do(func(ns.NetNS) error {
+			defer GinkgoRecover()
+
+			hostVeth, containerVeth, err := ip.SetupVeth(fmt.Sprintf(ifaceFormatString, ifaceCounter), mtu, hostNetNS)
+			if err != nil {
+				return err
+			}
+			Expect(err).NotTo(HaveOccurred())
+
+			hostVethName = hostVeth.Attrs().Name
+			containerVethName = containerVeth.Attrs().Name
+
+			return nil
+		})
+	})
+
+	AfterEach(func() {
+		Expect(containerNetNS.Close()).To(Succeed())
+		Expect(hostNetNS.Close()).To(Succeed())
+		ifaceCounter++
+	})
+
+	It("SetupVeth must put the veth endpoints into the separate namespaces", func() {
+		_ = containerNetNS.Do(func(ns.NetNS) error {
+			defer GinkgoRecover()
+
+			_, err := netlink.LinkByName(containerVethName)
+			Expect(err).NotTo(HaveOccurred())
+
+			return nil
+		})
+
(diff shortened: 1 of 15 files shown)


dcbw:
@steveeJ I'm a bit nervous about hardcoding this behavior in ConfigureInterface().  A plugin might create its own veth pair and then use ConfigureInterface() to do the actual IPAM setup, but may want to preserve the MAC it already set when the veth was created.  Is there a way to make this behavior optional, or only performed if the 'ptp' plugin is used?  Maybe it could be a separate function to set the hwaddr that the standard plugins call instead of being in ConfigureInteface().

steveeJ:
> I'm a bit nervous about hardcoding this behavior in ConfigureInterface()

I see your concern, and I think it's worth re-designing this. Instead of checking the interface type within IPAM, it should be called by the main plugins after IPAM allocated the IP address.

steveeJ:
I've restructured the code and started to add tests for the `ip` package. Not sure if ipam.ConfigureIface` should also go into `ip`.

Reviews welcome at this point! @dcbw

steveeJ:
ping @containernetworking/cni-maintainers

/cc @squeed you're invited to review this too!

pkg/utils/hwaddr/hwaddr.go
```diff
@@ -0,0 +1,68 @@
...
+	case ip.To4() == nil:
+		return nil, SupportIp4OnlyErr{msg: "GenerateHardwareAddr4 only supports valid IPv4 address as input"}
+
+	case len(prefix) != len(PrivateMACPrefix):
+		return nil, InvalidPrefixLengthErr{msg: fmt.Sprintf(
+			"Prefix has length %d instead  of %d", len(prefix), len(PrivateMACPrefix)),
+		}
+	}
+
+	mac := prefix
```
squeed: This string conversion is a little awkward. Both of these types are just wrappers around `[]byte`. So, you can do:

```
ip = ip.To4()
addr := (net.HardwareAddr)(append([]byte{0x0a, 0x58}, ip...))
```

This removes one of the `err` types

steveeJ: I can't see it working that easily because we want to strip out the `.` characters.

squeed: It really does :-). The byte arrays are the real numerical address value, not the string.

pkg/ip/link_test.go
```diff
@@ -129,4 +133,55 @@ var _ = Describe("Link", func() {
 			return nil
 		})
 	})
+
+	It("SetHWAddrByIP must change the interface hwaddr and be predictable", func() {
+		var ()
```
dcbw: Hmm?

pkg/ip/link_test.go
```diff
@@ -129,4 +133,55 @@ var _ = Describe("Link", func() {
...
+
+		_ = containerNetNS.Do(func(ns.NetNS) error {
+			defer GinkgoRecover()
+
+			getIP := func(linkname string) string {
```
dcbw: Shouldn't this be getHwAddr()?  It's returning the hardware address not the IP.

pkg/ip/link_test.go
```diff
@@ -129,4 +133,55 @@ var _ = Describe("Link", func() {
...
+
+			Expect(hwaddrBefore).NotTo(Equal(hwaddrAfter1))
+			Expect(hwaddrAfter1).To(Equal(ip4onehwaddr))
+
+			return nil
+		})
+	})
+
+	It("SetHWAddrByIP must be injective", func() {
+		var ()
```
dcbw: Seems useless...

pkg/ip/link_test.go
```diff
@@ -129,4 +133,55 @@ var _ = Describe("Link", func() {
...
+
+		_ = containerNetNS.Do(func(ns.NetNS) error {
+			defer GinkgoRecover()
+
+			getIP := func(linkname string) string {
```
dcbw: getHwAddr()?

pkg/utils/hwaddr/hwaddr.go
```diff
@@ -51,18 +54,10 @@ func GenerateHardwareAddr4(ip net.IP, prefix string) (net.HardwareAddr, error) {
...
-	}
-
-	hwAddr, err := net.ParseMAC(mac)
-	if err != nil {
-		return nil, MacParseErr{msg: fmt.Sprintf(
-			"Failed to parse mac address %q generated based on IP %q due to: %v", mac, ip, err),
-		}
-	}
-	return hwAddr, nil
+	ipByteLen := len(ip)
```
dcbw: You need to To4() the IP address, otherwise you can technically be passed a 128-bit IPv4 address which is really just an IPv4 address with 12 bytes of 0 on the front).

dcbw:
Looks mostly good to me.  You could squash some of the commits together since some are just fixups based on review comments.  And the coveralls thing...

steveeJ:
@dcbw I squashed as much as I felt comfortable doing so. (hopefully) last review please ;-)

dcbw:
GenerateHardwareAddr4()

dcbw:
@steveeJ see my comment here https://github.com/containernetworking/cni/pull/254/commits/924b30b57d8752bb4d375cc0aeb5ead515903f79#r71946686 , that still needs to be addressed.

dcbw:
After that, LGTM.

pkg/utils/hwaddr/hwaddr.go
steveeJ: Wouldn't the `case ip.To4() == nil` case catch this?

squeed: No, @dcbw is right. You need to use the results of To4().

squeed: Well, except that the last bytes are extracted, so it doesn't matter if there are the extra null bytes. So, it works, and I was wrong.

steveeJ: I've updated the PR with a test that shows that the v4 in v6 []byte is processed correctly.

steveeJ:
@dcbw 

> see my comment here 924b30b#r71946686 , that still needs to be addressed.
> After that, LGTM.

That case is already covered because the function in question only uses the last 4 bytes. I added a test to demonstrate the correctness.

/cc @squeed

squeed:
👍 LGTM

PR #255, opened 2016-06-23
added custom attributes for exchanging more information between ipam …

michaelhenkel:
…module and plugin
For some plugins more information must be exchanged between IPAM and plugin.
For OpenContrail the CNI OpenContrail IPAM module creates the virtual networks and assigns Ip addresses. 
The plugin itself creates the virtual port object in the database.
In order to assign the virtual port to the virtual network and the ip address to the port the plugin must know the UUID of the virtual network and virtual port the IPAM module created.
The CUSTOM attributes allow sending Name:Value information as part of the result struct from IPAM to plugin.

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -67,9 +67,10 @@ type NetConf struct {
 
 // Result is what gets returned from the plugin (via stdout) to the caller
 type Result struct {
-	IP4 *IPConfig `json:"ip4,omitempty"`
-	IP6 *IPConfig `json:"ip6,omitempty"`
-	DNS DNS       `json:"dns,omitempty"`
+	IP4 		*IPConfig `json:"ip4,omitempty"`
+	IP6 		*IPConfig `json:"ip6,omitempty"`
+	DNS 		DNS       `json:"dns,omitempty"`
+	CUSTOMATTR  	*CUSTOM    `json:"custom,omitempty"`
 }
 
 func (r *Result) Print() error {
@@ -87,6 +88,9 @@ func (r *Result) String() string {
 	if r.IP6 != nil {
 		str += fmt.Sprintf("IP6:%+v, ", *r.IP6)
 	}
+        if r.CUSTOMATTR != nil {
+		str += fmt.Sprintf("CUSTOMATTR:%+v, ", *r.CUSTOMATTR)
+        }
 	return fmt.Sprintf("%sDNS:%+v", str, r.DNS)
 }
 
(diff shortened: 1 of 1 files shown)


rosenhouse:
This looks related to the discussion at this week's CNI meeting about IPAM plugins returning arbitrary data.  @steveeJ

michaelhenkel:
Any outcome from the meeting?

bboreham:
It sounds like your net and ipam plugins would only work as a pair: they would not be composable with alternate implementations. If so, why not just combine them into the net plugin?

pkg/types/types.go
```diff
@@ -67,9 +67,10 @@ type NetConf struct {
...
 // Result is what gets returned from the plugin (via stdout) to the caller
 type Result struct {
-	IP4 *IPConfig `json:"ip4,omitempty"`
-	IP6 *IPConfig `json:"ip6,omitempty"`
-	DNS DNS       `json:"dns,omitempty"`
+	IP4 		*IPConfig `json:"ip4,omitempty"`
+	IP6 		*IPConfig `json:"ip6,omitempty"`
+	DNS 		DNS       `json:"dns,omitempty"`
+	CUSTOMATTR  	*CUSTOM    `json:"custom,omitempty"`
 }
```
dcbw: Two things here:

1. I think CUSTOMATTR should just be a dictionary, eg 'map[string]interface{}' in code.  That should marshal correctly to JSON an allow plugins to marshal their own structs into the dict.
2. It should just be named something like 'Args' or 'Extra' (no full capitalization required)
3. The JSON name should be the same as the Go struct member name, just without the initial capital.  Eg if the member is "Args" the JSON would be json:"args,omitempty".

pkg/types/types.go
```diff
@@ -105,6 +109,15 @@ type DNS struct {
...
 
+type Attribute struct {
+	Name 	string
+	Value 	string
+}
+
+type CUSTOM struct {
+        CustomAttributes []Attribute
+}
+
```
dcbw: If the extra arguments/results is a plain dict, then we don't need these extra structures.

dcbw:
> It sounds like your net and ipam plugins would only work as a pair: they would not be composable with alternate implementations. If so, why not just combine them into the net plugin?

We discussed this a bit in the CNI maintainers meeting today.

We think extra args are OK for passing information back from the CNI plugin(s) to the runtime just like https://github.com/containernetworking/cni/pull/247 did for passing information from the runtime to the plugin.

But we don't think it's a good mechanism for passing information between individual CNI plugins, since in that case if one plugin depends on args from a wrapped plugin then yes, they should probably be combined into one plugin.

pkg/types/types.go
steveeJ: I agree with @dcbw's here and suggest to use `Args`, so there is consistency with the recently introduced change to the input configuration. Resulting in this line
```Go
  	Args  	map[string]interface{}    `json:"args,omitempty"` 
```

Although on top of my head I'm not sure if we need to value type to be of `interface{}` or if `string` would suffice.

squeed:
It seems we've basically implemented this with the `args` information; closing.

PR #257, opened 2016-06-23
avoid allocating .0 and .255 ipv4 addresses with host-local.

### LINKED ISSUE #256: host-local and large subnets
with a range larger than /24, eg : 
```
{
    "name": "test",
    "type": "bridge",
    "bridge": "cni0",
    "ipMasq": true,
    "ipam": {
        "type": "host-local",
        "subnet": "10.1.0.0/16",
        "rangeStart": "10.1.2.2",
        "rangeEnd": "10.1.4.254",
        "gateway": "10.1.0.1",
        "routes": [
            { "dst": "0.0.0.0/0" }
        ]
    }
}
```

host-local allocate bad ips, eg 10.1.2.255 or 10.1.3.0

vixns:
fix #256

### DIFF
--- a/plugins/ipam/host-local/allocator.go
+++ b/plugins/ipam/host-local/allocator.go
@@ -119,12 +119,12 @@ func (a *IPAllocator) Get(id string) (*types.IPConfig, error) {
 			continue
 		}
 
-                // don't allocate .0 or .255 IPv4
-                if ip4 := cur.To4(); ip4 != nil {
-                        if ip4[3] == 0 || ip4[3] == 255 {
-                                continue
-                        }
-                }
+		// don't allocate .0 or .255 IPv4
+		if ip4 := cur.To4(); ip4 != nil {
+			if ip4[3] == 0 || ip4[3] == 255 {
+				continue
+			}
+		}
 
 		reserved, err := a.store.Reserve(id, cur)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


dcbw:
Could you run "gofmt -s -w plugins/ipam/host-local/allocator.go" to fix up the formatting?

vixns:
done

dcbw:
Although the allocator is supposed to skip the start and end addresses, if you pass larger subnets it will still allow 00 and FF bytes in the middle, so this seems reasonable.

Could you also add the following testcase to plugins/ipam/host-local/allocator_test.go under the `It("should allocate ips in round robin` case?

```golang
		It("should not allocate start and end addresses of a subnet", func() {
			subnet, err := types.ParseCIDR("10.0.0.0/22")
			Expect(err).ToNot(HaveOccurred())
			conf := IPAMConfig{
				Name:   "test",
				Type:   "host-local",
				Subnet: types.IPNet{IP: subnet.IP, Mask: subnet.Mask},
			}
			store := fakestore.NewFakeStore(make(map[string]string), nil)
			alloc, _ := NewIPAllocator(&conf, store)

			for i := 0; i <= 1024; i++ {
				res, err := alloc.Get("ID")
				if err != nil {
					// 1022 usable IPs in a /22
					Expect(i).To(BeNumerically(">=", 1022))
				} else {
					Expect(res.IP.IP[3]).NotTo(Equal(byte(0)))
					Expect(res.IP.IP[3]).NotTo(Equal(byte(255)))
				}
			}
		})
````

Also, please squash the commits into a single commit with 'git rebase'.

Lastly, please prefix the commit message with the module it touches, like:

`host-local: avoid allocating .0 and .255 IPv4 addresses`

PR #261, opened 2016-06-30
plugins/main/host-local: don't always assign gw

steveeJ:
The automatic assignment of the GW should not be enforced by the
host-local plugin. This commit stops the allocator from setting the gw in the
IPAM result if it hasn't been passed.

/cc @freehan

### DIFF
--- a/plugins/ipam/host-local/allocator.go
+++ b/plugins/ipam/host-local/allocator.go
@@ -74,9 +74,6 @@ func (a *IPAllocator) Get(id string) (*types.IPConfig, error) {
 	defer a.store.Unlock()
 
 	gw := a.conf.Gateway
-	if gw == nil {
-		gw = ip.NextIP(a.conf.Subnet.IP)
-	}
 
 	var requestedIP net.IP
 	if a.conf.Args != nil {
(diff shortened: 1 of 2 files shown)


dcbw:
This behavior was a nice shortcut actually, but I can see where it may not be desirable.  But it might trip people up if they were relying on the old behavior to set the gateway route in the container without having to specify the gateway.

Maybe instead the auto gw behavior should be tied to isGateway?  If isGateway=true but you don't pass a gateway in IPAM details, that would indicate you want the bridge to be the gateway (isGateway = "assign an IP address to the bridge. Defaults to false.") but you didn't pass any address to assign, so maybe do it automatically.

squeed:
Hmm, this is definitely a dark corner of the spec. We are certainly relying on the existing behavior in rkt's default network configuration, and I suspect we are not the only ones.

Spec version time?

```json
{
	"name": "default",
	"type": "ptp",
	"ipMasq": true,
	"ipam": {
		"type": "host-local",
		"subnet": "172.16.28.0/24",
		"routes": [
			{ "dst": "0.0.0.0/0" }
		]
	}
}
```

squeed:
I don't think we want this; and I'm certain we don't want this to be the default. However, feel free to re-open against the plugins repository.

Issue #266, opened 2016-07-11
Versioning!


rosenhouse:
I'm opening this issue so we can track and discuss work on versioning as first discussed at the [CNI meeting on 2016-06-29](https://docs.google.com/document/d/1_FDwd1QuFRxT_f40Wq4L6Q2jL2jqIDOoVhNBB26TViE/edit#heading=h.h0y3jbyyk1iw).  This is in part to address current confusion over versioning (e.g. https://github.com/containernetworking/cni/issues/150).

I will update this issue as we make progress.  Please leave general feedback here, and use the PRs for specific code review feedback.

Complete:
- ~~add `VERSION` command, #267~~
- ~~Plugin and runtime library support for plugins to report support for multiple versions of the CNI spec, #287~~
- ~~Plugins validate `cniVersion` of their input config, #296~~
- ~~Automated tests verify that `libcni` is backwards compatible with legacy plugins, #301~~

In flight:
- Modify #145 to use new versioning tooling.

Proposed next steps:
- Update spec with answers to these questions:
  - what do plugin authors need to do about versioning?
  - what do container runtime maintainers need to do about versioning?
  - under what circumstances will the spec version will change?
- Automated tests in the CNI repo verify version correctness of the plugins that we maintain.
- The runtime-facing `libcni` functions (`AddNetwork` and `DelNetwork`) reconcile the versions:
  - Reported by the plugin's VERSION command
  - Contained in the config
  - In use by the `libcni` package
  
  So that exactly one of the following occurs:
  - **Either**: the plugin is called with config matching a version in its VERSION list
  - **Or**: the function's caller receives an error that would be meaningful to a user, e.g. 
    
    ```
     the CNI library in use (spec 0.2.0) is not compatible with the plugin "foobar"
     (plugin reports compatibility with spec versions [0.5.0, 0.6.0])
    ```

References:
- Draft [document collecting uses cases](https://docs.google.com/document/d/1fKRHQ10vDgbGED4NwJU_EbPbt-J0fK8prHod_6yVFf0/edit#heading=h.hf1r2to8das2)
- [CNI meeting notes](https://docs.google.com/document/d/1_FDwd1QuFRxT_f40Wq4L6Q2jL2jqIDOoVhNBB26TViE/edit)

bboreham:
I'd like to suggest that failure to handle the VERSION command shall be interpreted the same as returning "0.1.0", so that older plugins do not become immediately useless.

squaremo:
[...]

PR #267, opened 2016-07-11
Adds VERSION command

rosenhouse:
At last CNI meeting, I volunteered to put together a first draft of spec versioning support, see https://github.com/containernetworking/cni/issues/266

This small PR is the first step -- add a VERSION command to all plugins transparently -- plugin authors only need to re-compile.

The `VERSION` is bumped to `0.2.0` because this new command was added.  There are no incompatibilities with `0.1.0`

### DIFF
--- a/Documentation/tuning.md
+++ b/Documentation/tuning.md
@@ -22,9 +22,7 @@ Other sysctls can be modified as long as they belong to the network namespace (`
 
 A successful result would simply be:
 ```
-{
-  "cniVersion": "0.1.0"
-}
+{ }
 ```
 
 ## Network sysctls documentation
--- a/SPEC.md
+++ b/SPEC.md
@@ -62,9 +62,14 @@ The operations that the CNI plugin needs to support are:
     - **Extra arguments**, as defined above.
     - **Name of the interface inside the container**, as defined above.
 
+- Report version
+  - Parameters: NONE.
+  - Result:
+    - The version of the CNI spec implemented by the plugin: `{ "cniVersion": "0.2.0" }`
+
 The executable command-line API uses the type of network (see [Network Configuration](#network-configuration) below) as the name of the executable to invoke.
 It will then look for this executable in a list of predefined directories. Once found, it will invoke the executable using the following environment variables for argument passing:
-- `CNI_COMMAND`: indicates the desired operation; either `ADD` or `DEL`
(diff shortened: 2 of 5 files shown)


rosenhouse:
TODO: figure out how to get coveralls to recognize that coverage of `skel` package has actually increased.  Likely by adding some kind of instrumentation to [this line](https://github.com/containernetworking/cni/pull/267/files#diff-73853e3f30a13566f553f54218ff940bR34).

steveeJ:
@rosenhouse I don't think you can get the code coverage of executed binaries to be recognized. I think that there should be both, tests that call code directly and tests that use the binaries. Why did you replace the current approach instead of adding the new one?

rosenhouse:
@steveeJ Ok, I was afraid that was the case.

I was trying to avoid having duplicate test coverage of the same behavior -- but I will re-work it to have both in-process and inter-process tests.

rosenhouse:
@steveeJ & @containernetworking/cni-maintainers  I'm re-writing this feature on top of the refactor I introduced in #269.  A prompt review of that PR would help me move forward here.  Thanks!

rosenhouse:
Based on the newly-refactored `skel` package, I've re-done this feature with unit test coverage.

One #270 is in place, we can add corresponding support in `libcni`.

Please take a look!

steveeJ:
@rosenhouse Pointing you at the _missing/spec change_ label I've added to this a week ago. Is it something you left out intentionally? The version string returned also suggests that you consider this a version bump.

rosenhouse:
@steveeJ thanks for pointing that out.  I've updated the spec.  let me know what you think.

dcbw:
LGTM.

Though would it be wild to pass back an array of versions that the plugin could support, and keep CNI_VERSION?  That way a plugin that supports multiple versions in a single binary could work with a broader set of runtimes.

rosenhouse:
@dcbw That's basically what I had in mind for a near-term next step in #266, by adding additional fields to the returned JSON data.

But first I just want to report the version of the CNI library that was compiled into the plugin.

rosenhouse:
@steveeJ would you like to take a second look?

dcbw:
> But first I just want to report the version of the CNI library that was compiled into the plugin.

OK, fair enough.

tomdee:
LGTM

PR #270, opened 2016-07-15
Add integration tests of libcni

rosenhouse:
The `libcni` package provides the public API used by container runtimes, including loading configuration files from disk and executing plugins.

It is currently untested.  This PR adds integration tests, primarily to catch any regressions that might arise as we add new functionality (e.g. #267).

The tests of the plugin execution use a new "no-op" plugin.  The no-op plugin does not touch namespaces or the network stack.  Instead it records its inputs and can be configured to respond with arbitrary data or to error.  The tests use it to verify behavior of the plugin execution code.

### DIFF
--- a/plugins/test/noop/noop_test.go
+++ b/plugins/test/noop/noop_test.go
@@ -29,15 +29,16 @@ import (
 
 var _ = Describe("No-op plugin", func() {
 	var (
-		cmd           *exec.Cmd
-		debugFileName string
-		debug         *noop_debug.Debug
+		cmd             *exec.Cmd
+		debugFileName   string
+		debug           *noop_debug.Debug
+		expectedCmdArgs skel.CmdArgs
 	)
 
+	const reportResult = `{ "ip4": { "ip": "10.1.2.3/24" }, "dns": {} }`
+
 	BeforeEach(func() {
-		debug = &noop_debug.Debug{
-			ReportResult: `{ "ip4": { "ip": "10.1.2.3/24" }, "dns": {} }`,
-		}
+		debug = &noop_debug.Debug{ReportResult: reportResult}
 
 		debugFile, err := ioutil.TempFile("", "cni_debug")
 		Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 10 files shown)


rosenhouse:
@containernetworking/cni-maintainers would appreciate a review by 1 other maintainer -- this is just adding test coverage, no functional changes here.

rosenhouse:
@dcbw mind reviewing this?  Thanks!

plugins/test/noop/noop_test.go
```diff
@@ -0,0 +1,135 @@
...
+	AfterEach(func() {
+		os.Remove(debugFileName)
+	})
+
+	It("responds to ADD using the ReportResult debug field", func() {
+		session, err := gexec.Start(cmd, GinkgoWriter, GinkgoWriter)
+		Expect(err).NotTo(HaveOccurred())
+		Eventually(session).Should(gexec.Exit(0))
+		Expect(session.Out.Contents()).To(MatchJSON(`{
+				 "ip4": { "ip": "10.1.2.3/24" },
```
dcbw: This gets used above too, maybe a constant?

plugins/test/noop/noop_test.go
```diff
@@ -0,0 +1,135 @@
...
+		Expect(debugFile.Close()).To(Succeed())
+		debugFileName = debugFile.Name()
+
+		Expect(debug.WriteDebug(debugFileName)).To(Succeed())
+
+		cmd = exec.Command(pathToPlugin)
+		cmd.Env = []string{
+			"CNI_COMMAND=ADD",
+			"CNI_CONTAINERID=some-container-id",
+			"CNI_ARGS=DEBUG=" + debugFile.Name(),
```
dcbw: debugFileName?

rosenhouse:
@dcbw thanks for the review, I've cleaned those things up.

plugins/test/noop/noop_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	It("records all the args provided by skel.PluginMain", func() {
+		session, err := gexec.Start(cmd, GinkgoWriter, GinkgoWriter)
+		Expect(err).NotTo(HaveOccurred())
+		Eventually(session).Should(gexec.Exit(0))
+
+		debug, err := noop_debug.ReadDebug(debugFileName)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(debug.Command).To(Equal("ADD"))
+		Expect(debug.CmdArgs).To(Equal(skel.CmdArgs{
```
dcbw: Since this skel.CmdArgs gets used a couple times, and since it's dependent on the data passed in BeforeEach() far above, perhaps it should be defined once after cmd.Env and then only matched here and in the DEL check?

dcbw:
@rosenhouse Mostly LGTM, just a few small comments.

rosenhouse:
@dcbw Thank you!  Good points, I've fixed up all of them.

PR #259, opened 2016-06-29
Add SR-IOV CNI plugin

hustcat:
Add SR-IOV CNI plugin.  [#netlink](https://github.com/vishvananda/netlink) needed to be bump to the latest.

PR #272, opened 2016-07-22
README: distinguish between runtimes and plugins

rosenhouse:
Following up on a suggestion I made in https://github.com/containernetworking/cni/pull/259#issuecomment-234377787 I've expanded the README's list of CNI users to explicitly call out 3rd party plugins.

The idea being that we can encourage plugin authors to publish their code in their own source code repository and then we can link to it.  This reduces the maintenance burden on us, but still provides a informal index for the plugin ecosystem.


tomdee:
This looks good to me - merging.

PR #273, opened 2016-07-29
Reconfigure bridge IP address

zreigz:
Instead of throwing error it is better reconfigure the bridge IP address.
Currently the CNI plugin is used in kubernetes and after hyperkube restart cni0 bridge retains its old value. This PR fix the issue.

### DIFF
--- a/libcni/conf.go
+++ b/libcni/conf.go
@@ -83,3 +83,28 @@ func LoadConf(dir, name string) (*NetworkConfig, error) {
 	}
 	return nil, fmt.Errorf(`no net configuration with name "%s" in %s`, name, dir)
 }
+
+func InjectConf(original *NetworkConfig, key string, newValue interface{}) (*NetworkConfig, error) {
+	config := make(map[string]interface{})
+	err := json.Unmarshal(original.Bytes, &config)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshal existing network bytes: %s", err)
+	}
+
+	if key == "" {
+		return nil, fmt.Errorf("key value can not be empty")
+	}
+
+	if newValue == nil {
+		return nil, fmt.Errorf("newValue must be specified")
+	}
+
+	config[key] = newValue
+
+	newBytes, err := json.Marshal(config)
+	if err != nil {
+		return nil, err
+	}
+
+	return ConfFromBytes(newBytes)
+}
--- a/libcni/conf_test.go
+++ b/libcni/conf_test.go
@@ -27,8 +27,9 @@ import (
 
 var _ = Describe("Loading configuration from disk", func() {
 	var (
-		configDir    string
-		pluginConfig []byte
+		configDir     string
+		pluginConfig  []byte
+		testNetConfig *libcni.NetworkConfig
 	)
 
 	BeforeEach(func() {
@@ -38,6 +39,9 @@ var _ = Describe("Loading configuration from disk", func() {
 
 		pluginConfig = []byte(`{ "name": "some-plugin", "some-key": "some-value" }`)
 		Expect(ioutil.WriteFile(filepath.Join(configDir, "50-whatever.conf"), pluginConfig, 0600)).To(Succeed())
+
+		testNetConfig = &libcni.NetworkConfig{Network: &types.NetConf{Name: "some-plugin"},
+			Bytes: []byte(`{ "name": "some-plugin" }`)}
 	})
 
 	AfterEach(func() {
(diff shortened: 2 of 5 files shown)


zreigz:
cc @luxas @cheld

plugins/main/bridge/bridge.go
```diff
@@ -74,8 +74,21 @@ func ensureBridgeAddr(br *netlink.Bridge, ipn *net.IPNet) error {
 			if a.IPNet.String() == ipnStr {
 				return nil
 			}
+
+			// reconfigure IP address when some old value is set
+			if err = netlink.LinkSetDown(br); err != nil {
+				return fmt.Errorf("could not down bridge %q: %v", br.Name, err)
```
luxas: maybe: `Could not turn down bridge...`

plugins/main/bridge/bridge.go
```diff
@@ -74,8 +74,21 @@ func ensureBridgeAddr(br *netlink.Bridge, ipn *net.IPNet) error {
...
+			}
+
+			addr := &netlink.Addr{IPNet: a.IPNet, Label: ""}
+
+			if err := netlink.AddrDel(br, addr); err != nil {
+				return fmt.Errorf("could not remove IP address from %q: %v", br.Name, err)
+			}
+			if err = netlink.LinkSetUp(br); err != nil {
+				return fmt.Errorf("could not up bridge %q: %v", br.Name, err)
```
luxas: `Could not turn up bridge...`

luxas:
LGTM, this is a required feature when simplifying Kubernetes deployments

@steveeJ @dcbw @tomdee

bboreham:
Is this a safe thing to do if what has really happened is the admin has supplied incorrect configuration?  Or two configurations with different gateway addresses?

plugins/main/bridge/bridge.go
dcbw: I've usually heard/seen "set up" or "set down" instead of "turn"

rosenhouse:
Perhaps there could be a new boolean field on the `NetConf` like `Force` or `ForceAddress`  which toggles this behavior? 

Also, please follow our [contribution guidelines](https://github.com/containernetworking/cni/blob/master/CONTRIBUTING.md#acceptance-policy).  In particular, please add test coverage of this change and use the preferred format for commit messages.

zreigz:
Thanks for review. I will add tests and use the preferred format for commit message. The `ForceAddress` flag sounds good for me.

luxas:
What do we need to do on the k8s side then? I'd prefer an option that is not visible to the "user", instead maybe an option that is set by k8s code. 

Or am I missing something?

bboreham:
@luxas I believe it would be similar to `IgnoreUnknown` here: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/network/cni/cni.go#L278

luxas:
Yep, if we put the option on RuntimeConf, then everything's alright :)

zreigz:
PTAL

plugins/main/bridge/bridge.go
```diff
@@ -52,15 +54,16 @@ func init() {
 
 func loadNetConf(bytes []byte) (*NetConf, error) {
 	n := &NetConf{
-		BrName: defaultBrName,
+		BrName:       defaultBrName,
+		ForceAddress: defaultForceAddress,
```
rosenhouse: This isn't strictly necessary, as `bool` default to `false` when not present.

rosenhouse:
Besides one small thing regarding defaults, this LGTM.  @dcbw what do you think?

But.... @luxas @bboreham  I think you may be mistaken about `RuntimeConf`.

The change which I suggested, and which @zreigz has implemented here, is a change to the plugin's `NetConf`, which is supplied to the plugin process as JSON on stdin.

Container runtimes (e.g. Kube) configure that using the `libcni.NetworkConfig` struct, not the `libcni.RuntimeConf` struct.

However, just because the option is consumed from `NetConf` doesn't mean that a user must set it.  Kube could inject that value into the `NetworkConfig` itself, before calling `AddNetwork()`.  Here's [an example of this in CloudFoundry](https://github.com/cloudfoundry-incubator/netman-release/blob/936ce5f/src/guardian-cni-adapter/controller/cni.go#L69-L105).

Unfortunately, the CNI library doesn't it very easy to do this.  We really ought to provide a utility function for container runtimes to solve this problem more generally, e.g.
```go
func InjectConf(original *NetworkConfig, key string, newValue interface{}) (*NetworkConfig, error) {
    // json.Unmarshal original.Bytes to a map[string]interface{}
    // set conf[key] = newValue
    // json.Marshal back to bytes
    // return new NetworkConfig with those bytes
}
```

Feel free to add this sort of function to `libcni` as part of this PR (but be sure to include tests too).  Or just do your special case in Kubelet.

dcbw:
@luxas the k8s "kubenet" plugin has a function here https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/network/kubenet/kubenet_linux.go#L243 that reconfigures the bridge's IP address before calling CNI ADD to ensure the plugin doesn't error out when the bridge address is unexpected.  This condition isn't expected during runtime (since changing the ClusterCIDR for your Kubernetes cluster will effectively render all pods mute) but can happen if you stop kubelet, change the Cluster CIDR, and then restart kubelet.  In that case the bridge will have the wrong address.

Once this change lands, we can revendor CNI in k8s and start using the new option instead of doing the address removal in kubenet itself.

plugins/main/bridge/bridge.go
zreigz: Done

zreigz:
PTAL

zreigz:
Regarding to `InjectConf` function I would like add it as a separate PR (Go language it is not my first programing language so I learn it when I prepare PR. It can take some time so it would be better for me to make second PR).  I assume this function should be added here: https://github.com/containernetworking/cni/blob/39a8fe476638e5ff315bb879e498c46ff842f242/libcni/conf.go

zreigz:
I've added 'InjectConf' function to this PR. PTAL

luxas:
I think this is really fine. Haven't reviewed the tests much, though.

I hope we can get this in today, so Kubernetes may re-vendor and get this functionality into the next alpha release.

bboreham:
@luxas @dcbw minor nit-pick, but do you really need Kubernetes to _vendor_ this source code?  I would think you need them to update their deployment scripts to pick up a new _build_, and to add the flag to turn it on.

jellonek:
@bboreham rather _build in particular version_ but yes, both - `rkt` and `kubernetes` imo should be builded in that way.

luxas:
Kubernetes does vendor in (lib)cni: https://github.com/kubernetes/kubernetes/blob/2480ef5f1ded8f17f21e56ef69b42898ba93024d/pkg/kubelet/network/cni/cni.go

I meant we should re-vendor CNI for the `InjectConf` function, and update k8s code to add the `ForceAddress` option to NetConf. Obviously we also need a newer build of the CNI binaries in order to get the ForceAddress functionality.

Thanks

bboreham:
OK, sorry, I missed the whole conversation about putting it on `NetworkConfig`.  

I don't see why you need all that complicated machinery - Kubernetes generates the bridge config like this: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/network/kubenet/kubenet_linux.go#L184 and you can just add the `ForceAddress` line.

luxas:
Yes, but still
a) this patch has to be merged
b) libcni is already vendored into k8s, so we can take advantage of the InjectConf method to avoid duplicating code
c) we want to avoid having to update/maintain manifests that are user-specified (for generic cni, not kubenet) in order to have k8s function properly.

libcni/conf_test.go
```diff
@@ -107,4 +111,38 @@ var _ = Describe("Loading configuration from disk", func() {
...
+			})
+
+			It("returns newValue  error", func() {
+				_, err := libcni.InjectConf(testNetConfig, "test", nil)
+				Expect(err).To(MatchError(HavePrefix(`newValue must be specified`)))
+			})
+		})
+
+		Context("when new string value added", func() {
+			It("returns unmarshal error", func() {
```
rosenhouse: Maybe this should say `"It adds the key & value to the config"`?

dcbw:
> minor nit-pick, but do you really need Kubernetes to vendor this source code? I would think you need them to update their deployment scripts to pick up a new build, and to add the flag to turn it on.

@bboreham yes, I mis-spoke.  The CNI binaries in k8s are salted into deployments but distributed as binaries.  k8s does use some vendored CNI code, but not the plugins.

libcni/conf.go
```diff
@@ -83,3 +83,31 @@ func LoadConf(dir, name string) (*NetworkConfig, error) {
...
+	}
+
+	config[key] = newValue
+
+	newBytes, err := json.Marshal(config)
+	if err != nil {
+		return nil, err
+	}
+
+	return &NetworkConfig{
```
dcbw: So this will return the new bytes with the injected value, but return the original network config without the injected value.  That might cause confusion?

rosenhouse: Good point.  Maybe an additional test should cover the case where the injected value is actually a sub-field of `NetworkConfig.Network`.

libcni/conf_test.go
zreigz: Yes my mistake I will change it

libcni/conf.go
zreigz: No problem I will add this kind of test

libcni/conf_test.go
```diff
@@ -107,4 +111,38 @@ var _ = Describe("Loading configuration from disk", func() {
...
+				newPluginConfig := []byte(`{"name":"some-plugin","some-key":"some-value","test":"test"}`)
+				resultConfig, err := libcni.InjectConf(testNetConfig, "test", "test")
```
luxas: There are two more cases you could add:
```
// ensure an existing key-value pair isn't modified if the InjectConf tries to inject the current value
before []byte(`{ "name": "some-plugin", "some-key": "some-value" }`)
libcni.InjectConf(testNetConfig, "some-key", "some-value")
after []byte(`{ "name": "some-plugin", "some-key": "some-value" }`)

// ensure an existing key is updated
before []byte(`{ "name": "some-plugin", "some-key": "some-value" }`)
libcni.InjectConf(testNetConfig, "some-key", "a-new-value")
after []byte(`{ "name": "some-plugin", "some-key": "a-new-value" }`)
```

zreigz:
I've changed a little bit implementation to return reloaded configuration (not only bytes)
Can it stay like this or there is any purpose to return only updated bytes array ?

luxas:
This LGTM

zreigz:
PTAL

zreigz:
@rosenhouse I've made necessary changes. Please take a look.

rosenhouse:
Yeah this looks great.  Any other @containernetworking/cni-maintainers want to give the second 👍 ?

tomdee:
LGTM - Merging

PR #279, opened 2016-08-06
Enable IPv6 in the host-local plugin

achanda:
The original PR is here https://github.com/containernetworking/cni/pull/137

cc @steveeJ @dcbw @tgraf PTAL

### DIFF
--- a/plugins/ipam/host-local/main.go
+++ b/plugins/ipam/host-local/main.go
@@ -15,6 +15,8 @@
 package main
 
 import (
+	"fmt"
+
 	"github.com/containernetworking/cni/plugins/ipam/host-local/backend/disk"
 
 	"github.com/containernetworking/cni/pkg/skel"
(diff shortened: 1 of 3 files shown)


andir:
Is there any ETA for this feature?

dcbw:
@containernetworking/cni-maintainers too...

Looks good to me in isolation.  But:

1) This is technically a spec change due to the addition of 'ipam6' to the config file; we need to update SPEC.md to add that key after maintainers agree.
2) in addition to the spec change, this PR should add stubs to the DHCP plugin to error out if DHCP is asked to do IPAM6

andir:
On 14:33 12.09.16, Dan Williams wrote:
> @containernetworking/cni-maintainers too...
> 
> Looks good to me in isolation.  But:
> 
> 1) This is technically a spec change due to the addition of 'ipam6' to the config file; we need to update SPEC.md to add that key after maintainers agree.
> 2) in addition to the spec change, this PR should add stubs to the DHCP plugin to error out if DHCP is asked to do IPAM6

If someone wishes to do DHCPv6 why shouldn't we or why can't we?

dcbw:
> If someone wishes to do DHCPv6 why shouldn't we or why can't we?

Two reasons: because the current DHCP plugin doesn't support IPv6, and second DHCPv6 must be used in conjunction with IPv6 Router Advertisements anyway, so much more thought needs to go into that part of the code.

The only reason this PR would need to touch the DHCP plugin is because it needs the spec change, which cascades down to the other internal plugins (of which there are only two, DHCP and host-local).

plugins/ipam/host-local/main.go
```diff
@@ -26,49 +28,99 @@ func main() {
...
+		ipConf, err := allocator.Get(args.ContainerID)
+		if err != nil {
+			return err
+		}
+		r.IP4 = ipConf
 	}
 
-	r := &types.Result{
-		IP4: ipConf,
+	if ipamConf6 != nil {
```
squeed: Be careful here, you can leak v4 IPs if anything in the v6 portion fails.

plugins/ipam/host-local/main.go
```diff
@@ -26,49 +28,99 @@ func main() {
...
+		if e != nil {
+			result[0] = e
+		}
 	}
 
-	return allocator.Release(args.ContainerID)
+	if ipamConf6 != nil {
+		store, err := disk.New(ipamConf6.Name)
+		if err != nil {
+			return err
```
squeed: This is also "leaky" - a failure here will cause any results from line 100 to be lost.

Instead, you should allocate your, IPAllocators first, *then* release.

plugins/ipam/host-local/main.go
achanda: Thanks for pointing that out. I was trying this out
```
        var allocator IPAllocator
        if ipamConf != nil {
                store, err := disk.New(ipamConf.Name)
                if err != nil {
                        return err
                }
                defer store.Close()

                allocator, err := NewIPAllocator(ipamConf, store)
                if err != nil {
                        return err
                }

                ipConf, err := allocator.Get(args.ContainerID)
                if err != nil {
                        return err
                }
                r.IP4 = ipConf
        }

        if ipamConf6 != nil {
                store, err := disk.New(ipamConf6.Name)
                if err != nil {
                        allocator.Release(args.ContainerID)
                        return err
                }
                defer store.Close()

                allocator6, err := NewIPAllocator(ipamConf6, store)
                if err != nil {
                        allocator.Release(args.ContainerID)
                        return err
                }
                ipConf6, err := allocator6.Get(args.ContainerID)
                if err != nil {
                        allocator.Release(args.ContainerID)
                        return err
                }
                r.IP6 = ipConf6
        }
```
Is there a better way to do this?

plugins/ipam/host-local/main.go
achanda: Assuming this will be a problem only when we have both v4 and v6, when we are in this block here, haven't we have already released the v4 address?

plugins/ipam/host-local/main.go
squeed: That's closer to what I had in mind, but will panic if there is no ipv4 ipamconf.
It's a question of style, but I tend to *first* acquire all handles, then perform the allocation. Something like

```go
var allocator4, allocator6 IPAllocator
var store4, store6 
if ipamConf4 {
    allocator4 = ...
    store4 = ...
}
if ipamConf6 {
    allocator6 = ...
    store6 = ...
}

if ipamConf4 != nil {
    ip4Conf, err := ip4allocator.Get ...
}
if ipamConf6 != nil {
    ip6Conf, err := ...
    if err != nil {
        if ipamConf4 != nil {
            allocator4.Release... 
        }
    }
}
```

plugins/ipam/host-local/main.go
squeed: I'm arguing that, in this case, you are no longer returning `result[0]` generated on line 102. I think you, again, want to get all the "resource handles" before proceeding with the "transaction".
It's not that the ipv4 won't be released, but that information is being lost.

squeed:
I could see the case for wanting to allocate an ipv6 address based off of the ipv4 address. Is that something we want to support, or am I bike-shedding?

dcbw:
> I could see the case for wanting to allocate an ipv6 address based off of the ipv4 address. Is that something we want to support, or am I bike-shedding?

@squeed I'd say bikeshedding for now :)  but certainly a valid scenario that I think we should address later.

vzcambria:
For IPv4 host-local IPAM, a static address is chosen via IP= parameter supplied via CNI_ARGS.  Is the intent to use IP as well for IPv6?  What about dual stack?

bboreham:
This is marked "needs review", but I think it has been reviewed; now it also needs upgrading to match the newer data structures.

rosenhouse:
Closing, because I believe this has been subsumed by https://github.com/containernetworking/plugins/pull/12

PR #280, opened 2016-08-07
pkg/ip: Return correct error if container name provided exists, and test cases

### LINKED ISSUE #155: failed to find a unique veth name
This error message happens to appear when the interface inside the container already exists, and this is supposed to be unreachable code... not sure what's exactly going on here.

What I'm doing is:
```
echo '{
  "bridge":"cbr0",
  "ipMasq":true,
  "ipam":{"routes":[{"dst":"0.0.0.0/0"}],"subnet":"10.240.1.0/24","type":"host-local"},
  "isGateway":true,
  "name":"kubenet",
  "type":"bridge"
}' \
  | env \
    CNI_COMMAND=ADD \
    CNI_CONTAINERID=5709bbff004a100c0a69837654f9d42c93cca6ff070f7a2509f2ebffb68ffe2a \
    CNI_NETNS=/proc/23995/ns/net CNI_IFNAME=eth12 CNI_PATH=/opt/cni/bin:/opt/bridge/bin \
   /opt/cni/bin/bridge
```
And this command works the first time, but returns the "failed to find a unique veth name" on the second run. If I bump `CNI_IFNAME`, it works again.

prateekgogia:
If interface name for a container provided by a user is already present,
Veth creation fails with incorrect error.
If os.IsExist error is returned by makeVethPair:
* Check for peer name, if exists generate another random peer name, 
* else, IsExist error is due to container interface present, return error.

Fixes #155

### DIFF
--- a/pkg/ip/link.go
+++ b/pkg/ip/link.go
@@ -41,6 +41,13 @@ func makeVethPair(name, peer string, mtu int) (netlink.Link, error) {
 	return veth, nil
 }
 
+func peerExists(name string) bool {
+	if _, err := netlink.LinkByName(name); err != nil {
+		return false
+	}
+	return true
+}
+
 func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err error) {
 	for i := 0; i < 10; i++ {
 		peerName, err = RandomVethName()
@@ -54,7 +61,11 @@ func makeVeth(name string, mtu int) (peerName string, veth netlink.Link, err err
 			return
 
 		case os.IsExist(err):
-			continue
+			if peerExists(peerName) {
+				continue
+			}
+			err = fmt.Errorf("container veth name provided (%v) already exists", name)
+			return
 
 		default:
 			err = fmt.Errorf("failed to make veth pair: %v", err)
--- a/pkg/ip/link_test.go
+++ b/pkg/ip/link_test.go
@@ -15,6 +15,8 @@
 package ip_test
 
 import (
+	"bytes"
+	"crypto/rand"
 	"fmt"
 	"net"
 
@@ -47,8 +49,9 @@ var _ = Describe("Link", func() {
 		hostVethName      string
 		containerVethName string
 
-		ip4one = net.ParseIP("1.1.1.1")
-		ip4two = net.ParseIP("1.1.1.2")
+		ip4one             = net.ParseIP("1.1.1.1")
+		ip4two             = net.ParseIP("1.1.1.2")
+		originalRandReader = rand.Reader
 	)
 
 	BeforeEach(func() {
@@ -59,6 +62,10 @@ var _ = Describe("Link", func() {
 
 		containerNetNS, err = ns.NewNS()
 		Expect(err).NotTo(HaveOccurred())
+
+		fakeBytes := make([]byte, 20)
+		//to be reset in AfterEach block
+		rand.Reader = bytes.NewReader(fakeBytes)
 
 		_ = containerNetNS.Do(func(ns.NetNS) error {
 			defer GinkgoRecover()
@@ -80,6 +87,7 @@ var _ = Describe("Link", func() {
 		Expect(containerNetNS.Close()).To(Succeed())
 		Expect(hostNetNS.Close()).To(Succeed())
 		ifaceCounter++
+		rand.Reader = originalRandReader
 	})
 
 	It("SetupVeth must put the veth endpoints into the separate namespaces", func() {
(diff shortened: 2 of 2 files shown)


rosenhouse:
@prateekgogia: Thank you for the fix.  Could you please update it to follow [our contribution guidelines](https://github.com/containernetworking/cni/blob/master/CONTRIBUTING.md)?  

Specifically:
- Please add test coverage for your change. 
- Please format your commit message(s) according to the preferred style.

For this change, the test case is non-trivial.  I'd suggest testing `ip.SetupVeth` by temporarily replacing `rand.Reader` in the [`BeforeEach` block](http://onsi.github.io/ginkgo/#structuring-your-specs).  You can replace it with a deterministic `io.Reader` (e.g. one that always returns 0s).  In the `It` blocks you can then call `SetupVeth` repeatedly and write expectations on the resulting behavior.  But `rand.Reader` is a process-global variable, so don't forget set it back to the original value (e.g. in the [`AfterEach()` block](http://onsi.github.io/ginkgo/#structuring-your-specs)) so as not to pollute other test cases that use it.

prateekgogia:
Hi @rosenhouse Thanks for the feedback. I am not familiar with Ginkgo testing framework. I will read about it and add the test cases soon.

prateekgogia:
Hi @rosenhouse - I have made the changes to commit message and added test cases, can you please review and let me know if anything else is required.

pkg/ip/link.go
```diff
@@ -41,6 +41,15 @@ func makeVethPair(name, peer string, mtu int) (netlink.Link, error) {
 	return veth, nil
 }
 
+func peerExists(name string) bool {
+
+	link, err := netlink.LinkByName(name)
+	if link == nil || err != nil {
```
rosenhouse: The test suite still passes if I simplify this to
```go
_, err := netlink.LinkByName(name)
if err != nil {
   ...
```

pkg/ip/link_test.go
```diff
@@ -102,6 +110,63 @@ var _ = Describe("Link", func() {
 		})
 	})
 
+	It("SetupVeth must return an Error if interface already exist in container ns", func() {
```
rosenhouse:
```go
It("returns an error if the container already has an interface with the same name", func() {
```

pkg/ip/link_test.go
```diff
@@ -102,6 +110,63 @@ var _ = Describe("Link", func() {
...
+			defer GinkgoRecover()
+
+			_, _, err := ip.SetupVeth(containerVethName, mtu, hostNetNS)
+			Expect(err.Error()).To(Equal(fmt.Sprintf("container veth name provided (%s) already exists", containerVethName)))
+
+			return nil
+		})
+	})
+
+	It("SetupVeth must return an Error if interface already exist in the host ns", func() {
```
rosenhouse:
```go
Context("when there is no name available for the host-side", func() {
   BeforeEach(func() {
        containerVethName += "0"
   })
   It("returns a useful error", func() {
          ...
```

pkg/ip/link_test.go
```diff
@@ -102,6 +110,63 @@ var _ = Describe("Link", func() {
...
+			//adding different interface to container ns
+			containerVethName = containerVethName + "0"
+			_, _, err := ip.SetupVeth(containerVethName, mtu, hostNetNS)
+			Expect(err.Error()).To(Equal("failed to move veth to host netns: file exists"))
+
+			return nil
+		})
+	})
+
+	It("SetupVeth should not return an Error when another veth pair is added to namespaces", func() {
```
rosenhouse:
```go
Context("when there is no name conflict for the host or container interfaces", func() {
   BeforeEach(func() {
	containerVethName += "0"
	rand.Reader = originalRandReader
   })
   It("successfully creates the second veth pair", func() {
      ...
```

rosenhouse:
Hi @prateekgogia,

Thank you so much, this test coverage is a huge improvement.

I noticed a small code simplification and also left some suggestions for improving the readability of the tests. The Ginkgo / Gomega BDD style is most valuable when the `Describe`, `Context` and `It` blocks all read as clear sentences.

I'd ask that you fix these up, and then squash all of your commits together into a single commit that follows the aforementioned style guidelines.

Thanks!

prateekgogia:
@rosenhouse: updated and squashed all commits in one commit. Thanks for the review :)

pkg/ip/link.go
```diff
@@ -41,6 +41,15 @@ func makeVethPair(name, peer string, mtu int) (netlink.Link, error) {
 	return veth, nil
 }
 
+func peerExists(name string) bool {
+
```
jellonek: Please remove this empty line.

pkg/ip/link.go
```diff
@@ -41,6 +41,15 @@ func makeVethPair(name, peer string, mtu int) (netlink.Link, error) {
...
+	_, err := netlink.LinkByName(name)
+	if err != nil {
```
jellonek: If any returned values (_, err) isn't used out of "if" block - this should be combined into:
```
  if _, err := netlink.LinkByName(name); err != nil {
```
Like shown in second example in https://golang.org/doc/effective_go.html#if

prateekgogia:
@rosenhouse @jellonek addressed.

rosenhouse:
LGTM.  Any other @containernetworking/cni-maintainers want to give the second 👍 ?

jellonek:
LGTM, but i'm not maintainer.

tomdee:
LGTM - merging.

Issue #281, opened 2016-08-09
How to find out ip address after it was assigned to a link?


dshulyak:
I am trying to understand who's responsibility is to store ip address after it was returned by cni ADD command. In kubernetes assigned ip needs to be known for multiple purposes and thus we need either:
1. store in database
   But what if assigned ip address is not static? With dhcp plugin it is certainly wont be
2. use other tools to query container namespace (in my case docker/rkt/nsenter)
   adds additional dependency either into container or host system (see https://github.com/kubernetes/kubernetes/pull/30225 for some context)
3. query application that returned ip

I like third way for several reasons:
- it hides the method which was used to assign ip address from cni clients, for example in bridge plugin application enters container network namespace and assigns ip to a link, and it is not my concern as a client
- cni is the source of ip, and in case of centralized control plane (dhcp for example) it would make sense to store it locally

Is there any plan/design how it will be addressed in long term?

jellonek:
CNI plugin which is responsible for creating NS/veth/tap/interfaces returns information about interface and it's mac address.
AFAIR now there is no other way to verify which address is connected with that mac address other than arp query, because of mentioned possible dynamic change of values returned by plugin performing ADD command - which returns values from IPAM, which could be in meanwhile adjusted to new values - like in dhcp example.

AFAIK there was discussion how to deal with dhcp which requires long running client even when container is paused or even shutted down (but not garbage collected) - but don't know how this discussion ended.

So for this moment searching arp table for mac address seems to be quite ugly but effective solution, which bypasses missing interface accessible from CNI.

dshulyak:
will it make sense to extend cni commands with GET that will have same input/output as ADD?
cc @steveeJ

freehan:
> will it make sense to extend cni commands with GET that will have same input/output as ADD? but will be used only for information

Yeah. Something like a STATUS command would be great.

leodotcloud:
@dshulyak 
Here are my 2 cents:
[...]

PR #282, opened 2016-08-12
Bridge plugin: if veth already exists - return information from corespodning ns

dshulyak:
addresses: https://github.com/containernetworking/cni/issues/281

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -29,6 +29,7 @@ import (
 	"github.com/containernetworking/cni/pkg/types"
 	"github.com/containernetworking/cni/pkg/utils"
 	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netlink/nl"
 )
 
 const defaultBrName = "cni0"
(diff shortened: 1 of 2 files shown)


dshulyak:
@steveeJ what do you think about this approach? Is it something that can be asked from every cni plugin?

plugins/main/bridge/bridge.go
```diff
@@ -170,6 +171,26 @@ func calcGatewayIP(ipn *net.IPNet) net.IP {
...
+		if err != nil {
+			return nil
+		}
+		// TODO check for v6
+		addrs, err := netlink.AddrList(link, nl.FAMILY_V4)
+		if err != nil {
+			return nil
+		}
+		result.IP4 = &types.IPConfig{
+			IP:      addrs[0].IPNet,
```
bboreham: check `len(addrs)>0` before doing this

dshulyak:
@rosenhouse i will cover my change with tests, but what is spec change?

rosenhouse:
@dshulyak If I'm understanding you correctly, you are proposing a change to every CNI plugin, and therefore to the CNI spec: that GET should be idempotent and return information about already-configured interfaces.

jellonek:
@dshulyak [this spec](https://github.com/containernetworking/cni/blob/master/SPEC.md).

dcbw:
@containernetworking/cni-maintainers to me GET is more appropriate than an idempotent ADD.  Semantically ADD implies doing the operation again, which seems odd since it was already done.  I'd rather do GET.  It seems like whatever calls CNI should really know if the container is already running, since it has to pass things like the container name, the namespace path, etc, and in that case it just wants to grab the current network config, so GET is appropriate.

leodotcloud:
I agree with @dcbw that's it's really odd to get the same call to ADD twice. We should just return an error and the calling party should handle it. CNI plugins are invoked by container runtimes which are generally very sophisticated and have the ability to store state. I believe adding a GET is also really an overhead to the API. Other comments: https://github.com/containernetworking/cni/issues/281#issuecomment-255271661

bboreham:
Although not fully decided on whether to add `GET` or not, we all seem agreed that `GET` would be preferable to this idempotent-`ADD` proposal, so I'm closing this one.

PR #287, opened 2016-08-22
Plugins report a list of supported versions

rosenhouse:
Further progress on versioning support (#266).

With this patchset:
- Every CNI plugin must opt in to reporting which CNI spec versions it supports
- Existing plugins may be re-compiled and report `version.Legacy` without any further code changes by the plugin author.
- The `libcni` API provides the `GetVersionInfo()` function which reports version information about a given plugin.  For older plugins which do not know the `VERSION` command, it reports support for version 0.1.0 only.

More drastic changes to the behavior of the runtime library (`libcni`) and plugin library (`skel`) is out of scope for this changeset.  Those will be added in follow-on PRs.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -64,8 +64,14 @@ The operations that the CNI plugin needs to support are:
 
 - Report version
   - Parameters: NONE.
-  - Result:
-    - The version of the CNI spec implemented by the plugin: `{ "cniVersion": "0.2.0" }`
+  - Result: information about the CNI spec versions supported by the plugin
+
+      ```
+      {
+        "cniVersion": "0.3.0", // the version of the CNI spec in use for this output
+        "supportedVersions": [ "0.1.0", "0.2.0", "0.3.0" ] // the list of CNI spec versions that this plugin supports
+      }
+      ```
 
 The executable command-line API uses the type of network (see [Network Configuration](#network-configuration) below) as the name of the executable to invoke.
 It will then look for this executable in a list of predefined directories. Once found, it will invoke the executable using the following environment variables for argument passing:
@@ -85,7 +91,7 @@ Success is indicated by a return code of zero and the following JSON printed to 
 
 ```
 {
-  "cniVersion": "0.2.0",
+  "cniVersion": "0.3.0",
   "ip4": {
     "ip": <ipv4-and-subnet-in-CIDR>,
     "gateway": <ipv4-of-the-gateway>,  (optional)
@@ -114,7 +120,7 @@ Examples include generating an `/etc/resolv.conf` file to be injected into the c
 Errors are indicated by a non-zero return code and the following JSON being printed to stdout:
 ```
 {
-  "cniVersion": "0.2.0",
+  "cniVersion": "0.3.0",
   "code": <numeric-error-code>,
   "msg": <short-error-message>,
   "details": <long-error-message> (optional)
@@ -151,7 +157,7 @@ Plugins may define additional fields that they accept and may generate an error 
 
 ```json
 {
-  "cniVersion": "0.2.0",
+  "cniVersion": "0.3.0",
   "name": "dbnet",
   "type": "bridge",
   // type (plugin) specific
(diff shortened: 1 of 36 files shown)


SPEC.md
```diff
@@ -85,7 +91,7 @@ Success is indicated by a return code of zero and the following JSON printed to
 
 ```
 {
-  "cniVersion": "0.2.0",
+  "cniVersion": "0.3.0",
```
dcbw: Are we bumping the CNI spec version to 0.3?  Do we need to do that, given we haven't had a CNI release yet with version 0.2?

pkg/version/version.go
```diff
@@ -16,27 +16,67 @@ package version
...
+	// Encode writes this CNI version information as JSON to the given Writer
 	Encode(io.Writer) error
 }
 
-// BasicVersioner is a PluginVersioner which reports a single cniVersion string
-type BasicVersioner struct {
-	CNIVersion string `json:"cniVersion"`
+type simple struct {
+	CNIVersion_        string   `json:"cniVersion"`
+	SupportedVersions_ []string `json:"supportedVersions,omitempty"`
```
dcbw: Curious about the _ suffix; is this a convention somewhere?  I probably didn't google for the right thing though, so it came up empty.

pkg/version/version.go
```diff
@@ -16,27 +16,67 @@ package version
...
+			return PluginSupports("0.1.0", "0.2.0"), nil
+		}
+		return nil, fmt.Errorf("decoding version info: missing field supportedVersions")
+	}
+	return &info, nil
 }
 
-// DefaultPluginVersioner reports the Current library spec version as the cniVersion
-var DefaultPluginVersioner = &BasicVersioner{CNIVersion: Current()}
+var Legacy = PluginSupports("0.1.0", "0.2.0", "0.3.0")
```
dcbw: We should document what Legacy contains, and are we going to expand it for future spec versions?  When should a plugin use Legacy?

pkg/version/version.go
rosenhouse: Not a conversion, AFAIK, just a hack so that we can have both public fields and functions with the same names.  The fields need to be public so that they get properly marshaled into JSON.  The function names need to match the public interface.

I'm very open to alternative naming schemes.

SPEC.md
rosenhouse: Good questions.

The main reason I changed it is because I was concerned that plugin authors and runtimes may be building against the `master` branch, and might be surprised to see changes in behavior without corresponding spec version bumps.  Bumping is low-cost and errs on the side of over-communicative, which I like.

Conversely, if we don't want plugin authors to be building against the `master` branch, then I'd say we need to be doing numbered releases much more often.

And if we are going to tie these together, it may make sense to actually require the numbers to be equal.  But we have already have a `v0.3` tag.  So perhaps we should immediately bump all version references on master to `0.4.0-rc`?

I know this topic was raised at our last maintainers meeting.  I can't recall if we came to a decision.  Other @containernetworking/cni-maintainers have thoughts?

rosenhouse:
@containernetworking/cni-maintainers I've revised and expanded this PR.  Please take a look.

Once these changes are in place, we can begin to add more interesting & useful versioning features to the runtime and plugin library packages.

SPEC.md
dcbw: If we care about people building from master, what's the point of having numbered releases then?  Our releases are currently match our Spec version, so we would either have to uncouple these two.  Plus we'd have to bump the Spec version every time a PR made an incompatible change.  We couldn't bundle up multiple spec changes spread across a couple commits into one real release, which will lead to Spec version inflation.

rosenhouse:
@dcbw @containernetworking/cni-maintainers In the interest of getting this merged, I've reverted the spec version bump - so it will stay at 0.2.0 for now.

I would like to continue the conversation about our release management process and the relationship between spec version and github release version.  But it need not block this feature.

dcbw:
@rosenhouse LGTM. Though I wish there was a way to disable the tests that check out and build stuff, since some build systems (especially package-based ones like most Linux distros use) don't always allow network access while building the package, to ensure reproducible builds from a given tarball.  But that's probably minority case...

rosenhouse:
@dcbw fwiw, the `git clone` operation shouldn't require network access -- it uses the local copy (where the test is running from) as the git "remote". 

Anyway, thanks for the review, merging...

PR #290, opened 2016-08-25
Fixed the check for network namespace path.

### LINKED ISSUE #288: CNI bridge plugin fails to read a bind mounted network namespace in CentOS 7
We are running Mesos on a CentOS7 with Kernel 3.10

The following check:
https://github.com/containernetworking/cni/blob/master/pkg/ns/ns.go#L104

seems to trip the CNI plugin. The `getNS` call expects a network namespace as a directory `ns/` while in Mesos we bind mount the process network namespace to a file. 

The plugins works fine on distros such as CoreOS and Ubuntu 14.04 which have a higher kernel version:

Why can't we check `/ns` instead of `ns/`? The check makes an explicit assumption that the network namespace will be directory which is what is causing this issue.

This is the error, we see from the CNI bridge plugin:
```
"msg": "failed to open netns \"/run/mesos/isolators/network/cni/d87e06fd-13b4-42ce-b767-5c0619247f51/ns\": path \"/run/mesos/isolators/network/cni/d87e06fd-13b4-42ce-b767-5c0619247f51/ns\" doesn't contain \"ns/\""
"code": 100
```

asridharan:
The expectation on older kernels (< 3.19) was to have the network
namespace always be a directory. This is not true if the network
namespace is bind mounted to a file, and will make the plugin fail
erroneously in such cases.

I have tried this with Mesos on a CentOS 7 cluster, running 3.10 kernel and it works fine. Moreover, have also tried it on Ubuntu 14.04 which has 3.13 kernel and works fine on that as well. 

This should fix #288

### DIFF
--- a/pkg/ns/ns.go
+++ b/pkg/ns/ns.go
@@ -20,7 +20,6 @@ import (
 	"os"
 	"path"
 	"runtime"
-	"strings"
 	"sync"
 	"syscall"
 
@@ -101,19 +100,7 @@ func IsNSorErr(nspath string) error {
 	}
 
 	switch stat.Type {
-	case PROCFS_MAGIC:
-		// Kernel < 3.19
-
-		validPathContent := "ns/"
-		validName := strings.Contains(nspath, validPathContent)
-		if !validName {
-			return NSPathNotNSErr{msg: fmt.Sprintf("path %q doesn't contain %q", nspath, validPathContent)}
-		}
-
-		return nil
-	case NSFS_MAGIC:
-		// Kernel >= 3.19
-
+	case PROCFS_MAGIC, NSFS_MAGIC:
 		return nil
 	default:
 		return NSPathNotNSErr{msg: fmt.Sprintf("unknown FS magic on %q: %x", nspath, stat.Type)}


jieyu:
I suggest we completely remove this check. As @asridharan pointed out, namespace handles can be bind mounted. Container runtime might choose to do the bind mount to keep the namespace open. This is a very common practice.

dcbw:
Seems OK to me... I think the main reason for the "ns" path check was to capture stuff like /proc/18768/task/18798/ns/net.

tomdee:
Yep, no objection from me either. @steveeJ any thoughts?

asridharan:
So the consensus is to remove the check completely under the `PROCFS_MAGIC` block? 
So I guess the point of the `switch stat.Type` check would be to only check that the file system type returned in PROCFS or NSFS ?

asridharan:
@tomdee @steveeJ thoughts on the new changes?

We were blocked on this for a DC/OS 1.8 release and would like to get these changes in, if the PR looks good.

Thanks a lot !!

rosenhouse:
LGTM, perhaps pending a change to the commit-message style so that it follows the [contribution guidelines](https://github.com/containernetworking/cni/blob/master/CONTRIBUTING.md#format-of-the-commit-message)?

asridharan:
@rosenhouse updated the commit message to follow the contribution guidelines. For the `subsystem` field I used `packages`. If there is a pre-defined list of sub-system that we can pick from and this needs to be changed please advise.

dcbw:
@asridharan if you change the prefix to "pkg/ns" then I'll merge it.  Thanks!

dcbw:
@asridharan I don't think there's a predefined list, but a quick "git log pkg/ns" or for whatever module will show what previous commits have used.

asridharan:
Ah !! @dcbw thanks!! Just updated the commit message.

dcbw:
@steveeJ this seems reasonable to me for now; if we want to catch random /proc paths passed in I guess we should figure out what to do there and take bind-mounted paths into account too.  Merging for now...

PR #291, opened 2016-08-29
pkg/ip: Ensure that SetupVeth returns correct hostVeth

tomdee:
The veth is moved from the container NS to the host NS.
This is handled by the code that sets the link to UP but the wrong
hostVeth is returned to the calling code.


jellonek:
LGTM.

rosenhouse:
@tomdee: is it possible to cover this change with a test?

tomdee:
I've updated the tests to add an assertion that the index of the returned veths is correct

PR #292, opened 2016-08-29
skip broadcast ip in host-local ipam 

freehan:
ref: https://github.com/kubernetes/kubernetes/issues/31630

@steveeJ @dcbw  We need this fix for k8s 1.4.

### DIFF
--- a/plugins/ipam/host-local/allocator.go
+++ b/plugins/ipam/host-local/allocator.go
@@ -114,8 +114,8 @@ func (a *IPAllocator) Get(id string) (*types.IPConfig, error) {
 
 	startIP, endIP := a.getSearchRange()
 	for cur := startIP; !cur.Equal(endIP); cur = a.nextIP(cur) {
-		// don't allocate gateway IP
-		if gw != nil && cur.Equal(gw) {
+		// don't allocate gateway IP or broadcast IP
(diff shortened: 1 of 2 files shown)


jellonek:
LGTM

bprashanth:
@steveeJ would be nice to get this into kube 1.4 which we're in the last stretch of debugging/testing for

rosenhouse:
👍

dcbw:
I would instead modify networkRange(), as this will ensure that the semantics conf.RangeEnd (eg, that it's inclusive) continue to be followed.  RangeEnd can certainly be set before the end of the subnet and thus before the broadcast address, eg start:192.168.1.1/24, end:192.168.1.10/24, and I think in this case your PR will disallow .10 from being allocated.

What I think you really want to do is simply disallow the end IP address if it wasn't specified in configuration, since that address will be the broadcast address.

But it also turns out that validation of RangeEnd is broken in host-local.  It's currently ignored for if Get(<requestedIP>) is passed, and it's ignored for getSearchRange() too when lastReservedIP is valid.  I'll work on a patch to fix that, and then we can base this PR on top of that one.  Sound OK?

freehan:
@dcbw SGTM. Thanks!

rosenhouse:
It looks like #294 superseded this one.  If I'm mistaken, feel free to rebase and re-open.

Regards.

PR #293, opened 2016-08-31
Backfill tests for plugin execution

rosenhouse:
Just adding tests, and introducing a bit of dependency injection.  Should make future refactors a little easier.


dcbw:
LGTM

PR #298, opened 2016-09-16
host-local: fix allocation of last address in a /30

freehan:
ref: https://github.com/kubernetes/kubernetes/issues/32844

### DIFF
--- a/plugins/ipam/host-local/allocator.go
+++ b/plugins/ipam/host-local/allocator.go
@@ -173,7 +173,7 @@ func (a *IPAllocator) Get(id string) (*types.IPConfig, error) {
 	}
 
 	startIP, endIP := a.getSearchRange()
-	for cur := startIP; !cur.Equal(endIP); cur = a.nextIP(cur) {
+	for cur := startIP; ; cur = a.nextIP(cur) {
 		// don't allocate gateway IP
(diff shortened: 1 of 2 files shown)


dcbw:
So this only happens with a /30 because the gateway IP gets skipped...  The existing loop logic is more complicated than I feel like it should be, but that's due to the wraparound behavior.  So the patch LGTM, but can you:

1. update the commit message to say this happens only with a /30 because the gateway IP gets skipped and in a /30 the next address is the end address
2. change the commit title to be prefixed with the plugin name, eg "host-local: fix allocation of last address in a /30"

Thanks!

freehan:
>The existing loop logic is more complicated than I feel like it should be

yeah, agreed

two causes:
1. the end address is not the broadcast address. 
2. the round robin search do not try the last reserve ip. Basically, it skips one ip in the cycle.

dcbw:
@freehan Sorry I didn't clarify, but I also meant update the actual commit name to be "host-local: fix allocation of last address in a /30", not just the PR.

freehan:
updated

dcbw:
@freehan seems the commit message got truncated to "3" not "30"...  I'll hit the merge button when thats fixed, thanks!

PR #301, opened 2016-10-03
A test verifies that libcni is backwards compatible with legacy plugins

rosenhouse:
Before we change the structures in the `types` package (e.g.. #145), I propose that we add tests to ensure that newer versions of `libcni` will continue to be able to parse the output from older plugins.

Right now, this new test passes trivially.  It will become valuable once our `types` start changing.

@dcbw @containernetworking/cni-maintainers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ env:
 
 matrix:
   fast_finish: true
-  allow_failures: 
+  allow_failures:
     - go: tip
   exclude:
     - go: 1.5.4
@@ -51,3 +51,6 @@ script:
 
 notifications:
   email: false
+
+git:
+  depth: 9999999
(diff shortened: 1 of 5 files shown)


steveeJ:
It's important to make sure that the referenced commits for the tests are always reachable within the travis runs. Otherwise new PRs could fail just for this reason without the PR author actually having introduced a bug.

dcbw:
LGTM; we'll probably want to solve the travis thing first though since we know we'll hit it.

tomdee:
So should this PR just add
```
git:
  depth: 9999999
```
To the travis YAML?

rosenhouse:
Thanks all, I've modified the travis file.  Should be good to go if someone else wants to click merge.

Issue #300, opened 2016-10-03
New IPAM plugin based on the IPGen Spec


rushmorem:
I've just [implemented] the [IPGen Spec] in `Go`. I would love to contribute and maintain an `ipam` plugin based on the new library. Would this be something you guys are interested in?

[implemented]: https://github.com/ipgen/go
[IPGen Spec]: https://github.com/ipgen/spec

rosenhouse:
Hi @rushmorem: That's an interesting idea for an `ipam` plugin.

One advantage of the CNI plugin model is that independent developers like yourself can build, distribute and use custom plugins without any changes to this repository or to the container runtimes that use it.

Perhaps you can start by building your CNI plugin & validating it for your use case.  Once you've got a repo with something functional, open a PR against our README to add it to our list of [3rd Party Network plugins](https://github.com/containernetworking/cni/blob/master/README.md#3rd-party-network-plugins).  (e.g. like #299)

I'm going to close this issue for now, but feel free to re-open this once you've got a working plugin that you think we ought to consider maintaining here.

rushmorem:
Thanks @rosenhouse. Let me do that.

PR #302, opened 2016-10-03
3rd party plugins belong in their own repos

rosenhouse:
Motivated by https://github.com/containernetworking/cni/issues/300#issuecomment-251032278 and https://github.com/containernetworking/cni/pull/259#issuecomment-234377787

@containernetworking/cni-maintainers : thoughts?


squaremo:
I agree with this policy. LGTM!

dcbw:
LGTM

PR #307, opened 2016-10-14
testing: test invocation of newer plugins with an older libcni

squeed:
This test compiles a trivial client with a fixed version libcni. It then creates and destroys a ptp network using the most recent plugin.

It's basically a simplified version of `ptp_test.go` in a single file.

### DIFF
--- a/libcni/api_test.go
+++ b/libcni/api_test.go
@@ -52,7 +52,7 @@ var _ = Describe("Invoking the plugin", func() {
 		}
 		Expect(debug.WriteDebug(debugFilePath)).To(Succeed())
 
-		cniBinPath = filepath.Dir(pathToPlugin)
+		cniBinPath = filepath.Dir(pluginPaths["noop"])
 		pluginConfig = `{ "type": "noop", "some-key": "some-value", "cniVersion": "0.2.0" }`
 		cniConfig = libcni.CNIConfig{Path: []string{cniBinPath}}
 		netConfig = &libcni.NetworkConfig{
--- a/libcni/backwards_compatibility_test.go
+++ b/libcni/backwards_compatibility_test.go
@@ -17,12 +17,15 @@ package libcni_test
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/containernetworking/cni/libcni"
 	"github.com/containernetworking/cni/pkg/version/legacy_examples"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
 )
 
 var _ = Describe("Backwards compatibility", func() {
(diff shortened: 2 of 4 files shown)


dcbw:
lgtm

pkg/version/legacy_examples/example_client.go
```diff
@@ -0,0 +1,131 @@
...
+		NetNS:       targetNs.Path(),
+		IfName:      ifName,
+	}
+
+	cniConfig := &libcni.CNIConfig{Path: strings.Split(os.Getenv("PATH"), ":")}
+
+	result, err := cniConfig.AddNetwork(netConf, runtimeConf)
+	fmt.Printf("Result: %+v", result)
+	if err != nil {
+		fmt.Printf("noop add failed: %+v", err)
```
rosenhouse: this isn't the no-op plugin, right?  maybe it could read `cni add failed` instead?

pkg/version/legacy_examples/example_client.go
```diff
@@ -0,0 +1,131 @@
...
+
+	ifName := "eth0"
+
+	runtimeConf := &libcni.RuntimeConf{
+		ContainerID: "some-container-id",
+		NetNS:       targetNs.Path(),
+		IfName:      ifName,
+	}
+
+	cniConfig := &libcni.CNIConfig{Path: strings.Split(os.Getenv("PATH"), ":")}
```
rosenhouse: This assumes that the `ptp` plugin is in the `PATH` when the tests are run.  That appears to be true of the top-level `./test` script, but it isn't generally true.  For example, if I just run 
```
go test github.com/containernetworking/cni/libcni
```
from my shell, then I get a failure with error 
```
failed to find plugin "ptp" in path [/usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin]
```

Perhaps you can build the `ptp` plugin outside of this program, e.g. as part of the new spec you're adding to `backwards_compatibility_test.go`.  Then pass in the path to that binary as an env var (e.g. `CNI_PATH`) or argument to this program.

That has the added benefit of making it clearer, when reading the test, that you're running a legacy runtime against a current-version plugin.  Because you can see the plugin gets built right there.

pkg/version/legacy_examples/example_client.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/containernetworking/cni/libcni"
+)
+
+func main(){
+	code :=	exec()
+	os.Exit(code)
+}
+
+func exec() int {
+	confStr := "{ \"name\": \"default\", \"type\":\"ptp\", \"ipam\":{ \"type\": \"host-local\", \"subnet\": \"10.1.2.0/24\"}}"
```
rosenhouse: This config is not part of the legacy runtime.  A user can change it independently of the runtime.

Therefore, I would prefer that this string be declared in the spec `It correctly handles the request from a legacy libcni` and somehow injected into this program (e.g. via env var or command line args).

That would make it clearer, when reading the test, what exactly is being tested.

rosenhouse:
Thanks @squeed, this is a great start, the right approach IMHO.  I left a few small suggestions to make this test more portable and readable.

pkg/version/legacy_examples/example_client.go
squeed: oops, good catch.

squeed:
@rosenhouse thanks for the feedback! Updated.

dcbw:
@squeed needs a fix for gopath/src/github.com/containernetworking/cni/pkg/version/legacy_examples/example_runtime.go:27: syntax error: non-declaration statement outside function body

squeed:
@dcbw sorry 'bout that. Off to buy a copy of vim for dummies.

PR #308, opened 2016-10-17
skel tests: correct name of CNI_CONTAINERID in tests of env vars

rosenhouse:
Fixes an oversight in the tests of the `skel` package


dcbw:
LGTM; clearly correct.

Issue #306, opened 2016-10-13
flannel cni plugin issues


chenchun:
**Kubernetes version** (use `kubectl version`):

```
Client Version: version.Info{Major:"1", Minor:"3+", GitVersion:"v1.3.6-dirty", GitCommit:"ae4550cc9c89a593bcda6678df201db1b208133b", GitTreeState:"dirty", BuildDate:"2016-09-02T04:15:50Z", GoVersion:"go1.6.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"3+", GitVersion:"v1.3.6-dirty", GitCommit:"ae4550cc9c89a593bcda6678df201db1b208133b", GitTreeState:"dirty", BuildDate:"2016-09-02T03:56:32Z", GoVersion:"go1.6.2", Compiler:"gc", Platform:"linux/amd64"}
```

**Environment**:
cni 0.3.0

**What happened**:
I was playing with cni flannel plugin, I changed flannel configure and restarted flannel on kubelet. Then I submitted a deployment, but the pods failed to start.

```
[root@kubernetes-master vagrant]# etcdctl --endpoint=http://10.245.1.2:4379 get /coreos.com/network/config
{
    "Network": "10.246.0.0/16",
    "SubnetLen": 24,
    "Backend": {
        "Type": "udp",
        "Port": 8285
     }
}

[root@kubernetes-master vagrant]# etcdctl --endpoint=http://10.245.1.2:4379 set /coreos.com/network/config '{"Network":"10.246.0.0/16","SubnetLen":24,"Backend":{"Type":"vxlan","Port":4789,"VNI":3}}'
{"Network":"10.246.0.0/16","SubnetLen":24,"Backend":{"Type":"vxlan","Port":4789,"VNI":3}}

[root@kubernetes-master vagrant]# kubectl describe pods
Name:       hello-2093073260-lk3f2
Namespace:  default
Node:       kubernetes-node-2/10.245.1.4
Start Time: Thu, 13 Oct 2016 02:56:21 +0000
Labels:     app=hello
        pod-template-hash=2093073260
Status:     Pending
IP:
Controllers:    ReplicaSet/hello-2093073260
Containers:
  hello:
    Container ID:
    Image:      chenchun/hello
    Image ID:
    Port:       80/TCP
    Requests:
      cpu:          100m
    State:          Waiting
      Reason:           ContainerCreating
    Ready:          False
    Restart Count:      0
[...]

PR #319, opened 2016-11-01
flannel: set forceAddress option of delegated bridge plugin to true by default

chenchun:
Yes, we can manually set `forceAddress` to true in flannel cni network config. But I think it's better to make this option be true by default because lots of people may not aware of this option.


chenchun:
ref https://github.com/containernetworking/cni/issues/306

squeed:
Closing; @chenchun do you want to re-submit for the plugins repository?

chenchun:
I'm quit on this due to the inactivity. Maybe it's a bad idea.

PR #321, opened 2016-11-02
dhcp: add support for DNS

### LINKED ISSUE #86: dhcp: add support for DNS
Option 6 is used to communicate the DNS server. The dhcp plugin should output this in "dns" field of the result JSON.

skysec:
Uses dhcp option 6 to populate result.DNS Nameservers,
and dhcp option 15 to populate result.DNS DNS Domain.
Added test cases.

Fixes #86


dcbw:
Generally LGTM, two requests: 1) add some length-checking to the domain name parsing, and 2) add testcases for the DNS server and domain name parsing where the values in the DHCP packet are too large or not divisible by 4.

skysec:
Thanks for your inputs. I've added a domain name validation, based on RFC 1035, max 255 bytes in total, and 63 per label, a couple of tests related to domain name string length and nameserver address (4 bytes boundary).

bboreham:
Is there an associated issue to make Kubernetes (or any other orchestrator) act on this info?

bboreham:
Why did you go with explicit byte slices like:

    []byte{101, 120, 97, 109, 112, 108, 101, 46, 99, 111, 109}

rather than casts like:

    []byte("example.com")

?

skysec:
I'm not aware of any requirement from a container orchestrator. However, there is an existing issue (#86),  also, I was working on a use case scenario, where a container needed to access the underlaying network using the parameters provided by a DHCP server on the Network, simulating a client in the network in  order to perform end-to-end service testing. On the other hand, besides IP Address/Mask and gateway, name server is the most common parameter provided by DHCP.

skysec:
Regarding byte arrays, there wasn't any specific reason to use an explicit byte slice. However, I understand your point regarding making the code and test more readable and easy to change.

plugins/ipam/dhcp/options.go
```diff
@@ -40,6 +42,45 @@ func classfulSubnet(sn net.IP) net.IPNet {
...
+
+func parseDNSDomain(opts dhcp4.Options) string {
+	// Parse domainname option
+	var domainname string
+	// Thanks to: https://github.com/asaskevich/govalidator
+	var DNSName string = `^([a-zA-Z0-9]{1}[a-zA-Z0-9_-]{1,62}){1}(\.[a-zA-Z0-9]{1}[a-zA-Z0-9_-]{1,62})*$`
+	if opt, ok := opts[dhcp4.OptionDomainName]; ok {
+		domainname = string(opt[:])
+		// RFC 1035 section 2.3.4 and 3.1
+		if len(strings.Replace(domainname, ".", "", -1)) > 255 {
```
squeed: minor nit: *technically* the limit here is 253 octets.

squeed:
Thanks for writing this.
It would be nice to parse the [domain search](https://tools.ietf.org/search/rfc3397) options as well (code 119), but it looks like the library we use doesn't support it.

Not only is the library missing the constant, but it also doesn't support RFC 3396 repeated options.
Are there any parsing libraries that do it correctly?

plugins/ipam/dhcp/options.go
skysec: Actually, it could be 254 (and after adding a single letter hostname, the result will be 255 bytes), which means that the regex is wrong (my bad), the domain "a.com" is valid, but the regex will fail validating this name (extra test on this, including dash, which are only accepted after a letter or digit).

skysec:
I didn't add the domain search option for the same reason: the DHCP library does not support this option. It's nice to have, however, it requires not only the addition of this extra constant, but handling the domain name "compression" algorithm used for DNS labels, using offsets to point to previous labels. The DNS library "https://github.com/miekg/dns", provides a function (UnpackDomainName) that already handles this algorithm. Regarding the encoding of long options (over 255 bytes), I don't know of any library that provides this functionality.

dcbw:
@skysec want to update the regex and repush?  Otherwise looks good, bonus points to make the byte arrays more readable.

@bboreham we've been talking about making Kubernetes mak euse of the DNS info for a while, but it's a bit complicated there because kube will construct the DNS info from a couple different places and we'd have to make sure that the existing behavior was preserved so that existing setups don't break.  That was what stopped from trivially implementing this already...

skysec:
Hi @dcbw, I updated the regex and test cases (more readable). For the regex, the validation is based on RFC 2181. relaxing the restrictions on the domain name.

squeed:
This LGTM as well. I agree that implementing the search domain isn't worth it.

squeed:
It looks like this is almost ready to go. Let's try and get this merged in 0.6.

rosenhouse:
This LGTM, but the plugins are now in [another repo](https://github.com/containernetworking/plugins)  🍄 

Would you mind squashing the commits and opening a new PR there?

PR #322, opened 2016-11-03
Allow flannel CNI plugin stateDir to be configurable

jaydunk:
We'd like to be able to configure where the flannel cni plugin stores its state. This is currently hard coded to `/var/lib/cni/flannel`. This PR adds an optional `stateDir` to the flannel NetConf, and if this is not present it defaults to `/var/lib/cni/flannel`.

### DIFF
--- a/plugins/meta/flannel/flannel_suite_test.go
+++ b/plugins/meta/flannel/flannel_suite_test.go
@@ -11,7 +11,7 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-package main_test
+package main
 
 import (
 	"encoding/json"
--- a/plugins/meta/flannel/flannel_suite_test.go
+++ b/plugins/meta/flannel/flannel_suite_test.go
@@ -1,0 +1,66 @@
+// Copyright 2015 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+package main_test
+
+import (
+	"encoding/json"
+	"fmt"
+	"path/filepath"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
+
+	"testing"
+)
+
+func TestFlannel(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Flannel Suite")
+}
+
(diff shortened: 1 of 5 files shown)


markstgodard:
Updated the PR to include an integration test suite that covers 
- flannel plugin delegates to noop plugin
- verifies that when passing `storeDir` option into NetConf for ADD and DEL allow overriding the default location of `/var/lib/cni/flannel`

This PR assumes that if you would want to override the default state dir, it must be passed in ADD and DEL.

Note: looks like the coveralls test fails since I added a new `TESTABLE` package to test script and it is an integration test vs. unit test.  I can look at backfilling unit tests as well if this blocks the PR

markstgodard:
Backfilled more unit tests to ensure coveralls passes  cc: @rosenhouse

rosenhouse:
LGTM, thanks for backfilling all that test coverage on a previously untested plugin!

Any other @containernetworking/cni-maintainers want to give the 👍  and merge?

squeed:
Can you update Documentation/flannel.md? Thanks!

markstgodard:
@squeed Sure, will update

markstgodard:
@squeed I updated flannel docs to include optional `stateDir`

Cheers

rosenhouse:
@dcbw mind looking at this one too?  similar to #327

plugins/meta/flannel/flannel_suite_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+var paths testPaths
+
+type testPaths struct {
+	PathToPlugin string
+	CNIPath      string
+}
+
+var _ = SynchronizedBeforeSuite(func() []byte {
+	noopBin, err := gexec.Build(noopPackage)
```
dcbw: I don't think we really need to build any of the plugins just to test them, take a look at pkg/testutils and how the other plugins do e2e testing with CmdAddWithResult() and CmdDelWithResult().  We should certainly extend those to capture other output though.

markstgodard: @dcbw Thanks for the feedback, I updated the PR to use `pkg/testutils` as recommended,  to test cmdAdd/Del for the stateDir test.

Cheers

markstgodard:
@dcbw Thanks for the feedback. I updated the flannel test to use the `pkg/testutils` CmdAdd/DelWithResult as suggested.

Cheers

dcbw:
@rosenhouse @squeed should we make some attempt to standardize the config option name?  I don't really care what it's called, DataDir or StateDir, and I"m happy to do a PR to change host-local to the consistent one.

dcbw:
Otherwise LGTM, but lets hold off on a merge a day or two until we figure out if/what a consistent statedir option would look like.

squeed:
Hm. While the "workdirs" in flannel vs. IPAM have pretty different uses, that doesn't seem like a good enough reason to use different names. Anyone else have an opinion?

rosenhouse:
At this risk of [bikeshedding](https://en.wiktionary.org/wiki/bikeshedding), I'd prefer `datadir` or `statedir` (no preference on casing).  Because those communicate that the data might need to stick around for a while.  `workdir` makes me think temporary, which isn't right.

markstgodard:
@rosenhouse @dcbw updated PR, renamed `stateDir` to `dataDir`

Cheers

rosenhouse:
This now matches the host-local IPAM `dataDir` (https://github.com/containernetworking/cni/blob/master/plugins/ipam/host-local/config.go#L34)

So, LGTM.

PR #327, opened 2016-11-10
host-local: add DataDir IPAM conf option and use it for testcases

dcbw:
Add an e2e host-local plugin testcase, which requires being able
to pass the datadir into the plugin so we can erase it later.
We're not always guaranteed to have access to the default data
dir location, plus it should probably be configurable anyway.

### DIFF
--- a/plugins/ipam/host-local/backend/disk/backend.go
+++ b/plugins/ipam/host-local/backend/disk/backend.go
@@ -31,8 +31,11 @@ type Store struct {
 	dataDir string
 }
 
-func New(network string) (*Store, error) {
-	dir := filepath.Join(defaultDataDir, network)
+func New(network, dataDir string) (*Store, error) {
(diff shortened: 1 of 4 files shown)


dcbw:

rosenhouse:
Awesome, LGTM!

squeed:
LGTM, thanks!
Do we want to consider adding a CNI_RUN_DIR environment variable to the invocation spec? Outside of ipam, are there any other plugins that would benefit from this?

dcbw:
> Do we want to consider adding a CNI_RUN_DIR environment variable to the invocation spec? Outside of ipam, are there any other plugins that would benefit from this?

@squeed @rosenhouse yeah, some others could potentially use it I'm sure, like the flannel stuff.  Do you think I should kill DataDir and rework to use CNI_RUN_DIR instead?

PR #330, opened 2016-11-17
Expand description of goroutine+netns problems

bboreham:
Re https://github.com/containernetworking/cni/issues/262#issuecomment-261224809

Issue #262, opened 2016-06-30
After changing namespace, other goroutines may switch to that namespace


bboreham:
In this repo you have `NetNS.Do`, but it is only called when you need a "specific namespace".  Turns out you also need to call it when you need the root namespace.

See discussion at https://github.com/vishvananda/netns/issues/17,  details at https://github.com/docker/libnetwork/issues/1113 or https://github.com/weaveworks/weave/issues/2388#issuecomment-228365069

The problem is that the Go runtime can create a new OS thread while you have changed namespace, and that new thread inherits that changed namespace, so now any operation elsewhere in the program can suddenly switch to the other namespace as Go schedules goroutines onto the new thread.

Given the nature of CNI, that each process tends to exit quickly after doing its work, it's perhaps hard to hit the problem.  But still possible.

rosenhouse:
Yes, thanks for bringing this up.  We've struggled with this issue for a while now (e.g. https://github.com/containernetworking/cni/issues/183).

A `reexec` pattern (like docker) can work, though its not exactly elegant.  Other process-level granularity would be good.  I know on some other projects that are otherwise Go-based, they still use a small C process to do the namespace switching.

It might be some static analysis tools could be useful here.

For now, the only suggestion I can make is that CNI plugins should obey the following 3 rules:
1. be short-lived (as you said)
2. be single-threaded, single-goroutine
3. never re-enter `NetNS.Do()`

Note that these rules apply to an entire process -- its impossible to enforce them via library code alone.

If anyone sees violations of any of these rules in CNI plugins in this repo, please bring them to the attention of myself or other maintainers.

bboreham:
My apologies: I thought I had read through #183 and found it was a different issue, hence I filed a new one.  But I see that your comment does make the exact point.

I have come up with roughly the same rules, but they are probably unachievable in any nontrivial program, considering they must apply to all libraries called. Nonetheless, short of a fix in the Go runtime, loudly announcing them may be the best we can do.

bboreham:
This came up again today; I'm wondering if https://github.com/containernetworking/cni/tree/master/pkg/ns/README.md should be updated to explain that "All code dependent on a particular network namespace" really means "All code that uses networking or may call into a library that uses networking".

rosenhouse:
Sounds good to me.  PR?

bboreham:
Issue now open for this specific point in Go https://github.com/golang/go/issues/20676

bboreham:
This is fixed in Go 1.10

PR #331, opened 2016-11-17
pkg/ns README: expand on danger in ns switching in long-lived programs

rosenhouse:
cc: @bboreham I realized there was another key point missing in the readme: that long-lived Go processes cannot switch namespaces safely, because even new goroutines may share threads with old goroutines.

Follow up to #330 and #262

### DIFF
--- a/pkg/ns/README.md
+++ b/pkg/ns/README.md
@@ -9,7 +9,7 @@ Go provides the `runtime.LockOSThread()` function to ensure a specific goroutine
 For example, you cannot rely on the `ns.Set()` namespace being the current namespace after the `Set()` call unless you do two things.  First, the goroutine calling `Set()` must have previously called `LockOSThread()`.  Second, you must ensure `runtime.UnlockOSThread()` is not called somewhere in-between.  You also cannot rely on the initial network namespace remaining the current network namespace if any other code in your program switches namespaces, unless you have already called `LockOSThread()` in that goroutine.  Note that `LockOSThread()` prevents the Go scheduler from optimally scheduling goroutines for best performance, so `LockOSThread()` should only be used in small, isolated goroutines that release the lock quickly.
 
 ### Do() The Recommended Thing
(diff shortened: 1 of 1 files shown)


bboreham:
> (which calls Unshare from inside a new goroutine)

This isn't a necessary condition for trouble.  Any time the Go runtime creates a new OS thread while the namespace was changed, you can expect some other goroutine to switch to that thread some time later.

rosenhouse:
Good point, I've re-written this section to focus on that point.  That's the "checkmate" aspect of this.  It is just totally outside the control of the Go programmer.

dcbw:
@rosenhouse can you explain the issue a bit more?  I'm not quite seeing how a new goroutine that's assigned to the same OS thread as a Do() can execute and do anything until that Do() has reset the netns and exited...  I'm very likely missing something :)

bboreham:
@dcbw It's not about a new goroutine assigned to the same OS thread; it's about a new OS thread created during the Do(), that is subsequently used for any goroutine. There's loads more information if you follow the links from #262

rosenhouse:
@dcbw @bboreham I put together a rough [demo of the problem here](https://github.com/rosenhouse/ns-mess).

The key part is [here](https://github.com/rosenhouse/ns-mess/blob/4ab0e68b273dfbb0873fcd4346079473069ea3d2/main.go#L64-L82): spin up some goroutines on the main thread, and observe if their namespace is as expected.

Failures are frequent but random.

rosenhouse:
Looks like Weave this issue, again: https://www.weave.works/blog/linux-namespaces-and-go-don-t-mix

bboreham:
@rosenhouse the ordering was: first Weave hit this issue, then Martynas found the cause, then Bryan came over here to point out it would hit CNI the same way 🙂, then Martynas wrote a blog post, then it sat in a publication queue for months.

(Bryan sat next to Martynas at the time)

bboreham:
This PR LGTM; @dcbw are you satisfied now?

rosenhouse:
Closing in favor of https://github.com/containernetworking/plugins/pull/14

PR #339, opened 2016-12-02
api/tuning: implement 'inner' delegate plugins

dcbw:
Implements @squeed "inside-out" suggestion in https://gist.github.com/squeed/c69206c51f7d38d1cbb5211a3fde5d06.

@containernetworking/cni-maintainers @rosenhouse @squeed

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -138,6 +138,7 @@ In addition, stderr can be used for unstructured output such as logs.
 The network configuration is described in JSON form. The configuration can be stored on disk or generated from other sources by the container runtime. The following fields are well-known and have the following meaning:
 - `cniVersion` (string): [Semantic Version 2.0](http://semver.org) of CNI specification to which this configuration conforms.
 - `name` (string): Network name. This should be unique across all containers on the host (or other administrative domain).
+- `inner` (dictionary): defines a configuration sub-block (see the [Plugin Chaining and Delegation](#plugin-chaining-and-delegation) section).
 - `type` (string): Refers to the filename of the CNI plugin executable.
 - `args` (dictionary): Optional additional arguments provided by the container runtime. For example a dictionary of labels could be passed to CNI plugins by adding them to a labels field under `args`.
 - `ipMasq` (boolean): Optional (if supported by the plugin). Set up an IP masquerade on the host for this network. This is necessary if the host will act as a gateway to subnets that are not able to route to the IP assigned to the container.
@@ -153,6 +154,47 @@ The network configuration is described in JSON form. The configuration can be st
   - `options` (list of strings): list of options that can be passed to the resolver
 
 Plugins may define additional fields that they accept and may generate an error if called with unknown fields. The exception to this is the `args` field may be used to pass arbitrary data which may be ignored by plugins.
+
+### Plugin Chaining and Delegation
+
+Multiple plugins may be chained together to perform more complex network configuration. IPAM plugins already operate this way, with the outer plugin delegating to the IPAM plugin and receiving its IPAM result.
+The network configuration may contain an `inner` field, which itself is a complete CNI network configuration block (except for the `cniVersion` and `name` fields, which cannot be specified in the `inner` dictionary).
+When the outer plugin finds the `inner` field, it proceeds to execute the inner plugin, passing the entire `inner` configuration via stdin, merging the top-level `cniVersion` and `name` fields.
+The inner plugin then executes, including possibly running yet another inner plugin if the top-level `inner` dictionary contains a second `inner` dictionary, specifying a third plugin.
+Each plugin then receives the IPAM result of the called plugin, and can use that result and optionally manipulate it before returning the result to its caller.
+
+For example, suppose traffic shaping should be enabled for a container whose network is configured with the standard `bridge` plugin and `host-local` IPAM plugins.
+
+```json
+{
+  "cniVersion": "0.2.0",
+  "name": "chaining-example",
+  "type": "traffic-shaping",
+  "ingressbw": "20M",
+  "egressbw": "10M",
+  "inner": {
+    "type": "bridge",
+    "bridge": "cni0",
+    "ipam": {
+      "type": "host-local",
+      "subnet": "10.1.0.0/16",
+      "gateway": "10.1.0.1"
+    }
+  },
+  "dns": {
+    "nameservers": [ "10.1.0.1" ]
+  }
+}
+```
+
+The top-level configuration would specify the traffic shaping plugin as the `type` along with plugin-specific fields for shaping parameters.
+The top-level configuration would also have an `inner` dictionary which would contain the normal `bridge` plugin configuration.
+The top-level `inner` dictonary's `bridge` configuration would also have an `ipam` section specifying the `host-local` plugin and normal `host-local` IPAM options.
+The traffic shaping plugin would first delegate to the `bridge` plugin (specified by the `inner` dictionary's `type` field), and the `bridge` plugin would delegate to the `host-local` IPAM plugin.
+The IPAM plugin would return the IPAM result to the bridge plugin which configures the container's networking.
+The bridge plugin then returns the IPAM result to the traffic shaping plugin, which can now use the container's IP address and interface name to set traffic shaping parameters.
+The traffic shaping plugin then returns the IPAM result itself to the runtime.
+
 ### Example configurations
 
 ```json
--- a/pkg/invoke/delegate.go
+++ b/pkg/invoke/delegate.go
@@ -15,12 +15,51 @@
 package invoke
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
 
 	"github.com/containernetworking/cni/pkg/types"
 )
+
+func GetNextPlugin(netconf []byte) (string, []byte, error) {
+	conf := struct {
+		types.NetConf
+		Inner map[string]interface{} `json:"inner,omitempty"`
+	}{}
+
+	err := json.Unmarshal(netconf, &conf)
+	if err != nil {
+		return "", nil, fmt.Errorf("error decoding net config: %v", err)
+	}
+
+	if conf.Inner != nil {
+		tmp, ok := conf.Inner["type"]
+		if !ok {
+			return "", nil, fmt.Errorf("inner plugin type not specified")
+		}
+		innerType, ok := tmp.(string)
+		if !ok {
+			return "", nil, fmt.Errorf("inner plugin type not a string")
+		}
+
+		// Copy network name and CNI version to inner plugin
+		conf.Inner["name"] = conf.Name
+		conf.Inner["cniVersion"] = conf.CNIVersion
+		innerConf, err := json.MarshalIndent(conf.Inner, "", "    ")
+		if err != nil {
+			return "", nil, fmt.Errorf("error marshalling inner plugin config: %v", err)
+		}
+		return innerType, innerConf, nil
+	} else if conf.IPAM.Type != "" {
+		// IPAM plugin gets run with normal netconf
+		return conf.IPAM.Type, netconf, nil
+	}
+
+	// No delegate to run
+	return "", nil, nil
+}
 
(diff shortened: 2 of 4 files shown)


SPEC.md
```diff
@@ -153,6 +154,47 @@ The network configuration is described in JSON form. The configuration can be st
   - `options` (list of strings): list of options that can be passed to the resolver
 
 Plugins may define additional fields that they accept and may generate an error if called with unknown fields. The exception to this is the `args` field may be used to pass arbitrary data which may be ignored by plugins.
+
+### Plugin Chaining and Delegation
+
+Multiple plugins may be chained together to perform more complex network configuration. IPAM plugins already operate this way, with the outer plugin delegating to the IPAM plugin and receiving its IPAM result.
+The network configuration may contain an `inner` field, which itself is a complete CNI network configuration block (except for the `cniVersion` and `name` fields, which cannot be specified in the `inner` dictionary).
+When the outer plugin finds the `inner` field, it proceeds to execute the inner plugin, passing the entire `inner` configuration via stdin, merging the top-level `cniVersion` and `name` fields.
+The inner plugin then executes, including possibly running yet another inner plugin if the top-level `inner` dictionary contains a second `inner` dictionary, specifying a third plugin.
```
tgraf: The wording on the exact execution/delegation order could be a bit more exact.

pkg/invoke/delegate.go
```diff
@@ -15,13 +15,52 @@
...
+			return "", nil, fmt.Errorf("inner plugin type not specified")
+		}
+		innerType, ok := tmp.(string)
+		if !ok {
+			return "", nil, fmt.Errorf("inner plugin type not a string")
+		}
+
+		// Copy network name and CNI version to inner plugin
+		conf.Inner["name"] = conf.Name
+		conf.Inner["cniVersion"] = conf.CNIVersion
```
asridharan: We should be setting the `args` field for the inner conf, by copying the `args` field from the outer conf.

asridharan:
Has it been already decided that the CNI Spec is going to use the "inside-out" approach to chain plugins? 

I can imagine this being quite painful for plugins, since all plugins will need to implement functionality for copying fields such as `args` that need to be passed to the "inner" plugins. 

Were there any cons in the "ordered" proposal. The "ordered" proposal seems to be devoid of these complications?

dcbw:
> Has it been already decided that the CNI Spec is going to use the "inside-out" approach to chain plugins?

@asridharan no, it hasn't.  I just did the PR because it was a simple way to start the conversation, and it looks like it has worked :)  See some other ideas in @squeed suggestions at https://gist.github.com/squeed/c69206c51f7d38d1cbb5211a3fde5d06

asridharan:
Got it !! Thanks for this PR to get the discussion started.

dcbw:
> I can imagine this being quite painful for plugins, since all plugins will need to implement functionality for copying fields such as args that need to be passed to the "inner" plugins.

@asridharan this patch includes Go library functions that handle this for you, so all you need to do is pass the config your plugin received to the function, which will then execute the next plugin for you.  If you don't use Go, then yes, it would have to be done manually.

Note that plugins already have to chain internally for IPAM. This piggy-backs on that functionality and re-uses a lot of the same code.

> Were there any cons in the "ordered" proposal. The "ordered" proposal seems to be devoid of these complications?

The cons are:

1. requires a change in the CNI configuration format specification, including rewriting configuration files that already exist
2. requires changes to all plugins to handle the new format, OR:
3. requires all runtimes to implement (possibly through libcni helper functions) the configuration format manipulation to construct the final CNI configuration passed to the plugin
4. requires all runtimes to implement (possibly through libcni helper functions) the plugin chaining functionality and passing results from one plugin to another, and merging IPAM results from multiple plugins together somehow in a well-defined way.  eg, what if a subsequent plugin wants to change the results?  what if a subsequent plugin wants to clear a field that a previous plugin emitted?  what if a subsequent plugin is unaware of chaining?
5. requires changes to plugins anyway to accept some new argument that contains the Result structure of the previous plugin
6. doesn't allow plugins to both pre and post-process operations.  For example, what if a global sysctl should be set *before* performing IPAM (like, say "disable_ipv6") and one needs to be set after IPAM on a specific interface?  Ok, this can be worked around by specifying the tuning plugin twice, once early for the global and once after IPAM for the interface-specific sysctl.  But doing so is less flexible in any case.

Just a few thoughts on the ordered mechanism.

This isn't to say that this inside-out approach is the best, it clearly has some clarity issues and does require changes to all plugins.  But I think any chaining operation will require changes to a majority of plugins and runtimes anyway.

squeed:
Argh, sorry I've let the discussion falter. I've been completely swamped with the pre-Christmas rush. I have another thought that doesn't result in a spec change - just some scaffolding in libCNI. I'll try and write it up in the coming days.

dcbw:
Thinking about it more, I'm guilty of overexpecting from the CNI spec.  There's nothing in the spec about CNI config files, only JSON configuration.  It's just convention that runtimes which implement CNI store config in a single file, and that they execute these files one-by-one (depending on the runtime).

So yes, we could just add a section to the spec for the "ordered" proposal and declare a clear separation between what plugins themselves receive (the current CNI config JSON) and what runtimes receive (either current CNI JSON or an ordered list) and that if runtimes read an ordered list they must handle chaining plugins and converting that ordered list to the current CNI config JSON for each plugin.  We would then add a config key like "prevResult" to the config JSON that would contain the previous plugins Result structure, which the next plugin could manipulate and return as its own Result structure.  This chaining could be handled via libcni functions much like existing runtimes use libcni for executing plugins.

This doesn't address the pre/post issue or the requirement that plugins need to be changed.

If we don't want to change plugins, then we would have to specify clear rules for how a runtime treats results from each plugin in the chain and when it merges new results into previous results.  That seems more complicated than just requiring plugins to handle chaining if they support a new-enough version of the CNI spec.

@squeed in any case, how about some review on #145 too, so we can move the versioning debate forward?

asridharan:
> Note that plugins already have to chain internally for IPAM. This piggy-backs on that functionality and re-uses a lot of the same code.

Valid point. Though we in Mesos use C++ to write the CNI plugin https://github.com/apache/mesos/blob/master/docs/cni.md#a-port-mapper-plugin , I agree its doable.

My worry is that in the future if fields get added to the spec (such as `args`) which are optional,  the "in-and-out" mechanism will not be usable unless all libraries update support for passing these new fields to the "inner" plugins. So I have a feeling that this might restrict the flexibility of expanding the spec. 

That's why I prefer the "ordered" proposal since it allows for aggregating most of the changes into the container run-time since then the container run-time can peg its support against a specific version of the spec (which already does) and the changes can be more centralized.

That said, I have to agree that any of the proposals will require a change to the plugins. I think we need an implementation of the `ordered` proposal. Will see if I follow your lead and do something about it. Not a golang expert but will try to dabble.

SPEC.md
```diff
@@ -153,6 +154,47 @@ The network configuration is described in JSON form. The configuration can be st
...
+
+For example, suppose traffic shaping should be enabled for a container whose network is configured with the standard `bridge` plugin and `host-local` IPAM plugins.
+
+```json
+{
+  "cniVersion": "0.2.0",
+  "name": "chaining-example",
+  "type": "traffic-shaping",
+  "ingressbw": "20M",
+  "egressbw": "10M",
```
freehan: Say if there is a traffic-shaping plugin, does it make more sense to take ingress/egress bandwidth configs thru runtime configs instead of network configs? In k8s CNI network plugin, we do not manipulate net config once it is settled. How can we handle configs like this? What is the boundary between runtime configs and network configs?

SPEC.md
```diff
@@ -153,6 +154,47 @@ The network configuration is described in JSON form. The configuration can be st
...
+
+### Plugin Chaining and Delegation
+
+Multiple plugins may be chained together to perform more complex network configuration. IPAM plugins already operate this way, with the outer plugin delegating to the IPAM plugin and receiving its IPAM result.
+The network configuration may contain an `inner` field, which itself is a complete CNI network configuration block (except for the `cniVersion` and `name` fields, which cannot be specified in the `inner` dictionary).
+When the outer plugin finds the `inner` field, it proceeds to execute the inner plugin, passing the entire `inner` configuration via stdin, merging the top-level `cniVersion` and `name` fields.
+The inner plugin then executes, including possibly running yet another inner plugin if the top-level `inner` dictionary contains a second `inner` dictionary, specifying a third plugin.
+Each plugin then receives the IPAM result of the called plugin, and can use that result and optionally manipulate it before returning the result to its caller.
```
freehan: Only share the ipam result?

SPEC.md
```diff
@@ -153,6 +154,47 @@ The network configuration is described in JSON form. The configuration can be st
...
+    "ipam": {
+      "type": "host-local",
+      "subnet": "10.1.0.0/16",
+      "gateway": "10.1.0.1"
+    }
+  },
+  "dns": {
+    "nameservers": [ "10.1.0.1" ]
+  }
+}
```
freehan: Just another thought. Is it also possible to do something like this:
```
{
  "cniVersion": "0.3.0",
  "name": "chaining-example",
  "type": "hostport", 
  "port": 80,
  "targetPort": 8080
  "inner": {
    "type": "traffic-shaping",
    "ingressbw": "20M",
    "egressbw": "10M",
    "inner": {
      "type": "bridge",
      "bridge": "cni0",
      "ipam": {
        "type": "host-local",
        "subnet": "10.1.0.0/16",
        "gateway": "10.1.0.1"
      }
    },
    "dns": {
      "nameservers": [ "10.1.0.1" ]
    }
  }
}
```

dcbw:
> That said, I have to agree that any of the proposals will require a change to the plugins. I think we need an implementation of the ordered proposal. Will see if I can follow your lead and do something about it. Not a golang expert but will try to dabble.

@asridharan I'm working on an ordered implementation to compare and contrast to this one, stay tuned...

SPEC.md
dcbw: For k8s, @thockin wanted the CNI network configs to be as kubernetes agnostic as possible.  That meant that we have a mostly static CNI network config like what I list here, so that plugins don't have to know kubernetes is running the show.

What I proposed instead in the kube hostports issue was CNI "template" files that kubernetes would read and interpret, then replace certain fields at runtime so that when the CNI plugin was called, it had a complete static config on stdin.

The template approach would allow plugin writers to create kube-agnostic network plugin chains (well, except for pod name/namespace in CNI_ARGS), making on the template files container-runtime-specific.

dcbw:
See https://github.com/containernetworking/cni/pull/346 for an implementation of the 'ordered' case.

squeed:
@dcbw I thought a bit about what you said (re: asking too much of the spec) and added an additional case to the chaining discussion: [multi-call](https://gist.github.com/squeed/c69206c51f7d38d1cbb5211a3fde5d06#multi-call). The basic idea is to just punt on the discussion entirely, and let the runtime handle chaining.

This has the advantage of not requiring any spec changes, but the disadvantage of requiring the runtime to keep a list of plugins for teardown.

Thoughts?

SPEC.md
freehan: For runtime configs, I am referring to CNI runtime configs like CNI_NETNS, CNI_IFNAME.

I am fine with either the template approach or extending the CNI runtime configs to include something like CNI_INGRESS_BW. I can see one potential problem with the template approach in CRI world. CRI runtime shims will not deal with PodSpec directly. They deal with podSandboxConfig and not all fields are stored in shims.

thockin:
Hey all, I've been sort of ignoring this but @freehan and I just had an interesting discussion and I want to capture some notes.

The biggest note I have is that I think the inside-out model is wrong.  It binds us wrt possible future changes in execution model (e.g. if exec becomes "exec in container" or even daemons - there are advantages, my point ins not to debate that here :)

I really think this should be a serial model.  With that, I will write the rest of my notes on https://github.com/containernetworking/cni/pull/346

dcbw:
I think I'm going to close this one as #346 seems have more momentum.

PR #340, opened 2016-12-06
Fix README.md host-local IPAM range start/end typo

vzcambria:
The configuration example in host-local README.md used wrong syntax for rangeStart and rangeEnd.


dcbw:
LGTM, thanks!

PR #341, opened 2016-12-10
host-local: trim whitespace from container IDs and disk file contents

dcbw:
It doesn't seem like container IDs should really have whitespace or
newlines in them.  As a complete edge-case, manipulating the host-local
store's IP reservations with 'echo' puts a newline at the end, which
caused matching to fail in ReleaseByID().  Don't ask...

@containernetworking/cni-maintainers @squeed @steveeJ


dcbw:
@squeed does this seem reasonable?

rosenhouse:
LGTM

PR #342, opened 2016-12-12
pkg: ns: Fix GetNS() by assigning mounted flag

sboeuf:
In case you create a new netns, the handler returned has the mounted flag
assigned to true. If you want to get another handler with GetNS() because
you didn't store the first one, you won't have the mounted flag properly
assigned, meaning the Close() function won't work as expected because it
won't unmount the netns. This is a bug because the close is supposed to
remove the netns completely.

### DIFF
--- a/pkg/ns/ns.go
+++ b/pkg/ns/ns.go
@@ -17,9 +17,12 @@ package ns
 import (
 	"crypto/rand"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"runtime"
+	"strings"
 	"sync"
 	"syscall"
 
@@ -81,6 +84,8 @@ const (
 	// https://github.com/torvalds/linux/blob/master/include/uapi/linux/magic.h
 	NSFS_MAGIC   = 0x6e736673
 	PROCFS_MAGIC = 0x9fa0
+
+	NsRunDir = "/var/run/netns"
 )
 
 type NSPathNotExistErr struct{ msg string }
@@ -110,6 +115,28 @@ func IsNSorErr(nspath string) error {
 	}
 }
 
+// isMounted checks if the network namespace path provided is mounted
+func isMounted(nsPath string) (bool, error) {
+	// Handle the case where the path has been returned
+	// from getCurrentThreadNetNSPath(). In that case, it
+	// is not mounted, thus no need to check.
+	if !strings.Contains(nsPath, NsRunDir) {
+		return false, nil
+	}
+
+	mountsContent, err := ioutil.ReadFile("/proc/self/mounts")
+	if err != nil {
+		return false, err
+	}
+
+	_, nsFile := filepath.Split(nsPath)
+	if nsFile == "" {
+		return false, fmt.Errorf("failed to find netns filename\n")
+	}
+
+	return strings.Contains(string(mountsContent), nsFile), nil
+}
+
 // Returns an object representing the namespace referred to by @path
 func GetNS(nspath string) (NetNS, error) {
 	err := IsNSorErr(nspath)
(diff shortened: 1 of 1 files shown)


dcbw:
@sboeuf The current PR doesn't really work. You can use GetNS() to return unmounted namespaces, like that returned from GetCurrentNS().  Attempting to unmount /proc/self/ns/net isn't going to work well.

If you really can't keep the original NetNS object around, and I'm guessing you can't because this is a process restart and you're picking up the pieces of where you left off, then I'd rather see:

1. a new "takeoverMount bool" argument to GetNS() that checks if the thing is really a CNI ns mountpoint by using some of the logic from NewNS, and then sets mounted:true in the return
2. some new code doc on GetNS() that very explicitly states what takeoverMount is for, when to use it, when not to use it, and what happens when you use it wrong.

Does that sound like it would solve your issue?  At least then you could just pass "restored" into GetNS() and not have the out-of-band tracking logic.

sboeuf:
@dcbw I think you have good ideas here and I would be more than happy to rework my PR to get something which does not break anything.
But I have a question regarding the point 1, why couldn't we check all the time whether the ns we want to get is mounted or not ?
I mean you are suggesting to make this check only if we specify takeoverMount = true, but we could do that all the time, without adding this parameter to GetNS().

Let me know if I had a misunderstanding here :)

sboeuf:
@dcbw let me know what you think about this new commit.

sboeuf:
Okay here is my analysis about the problem we have. With this PR, we are now reallly removing a netns (that is close it + unmount it + remove the file) because as long as it is recognized as mounted, at the first Close(), it is unmounted and removed. But in case of your legacy test cases, you are calling AddNetwork() and then DelNetwork() later. For all your plugins, AddNetwork() defers the closing of netns and at the time DelNetwork() is called, there is no netns corresponding to the given path.
It is okay without my PR because you are actually on closing only the file but you are not doing more than that because at each GetNS() call, your "mounted" flag is not set (false by default).

1) Either, I would suggest that I could split Close() in two functions such as Close() and Remove(), so that we would not need to replace all the calls to Close() in different plugins, and we would just call Remove() at the time we know we would want to unmount and remove the netns.
2) Or, I can only add a GetNSMounted function (still checking the path is actually mounted), duplicating the behavior of GetNS() but adding mounted flag assigned to true in case it is mounted, and returning an error in case it is not. Indeed we expect this path to be mounted. This is the easiest way but I still think this is a bit implicit about what is done.

@sameo, @dcbw Let me know your thoughts about the more appropriate solution to this issue.

dcbw:
> I mean you are suggesting to make this check only if we specify takeoverMount = true, but we could do that all the time, without adding this parameter to GetNS().

@sboeuf sorry for the lag on this one.

I think the disconnect here is between CNI plugins themselves, and the runtime orchestrator that runs the CNI plugin.

CNI plugins themselves *almost never* want to create a new NetNS and almost never want to tear one down and unmount it.  The CNI plugins make heavy use of GetNS() and they would never want to set takeoverMount=true.

But the *orchestrator* that runs the CNI plugin does potentially want to set takeoverMount=true, because it manages the lifetime of the NetNS and does create and tear it down.

So we cannot automatically check in GetNS() whether to take over the mount, because many things that don't want to tear the mount down still call GetNS().  Does that make things a bit clearer?  Sorry if I didn't explain my thoughts better before.

> BTW, where is the logic checking it is a CNI ns mountpoint in NewNS() ?

There isn't any, because pkg/ns currently assumes that the NS object will live for the entire time the namespace does.  That's somewhat of an oversight, I admit.

Inside the CNI code, NewNS() is only ever used in testcases.  The plugins themselves always use GetNS() because the namespace path is always created by the thing that calls the CNI plugin, and CNI assumes that thing will manage the namespace.  Often pkg/ns gets vendored into those runtimes, like Kubernetes or rkt, and these are often long-lived processes that can hold the namespace reference and close it when the container exits.

But you're correct that there's currently no easy provision to reclaim the namespace if the container orchestrator restarts while pods are alive.

dcbw:
> @dcbw let me know what you think about this new commit.

As described above, I don't think we can do this, because the plugins themselves make heavy use of GetNS() and thus, with this automatic checking, would attempt (and often fail) to tear down the namespace when they call Close() to release their reference.

This also wouldn't work for an orchestrator (or any other utility) that wants to manipulate the namespace without attempting to tear it down.

dcbw:
> Either, I would suggest that I could split Close() in two functions such as Close() and Remove(), so that we would not need to replace all the calls to Close() in different plugins, and we would just call Remove() at the time we know we would want to unmount and remove the netns.

@sboeuf I thought about this for a while, and this solution could work.  The only reservation I have is that I'd intended NetNS to be refcounted originally, so that it could be used from multiple places in the same program without having to track whether that place was the last user.  Having Close() and Remove() would break that, but NetNS doesn't currently implement refcounting yet.

So I'll check with other CNI maintainers and if they agree refcounting seems useful, I will add it and then Remove() would not be a great idea.  But if they do not agree that refcounting is useful, then I am OK with Remove().

sboeuf:
@dcbw thanks for your answers. Regarding your last comment, please let me know about other maintainers feedback and we will work on a proper solution :)

dcbw:
@containernetworking/cni-maintainers thoughts on the question posed in https://github.com/containernetworking/cni/pull/342#issuecomment-273814455 ?

squeed:
You can't "refcount," because who knows who else is using the namespace out-of-process. I suppose we could add an "unmount" command.

Note, of course, that this doesn't actually delete the namespace, it just removes the permanent reference to it. The namespace exists until the last process leave sit.

Closing due to repo split; feel free to reopen.

sboeuf:
I have just opened https://github.com/containernetworking/plugins/issues/18
Thanks!

PR #345, opened 2016-12-16
travis: roll forward the versions of Go that we test

rosenhouse:
- Start testing against Go 1.7.x
- No longer test against Go 1.5.x

According to the Go language [Release Policy](https://golang.org/doc/devel/release.html#policy):

> Each major Go release obsoletes and ends support for the previous one. 
> ...
> As a special case, we issue minor revisions for critical security problems in both the current release and the previous release. 

Therefore, Go 1.5.x hasn't been supported for security fixes since last August.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ sudo: required
 dist: trusty
 
 go:
-  - 1.5.4
-  - 1.6.2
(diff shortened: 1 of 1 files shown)


tomdee:
LGTM

Issue #309, opened 2016-10-17
Should plugin DEL ever error?


rosenhouse:
Issue #306 raises an interesting question that applies generally to all plugins: how should runtimes react if an `ADD` fails?

One possibility is for `DEL` to be best-effort idempotent.  It tries to clean-up any & all left-over resources and swallows any errors that might arise if a resource was never created in the first-place.

Then we could recommend that a runtime call `DEL` after a failed `ADD`.

dcbw:
Note that skel.go needs to be changed here too, specifically:
```
                {
                        "CNI_NETNS",
                        &netns,
                        reqForCmdEntry{
                                "ADD": true,
                                "DEL": false,
                        },
                },
```

Since we should still try to clean things up even when the netns has gone away.

bboreham:
We could have `CNI_NETNS==""` mean "clean up what you can, but there's no namespace so don't worry about that".

dcbw:
Though we could just do this now by making the internal plugins do best-effort cleanup, without changing the spec...  not sure if we'd consider that a material spec change or not.

bboreham:
The phrasing "tries to clean-up any & all left-over resources" - is that meant to mean "... relating to the specific container ID supplied", or more broadly, clean up anything that seems to be left-over?  The latter sounds dangerous.

I think I'm ok with best-efforts meaning "I looked for resources assigned to that container ID; didn't find anything; returning success"; I would still like to error on a completely invalid parameter.

dcbw:
@bboreham yes, I would expect that the operation would still be scoped to the specific container ID.

chenchun:
> add this as a spec requirement, not just a recommendation

+1 of adding this as a spec requirement, and we could still keep the error return of `DEL`.
Make `DEL` return nothing means it can't complain any mistakes during cleanup.

jieyu:
+1 on making this a spec requirement.

It's very hard for container runtime to react if cleanup fails. It has no idea if it should proceed with the rest of the cleanup or not.

bboreham:
The spec was changed in #346 to say "Plugins should generally complete a DEL action without error even if some resources are missing [...] even if the container network namespace no longer exists"

So I think this should be closed.

rosenhouse:
Agreed.  There is outstanding work to make the plugins actually behave this way.

bboreham:
[...]

PR #346, opened 2016-12-17
api/plugins: implement ordered plugin chaining

dcbw:
Implements @squeed "ordered" suggestion in https://gist.github.com/squeed/c69206c51f7d38d1cbb5211a3fde5d06.

Shows how both IPAM and regular plugins need to be modified to support prevResult and missing IPAM structures (since IPAM could also be in a standalone config block).

One missing piece is versioning of course, which needs to be bumped for this. I don't think we yet need to do everything #145 does since this doesn't change structures in a backwards compatible way.

@containernetworking/cni-maintainers @rosenhouse @squeed @freehan @asridharan @rajatchopra @caseydavenport

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -216,6 +216,167 @@ Plugins may define additional fields that they accept and may generate an error 
 }
 ```
 
+### Network Configuration Lists
+
+Network configuration lists provide a mechanism to run multiple CNI plugins for a single container in a defined order, passing the result of each plugin to the next plugin.
+The list is composed of well-known fields and list of one or more standard CNI network configurations (see above).
+
+The list is described in JSON form, and can be stored on disk or generated from other sources by the container runtime. The following fields are well-known and have the following meaning:
+- `cniVersion` (string): [Semantic Version 2.0](http://semver.org) of CNI specification to which this configuration list and all the individual configurations conform.
+- `name` (string): Network name. This should be unique across all containers on the host (or other administrative domain).
+- `plugins` (list): A list of standard CNI network configuration dictionaries (see above).
+
+When executing a plugin list, the runtime MUST replace the `name` and `cniVersion` fields in each individual network configuration in the list with the `name` and `cniVersion` field of the list itself.
+This ensures that the name and CNI version is the same for all plugin executions in the list, preventing versioning conflicts between plugins.
+
+For the ADD action, the runtime MUST also add a `prevResult` field to the configuration JSON of any plugin after the first one, which MUST be the Result of the previous plugin (if any) in JSON format ([see below](#network-configuration-list-runtime-examples)).
+For the ADD action, plugins SHOULD echo the contents of the `prevResult` field to their stdout to allow subsequent plugins (and the runtime) to receive the result, unless they wish to modify or suppress a previous result.
+Plugins are allowed to modify or suppress all or part of a `prevResult`.
+However, plugins that support a version of the CNI specification that includes the `prevResult` field MUST handle `prevResult` by either passing it through, modifying it, or suppressing it explicitly.
+It is a violation of this specification to be unaware of the `prevResult` field.
+
+The runtime MUST also execute each plugin in the list with the same environment.
+
+For the DEL action, the runtime MUST execute the plugins in reverse-order.
+
+#### Network Configuration List Error Handling
+
+When an error occurs while executing an action on a plugin list (eg, either ADD or DEL) the runtime MUST stop execution of the list.
+
+If an ADD action fails, when the runtime decides to handle the failure it should execute the DEL action (in reverse order from the ADD as specified above) for all plugins in the list, even if some were not called during the ADD action.
+
+Plugins should generally complete a DEL action without error even if some resources are missing.  For example, an IPAM plugin should generally release an IP allocation and return success even if the container network namespace no longer exists, unless that network namespace is critical for IPAM management. While DHCP may usually send a 'release' message on the container network interface, since DHCP leases have a lifetime this release action would not be considered critical and no error should be returned. For another example, the `bridge` plugin should delegate the DEL action to the IPAM plugin and clean up its own resources (if present) even if the container network namespace and/or container network interface no longer exist.
+
+#### Example network configuration lists
+
+```json
+{
+  "cniVersion": "0.2.0",
+  "name": "dbnet",
+  "plugins": [
+    {
+      "type": "bridge",
+      // type (plugin) specific
+      "bridge": "cni0",
+      // args may be ignored by plugins
+      "args": {
+        "labels" : {
+            "appVersion" : "1.0"
+        }
+      },
+      "ipam": {
+        "type": "host-local",
+        // ipam specific
+        "subnet": "10.1.0.0/16",
+        "gateway": "10.1.0.1"
+      },
+      "dns": {
+        "nameservers": [ "10.1.0.1" ]
+      }
+    },
+    {
+      "type": "tuning",
+      "sysctl": {
+        "net.core.somaxconn": "500"
+      }
+    }
+  ]
+}
+```
+
+#### Network configuration list runtime examples
+
+Given the network configuration list JSON [shown above](#example-network-configuration-lists) the container runtime would perform the following steps for the ADD action.
+Note that the runtime adds the `cniVersion` and `name` fields from configuration list to the configuration JSON passed to each plugin, to ensure consistent versioning and names for all plugins in the list.
+
+1) first call the `bridge` plugin with the following JSON:
+
+```json
+{
+  "cniVersion": "0.2.0",
+  "name": "dbnet",
+  "type": "bridge",
+  "bridge": "cni0",
+  "args": {
+    "labels" : {
+        "appVersion" : "1.0"
+    }
+  },
+  "ipam": {
+    "type": "host-local",
+    // ipam specific
+    "subnet": "10.1.0.0/16",
+    "gateway": "10.1.0.1"
+  },
+  "dns": {
+    "nameservers": [ "10.1.0.1" ]
+  }
+}
+```
+
+2) next call the `tuning` plugin with the following JSON, including the `prevResult` field containing the JSON response from the `bridge` plugin:
+
+```json
+{
+  "cniVersion": "0.2.0",
+  "name": "dbnet",
+  "type": "tuning",
+  "sysctl": {
+    "net.core.somaxconn": "500"
+  },
+  "prevResult": {
+    "ip4": {
+      "ip": "10.1.0.3/16",
+      "gateway": "10.1.0.1",
+    },
+    "dns": {
+      "nameservers": [ "10.1.0.1" ]
+    }
+  }
+}
+```
+
+Given the same network configuration JSON list, the container runtime would perform the following steps for the DEL action.
+Note that no `prevResult` field is required as the DEL action does not return any result.
+Also note that plugins are executed in reverse order from the ADD action.
+
+1) first call the `tuning` plugin with the following JSON:
+
+```json
+{
+  "cniVersion": "0.2.0",
+  "name": "dbnet",
+  "type": "tuning",
+  "sysctl": {
+    "net.core.somaxconn": "500"
+  }
+}
+```
+
+2) next call the `bridge` plugin with the following JSON:
+
+```json
+{
+  "cniVersion": "0.2.0",
+  "name": "dbnet",
+  "type": "bridge",
+  "bridge": "cni0",
+  "args": {
+    "labels" : {
+        "appVersion" : "1.0"
+    }
+  },
+  "ipam": {
+    "type": "host-local",
+    // ipam specific
+    "subnet": "10.1.0.0/16",
+    "gateway": "10.1.0.1"
+  },
+  "dns": {
+    "nameservers": [ "10.1.0.1" ]
+  }
+}
+```
 
 ### IP Allocation
 
--- a/plugins/test/noop/main.go
+++ b/plugins/test/noop/main.go
@@ -37,13 +37,14 @@ import (
 
 type NetConf struct {
 	types.NetConf
-	DebugFile string `json:"debugFile"`
+	DebugFile  string        `json:"debugFile"`
+	PrevResult *types.Result `json:"prevResult,omitempty"`
 }
 
 func loadConf(bytes []byte) (*NetConf, error) {
 	n := &NetConf{}
 	if err := json.Unmarshal(bytes, n); err != nil {
-		return nil, fmt.Errorf("failed to load netconf: %v", err)
+		return nil, fmt.Errorf("failed to load netconf: %v %q", err, string(bytes))
 	}
 	return n, nil
 }
@@ -66,15 +67,15 @@ func parseExtraArgs(args string) (map[string]string, error) {
 	return m, nil
 }
 
-func getDebugFilePath(stdinData []byte, args string) (string, error) {
+func getConfig(stdinData []byte, args string) (string, *NetConf, error) {
 	netConf, err := loadConf(stdinData)
 	if err != nil {
-		return "", err
+		return "", nil, err
 	}
 
 	extraArgs, err := parseExtraArgs(args)
 	if err != nil {
-		return "", err
+		return "", nil, err
 	}
 
 	debugFilePath, ok := extraArgs["DEBUG"]
@@ -82,11 +83,11 @@ func getDebugFilePath(stdinData []byte, args string) (string, error) {
 		debugFilePath = netConf.DebugFile
 	}
 
-	return debugFilePath, nil
+	return debugFilePath, netConf, nil
 }
 
 func debugBehavior(args *skel.CmdArgs, command string) error {
-	debugFilePath, err := getDebugFilePath(args.StdinData, args.Args)
+	debugFilePath, netConf, err := getConfig(args.StdinData, args.Args)
 	if err != nil {
 		return err
 	}
@@ -118,6 +119,15 @@ func debugBehavior(args *skel.CmdArgs, command string) error {
 
 	if debug.ReportError != "" {
 		return errors.New(debug.ReportError)
+	} else if debug.ReportResult == "PASSTHROUGH" || debug.ReportResult == "INJECT-DNS" {
+		if debug.ReportResult == "INJECT-DNS" {
+			netConf.PrevResult.DNS.Nameservers = []string{"1.2.3.4"}
+		}
+		newResult, err := json.Marshal(netConf.PrevResult)
+		if err != nil {
+			return fmt.Errorf("failed to marshal new result: %v", err)
+		}
+		os.Stdout.WriteString(string(newResult))
 	} else {
 		os.Stdout.WriteString(debug.ReportResult)
 	}
@@ -132,7 +142,7 @@ func debugGetSupportedVersions(stdinData []byte) []string {
 		return vers
 	}
 
-	debugFilePath, err := getDebugFilePath(stdinData, cniArgs)
+	debugFilePath, _, err := getConfig(stdinData, cniArgs)
 	if err != nil {
 		panic("test setup error: unable to get debug file path: " + err.Error())
 	}
--- a/plugins/test/noop/main.go
+++ b/plugins/test/noop/main.go
@@ -22,19 +22,38 @@ the JSON encoding of a Debug.
 package main
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strings"
 
 	"github.com/containernetworking/cni/pkg/skel"
+	"github.com/containernetworking/cni/pkg/types"
 	"github.com/containernetworking/cni/pkg/version"
 	noop_debug "github.com/containernetworking/cni/plugins/test/noop/debug"
 )
 
+type NetConf struct {
+	types.NetConf
+	DebugFile string `json:"debugFile"`
+}
+
+func loadConf(bytes []byte) (*NetConf, error) {
+	n := &NetConf{}
+	if err := json.Unmarshal(bytes, n); err != nil {
+		return nil, fmt.Errorf("failed to load netconf: %v", err)
+	}
+	return n, nil
+}
+
 // parse extra args i.e. FOO=BAR;ABC=123
 func parseExtraArgs(args string) (map[string]string, error) {
 	m := make(map[string]string)
+	if len(args) == 0 {
+		return m, nil
+	}
 
 	items := strings.Split(args, ";")
 	for _, item := range items {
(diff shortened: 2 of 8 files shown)


dcbw:
One question I had was ordering; should DEL order be the opposite of ADD order?  IPAM needs to happen before the 'bridge' plugin for ADD for example, so that 'bridge' can consume the IPAM result.

But on DEL we don't want to reverse that, eg we don't want to DEL 'bridge' first and the DEL IPAM, because plugins like DHCP may want the container interface to send a release on.  This PR executes plugins top to bottom for both ADD and DEL.  But I'm interested in use-cases where that doesn't work. One example could be tuning for interface-specific parameters like /proc/sys/net/ipv4/conf/wlp4s0/XXX where in the top-bottom case the interface would already be torn down by the 'bridge' plugin so tuning would error.  Maybe we need to adjust plugins to silently ignore errors on teardown when interfaces don't exist?

asridharan:
@dcbw good question. One use-case that I can think for reversing the ordering is for a port-mapping plugin. For port-mapping we want to install the iptables rule only after the brdige and IPAM plugin are run, however during DEL it would be better to remove the iptables rule before removing the bridge and freeing the IPAM, else it might happen that if you release the IP address before removing the iptables entry, and the IP address gets allocated before the entry is removed, traffic might get directed to a wrong container.

dcbw:
@asridharan one option might be a top-level key called "del-order" in which you could specify "bottom-up" or "reverse" or something.  Alternatively, we could a separate (optional) "del-plugins" key, which would also be a list of network configs.  That way you could have different add/del ordering.

asridharan:
@dcbw I do agree that having a field to specify the order of addition and deletion would be necessary. This is one advantage of having the configs defined as an ordered structure vs using the "inside-out" approach.

rajatchopra:
However in the 'inner' approach there is a possibility of pre and post processing. Can't we have both?

bboreham:
It seems intuitive to me that DEL should do everything in the opposite order to ADD.

However

> because plugins like DHCP may want the container interface to send a release on

I can't see how the interface could be a requirement to release an address but not to claim an address.
Most likely this is because I don't know enough about DCHP; would you mind explaining?

asridharan:
@bboreham I think @dcbw is making the same point as you. The reverse order is required because we need the interface to exist before you do a DHCPREQUEST or DHCPRELEASE?

bboreham:
@asridharan what you just said is the opposite of https://github.com/containernetworking/cni/pull/346#issuecomment-267744039

asridharan:
Ah !! @bboreham I see your point. I didn't pay attention to @dcbw comments on ADD.

asridharan:
@dcbw the example you sighted in [#346 (comment)](https://github.com/containernetworking/cni/pull/346#issuecomment-267744039), i.e., keeping the order the same for ADD and DEL would work only for host-local IPAM, right? As @bboreham pointed out I think you would need to reverse the order for DHCP IPAM?

dcbw:
@asridharan @bboreham yes, you're right.  Only IPAM plugins that do not require a container interface or network connectivity could be pulled out of the 'ipam' section and get their own section.  Since DHCP needs the interface, and since the main plugins (eg bridge, etc) set the IP configuration on the interface, the main plugin needs to delegate to DHCP itself and process the result immediately rather than having a separate section for DHCP.

So yes, perhaps I should just remove the whole thing about top-level IPAM plugins since not all of them can really be pulled out?  We can always pull them out later too.

dcbw:
> However in the 'inner' approach there is a possibility of pre and post processing. Can't we have both?

@rajatchopra I don't see how we could using the 'ordered' approach, except by specifying the same plugin twice in the list, once before and once after, like I show for the 'tuning' plugin.  That is one downside of the ordered approach.

rajatchopra:
@dcbw Yes, that is the downside. I meant to suggest a hybrid approach where one could specify 'inner' as well as an ordered list. One day we might have to introduce pre/post keywords. Too unwieldy for now I guess.

jieyu:
From a container runtime's perspective, I am more curious about the error handling part. For instance, during the destroy path, what if the first plugin's DEL failed? Should the container runtime call 'DEL' for the subsequent plugins in the chain? If not, how can we make sure things get properly cleaned up?

As a container runtime writer, i feel that it's very hard for a container runtime to deal with cleanup failures because container runtime has no idea if it should proceed with the rest of the cleanup or not. Maybe we should add something to the spec that 'DEL' for each plugin should be best effort. As long as the plugin is sure that the assets related to the plugin do not exist anymore, it should return a success so that container runtime can proceed and do the rest of the cleanup.

Thoughts?

freehan:
@jieyu Good point. Need to set the expectation. +1 to add it into spec

Ref: https://github.com/containernetworking/cni/issues/309

thockin:
@bboreham re DHCP - the "up" process involves broadcasting on the interface and getting an IP assigned.  Presumably a "down" would also need the interface, though you could argue that it could just be left to expire.  I don't know who, if anyone would do refresh, so maybe expiring is a terrible idea.

thockin:
Hey all, I've been sort of ignoring this but @freehan and I just had an interesting discussion and I want to capture some notes.

The biggest note I have is that I think the inside-out model is wrong. It binds us wrt possible future changes in execution model (e.g. if exec becomes "exec in container" or even daemons - there are advantages, my point is not to debate that here :)

So, without having read this PR I basically reinvented this proposal, even as far as extracting IPAM to be the first plugin in sequence.  The DHCP complication is interesting and warrants more thought.  Maybe the order for DHCP needs to be (interface, IPAM, port-forward) rather than (IPAM, interface, port-forward)?

What I described differently was how port-forwarding would work.  The port numbers are akin to the IP address - they don't belong in the `NetworkConfig`.  All the network config needs is a record indicating "use the iptables_port_forward plugin".  The actual ports should be passed in as part of the args.

Now we can debate whether port-forwarding is actually part of CNI (e.g. `CNI_PORT_FORWARD=80:8080,443:8443`) or whether it is an "optional extension" (e.g. `CNI_EXT_PORT_FORWARD=80:8080,443:8443`) or whether it is a kube-specific things (e.g. `CNI_ARGS=K8S_PORT_FORWARD=80:8080,443:8443`).  I really want to decouple, so I'll fight for NOT the last choice.  The followup is what happens if the user asked for ports to be forwarded and no plugin actually forwarded them - ignore or fail?

We can repeat the debate for traffic shaping.

So TL;DR I think this sounds close to what we want, or at least @freehan and I independently derived the same basic design :)

bboreham:
@thockin that is the same point I was making; I believe we all subsequently agreed.

thockin:
yeah, I replied before reading ahead :)

asridharan:
@thockin 
> What I described differently was how port-forwarding would work. The port numbers are akin to the IP address - they don't belong in the NetworkConfig. All the network config needs is a record indicating "use the iptables_port_forward plugin". The actual ports should be passed in as part of the args.

We actually implemented a CNI port-mapper plugin for Mesos using the above logic. 
http://mesos.apache.org/documentation/latest/cni/#a-port-mapper-plugin

Though it uses the delegate model, but it could have been implemented using this "ordered" proposal as well.

tomdee:
The CNI maintainers just discussed this and agreed that

1. DEL should call the plugins from bottom to top (i.e. in reverse order)
2. We'll maintain that IPAM plugins are wrapped

asridharan:
@tomdee @dcbw what about cases where just reverse the order of plugins during DEL would not work. I thought we had sighted such instances and the idea was to introduce a field to explicitly specify the order for DEL. Any specific reason why we are not allowing the operator to specify the order here?

dcbw:
> @tomdee @dcbw what about cases where just reversing the order of plugins during DEL would not work. I thought we had sighted such instances and the idea was to introduce a field to explicitly specify the order for DEL. Any specific reason why we are not allowing the operator to specify the order here?

@asridharan we looked over the comments and couldn't find examples other than potentially DHCP.  Given that we are going to punt top-level IPAM until a later commit, the DHCP issue isn't a problem.

My comment https://github.com/containernetworking/cni/pull/346#issuecomment-267744039 was predicated on having IPAM be pulled-out, and since that's not the case yet, that reason for configurability doesn't stand anymore.

For your previous comment "For port-mapping we want to install the iptables rule only after the brdige and IPAM plugin are run, however during DEL it would be better to remove the iptables rule before removing the bridge and freeing the IPAM, else it might happen that if you release the IP address before removing the iptables entry, and the IP address gets allocated before the entry is removed, traffic might get directed to a wrong container." I believe the current proposal addresses that.  Since for bottom-to-top, if you have the iptables rule plugin last in the config list, then it would be called last for ADD and first for DEL, ensuring the iptables rule only gets added *after* everything is configured, and would be removed *before* anything is deconfigured.

dcbw:
@containernetworking/cni-maintainers I have pushed a version that removes the top-level IPAM blocks and implements the reverse-ordering on DEL.  PTAL thanks!

plugins/test/noop/main.go
```diff
@@ -123,7 +155,36 @@ func cmdDel(args *skel.CmdArgs) error {
...
+	if _, err := w.Write(stdinData); err != nil {
+		return nil, err
+	}
+	if err := w.Close(); err != nil {
+		return nil, err
+	}
+
+	os.Stdin = r
+	return stdinData, nil
+}
```
rosenhouse: What about:
```go
os.Stdin = bytes.NewReader(stdinData)
```
I think that can substitute for all the `os.Pipe` stuff.

asridharan:
@dcbw you are right current proposal meets the need for the port-mapping plugin. Should have phrased my question better. I was actually wondering given that we have seen one instance of an "ordered" sequence of plugins (IPAM in this case) requiring a user-defined ordering, couldn't there be other combination of plugins for which we might not be able to predict the behavior on DEL. 

But I guess from your response it does seem like the maintainers have given this enough thought. Also, thinking about this a bit more, not introducing "del-plugins" at this point keeps things simple and  in case we do encounter such combinations we can always think about introducing the "del-plugins" field at that point.

tomdee:
Let's move the discussion on passing portmapping info to plugins to #351

SPEC.md
```diff
@@ -211,6 +211,65 @@ Plugins may define additional fields that they accept and may generate an error
...
+- `cniVersion` (string): [Semantic Version 2.0](http://semver.org) of CNI specification to which this configuration list and all the individual configurations conform.
+- `name` (string): Network name. This should be unique across all containers on the host (or other administrative domain).
+- `plugins` (list): A list of standard CNI network configuration dictionaries (see above).
+
+When executing a plugin list, the runtime MUST replace the `name` and `cniVersion` fields in each individual network configuration in the list with the `name` and `cniVersion` field of the list itself.
+This ensures that the name and CNI version is the same for all plugin executions in the list, preventing versioning conflicts between plugins.
+
+For the ADD action, the runtime MUST also add a `prevResult` field to the configuration JSON of any plugin after the first one, which MUST be the Result of the previous plugin (if any) in JSON format.
+For the ADD action, plugins SHOULD echo the contents of the `prevResult` field to their stdout to allow subsequent plugins (and the runtime) to receive the result, unless they wish to modify or suppress a previous result.
```
tomdee: How about making this a bit more concrete with an example? You could update the result section below to show an example which includes the prevResult.

tomdee:
@dcbw This is looking really good to me. I've just focused on reviewing the spec and have two comment, 1 is inline above and the other is: Does this bump the spec version? If so, you should update the versions in the spec.

SPEC.md
```diff
@@ -211,6 +211,65 @@ Plugins may define additional fields that they accept and may generate an error
 }
 ```
 
+### Network Configuration Lists
+
+Network configuration lists provide a mechanism to run multiple CNI plugins for a single container in a defined order, passing the result of each plugin to the next plugin.
```
thockin: In a world where Kubernetes has a Network object, it seems that object would translate to CNI plugins.  Do we expect the user who creates the Network block to be forced to specify both portmapping *and* 'real" driver?  That would force end-users to know about CNI.  Or do we expect them to only specify the real driver, forcing kubelet or dockershim to add the portmapping plugin?  How does kubelet know that the driver specified by th euser doesn't do it natively?

SPEC.md
```diff
@@ -211,6 +211,65 @@ Plugins may define additional fields that they accept and may generate an error
...
+
+The list is described in JSON form, and can be stored on disk or generated from other sources by the container runtime. The following fields are well-known and have the following meaning:
+- `cniVersion` (string): [Semantic Version 2.0](http://semver.org) of CNI specification to which this configuration list and all the individual configurations conform.
+- `name` (string): Network name. This should be unique across all containers on the host (or other administrative domain).
+- `plugins` (list): A list of standard CNI network configuration dictionaries (see above).
+
+When executing a plugin list, the runtime MUST replace the `name` and `cniVersion` fields in each individual network configuration in the list with the `name` and `cniVersion` field of the list itself.
+This ensures that the name and CNI version is the same for all plugin executions in the list, preventing versioning conflicts between plugins.
+
+For the ADD action, the runtime MUST also add a `prevResult` field to the configuration JSON of any plugin after the first one, which MUST be the Result of the previous plugin (if any) in JSON format.
```
asridharan: We should specify the behavior expected of the container run-time when an ADD operation fails for a given list of plugins. Should the container run-time invoke DEL on the plugins it has invoked ADD on in reverse order?

SPEC.md
```diff
@@ -211,6 +211,65 @@ Plugins may define additional fields that they accept and may generate an error
...
+For the ADD action, plugins SHOULD echo the contents of the `prevResult` field to their stdout to allow subsequent plugins (and the runtime) to receive the result, unless they wish to modify or suppress a previous result.
+Plugins are allowed to modify or suppress all or part of a `prevResult`.
+However, plugins that support a version of the CNI specification that includes the `prevResult` field MUST handle `prevResult` by either passing it through, modifying it, or suppressing it explicitly.
+It is a violation of this specification to be unaware of the `prevResult` field.
+
+The runtime MUST also execute each plugin in the list with the same environment.
+
+For the DEL action, the runtime MUST execute the plugins in reverse-order.
```
asridharan: We should specify behavior of container runtime in case of failure of DEL as well. If we see DEL failure on a given plugin should the container run-time just bail (possibly leaking resources) or should we continue with DEL operations on the remaining plugins? My opinion would be to bail and allow the operator to clean up these resources manually.

SPEC.md
asridharan: @thockin shouldn't this discussion be part of the #sig-network rather than part of the CNI spec?

thockin: Perhaps, but if this feature isn't useful for Kubernetes, I'd argue the feature isn't right.

asridharan: True. But if I understand correctly, what you are describing seems to be internal details of Kubernetes, in other words Kubernetes perspective of how a CNI network (here a combination of portmapping and real driver) is exposed to the user? I believe the spec should be independent of assumptions of the internal details of the container runtime. So still trying to understand your argument as why the internal specifics of Kubernetes should be influential in ratifying the spec?

thockin:
I guess I am asking for opinions - given this design, all the answers I see
are bad.  Is there an answer that doesn't lead to bad news for kube?

asridharan:
@thockin i c... Could you point to doc discribing Kuberentes network objects and how are they being used. Sorry, work on CNI support Mesos so not familiar with the Kubernetes code base.

SPEC.md
dcbw: > In a world where Kubernetes has a Network object, it seems that object would translate to CNI plugins.

@thockin I would expect that a world with Kube Network objects would map whatever the kube network object was to a single CNI plugin config file, or a single CNI plugin config list.  Much like it does now, except that kube just picks the first "network" in the directory.

So, say Kube Network objects have a name.  You have some networks (a) "pod-network" and (b) "storage-network", (c) "really-cool-nfv-stuff".  You also have three CNI network config lists named /etc/cni/networks/pod-network.conflist, /etc/cni/networks/storage-network.conflist, and /etc/cni/networks/really-cool-nfv-stuff.conflist.

Most pods are tagged with "pod-network".  Only some pods (like your database) are tagged with "storage-network".  You have some firewall-as-a-service tagged with only "really-cool-nfv-stuff".

The storage-network conflist doesn't include things like hostport, since the database isn't serving any clients on the storage network, it's just consuming storage.

The really-cool-nfv-stuff pod uses some custom network plugin stuff, and so its conflist only has one plugin in it and none of the standard CNI ones.

>  Do we expect the user who creates the Network block to be forced to specify both portmapping and 'real" driver? That would force end-users to know about CNI. Or do we expect them to only specify the real driver, forcing kubelet or dockershim to add the portmapping plugin?

I don't think we can get away from this.  We have two choices: (a) make Kube Network objects a representation of some other object defined outside kube, or (b) force plugin writers and backends to translate IPAM, hostport, interface name, whatever details from a highly-specified Kube network object into something they can actually run.

I really don't think we want Kube to start carrying full network specifications, because there's so much variability in how plugins operate and what configuration they need.  eg, I don't see kube ever creating docker networks or CNI config files on its own, because we'll never get a standard network config format that works for most plugin vendors, and even if we did it would always be less capable.

Maybe I'm setting up a strawman here, but I think Kube Network objects should probably only hold basic information and essentially "refer" to some other network object that the network plugin actually knows about.  For CNI that would be a config file name or a conflist name.  For Docker that could be a docker network.

>  How does kubelet know that the driver specified by th euser doesn't do it natively?

Kubelet probably can't.  Depending on the network plugin the cluster has chosen, or the network plugin specified in the Network object, the user may have to do some small configuration.  Kubelet can provide all the information in a standardized way, but asking network plugins to optionally consume that information has two problems.  First, it creates plugins that are kubernetes-specific, which historically you did not want.  Second, it requires some system of capabilities of network plugins to ensure that if the user wants hostports but hte plugin doesn't support it, that an error occurs even if the plugin doesn't recognize hostports.

dcbw:
@asridharan that doc doesn't yet exist.  I'm very curious what @thockin thinks about them though, since they appear quite different than mine...

squeed:
The intention behind my original proposal was that the runtime (kubelet, e.g.) could read a network config file from somewhere static, mutate it, and present it to CNI. That is the justification for such a trivial structure - to make appending multiple plugins programmatically feasible.

Obviously in order for DELETE to work, the runtime needs to be able to recreate the original config file - which is why we should probably add a restriction such that `args` must not be needed for DELETE.

@thockin were you intending to integrate PodCIDR in to the kubelet CNI client? That would also require a network file mutation. (Right now, it ignores it).

asridharan:
@squeed 
> Obviously in order for DELETE to work, the runtime needs to be able to recreate the original config file - which is why we should probably add a restriction such that args must not be needed for DELETE.
Could you elaborate a bit more why we need this restriction? If args is optional during DELETE, how does the container run time make decisions when it should and when should not pass args ?

thockin:
There is no such abstraction yet.

Kube has an abstraction that includes port-mapping *and* IP assignment
together.  CNI is decomposing them, which sounds good but I can't see how
to re-compose them without it being visible in places it shouldn't be.

> Maybe I'm setting up a strawman here, but I think Kube Network objects
should probably only
> hold basic information and essentially "refer" to some other network
object that the network
> plugin actually knows about.

Hmm, that could work.  It's a little more hands-off than I was imagining,
but it is clean...

dcbw:
> Hmm, that could work.  It's a little more hands-off than I was imagining,
> but it is clean...

@thockin we can always move in the "more hands on" direction in the future if there are common things that should be abstracted in Kube.  The other direction is somewhat harder because you're always going to be taking away something somebody was using.  I'm trying to do simple-first to reduce initial bike-shedding and get something in, then expand later if necessary.

thockin:
ACK.  I guess I can buy that, for now.

SPEC.md
rosenhouse: @asridharan Related (but not exactly the same) discussion in #309, but looking at the plugin-side of things.  Perhaps the recommendation is that plugin DELs should try not to exit with status 1, particularly if the resource to delete was missing.  But if the plugin does exit, the runtime respects that and does not continue the chain.

plugins/test/noop/main.go
dcbw: @rosenhouse I think the issue here is that we need an *os.File to assign back to stdin, which bytes.NewReader() doesn't give us...

SPEC.md
dcbw: @asridharan @rosenhouse I would generally agree, though it's good practice for the plugin to be flexible on DEL, like if the netns isn't around, still release the IPAM allocation instead of erroring out.  So +1 for the "bail in the middle".

SPEC.md
dcbw: @asridharan to keep things simplest, I would advocate two things:

1) plugins should not return errors on non-critical failures; eg if some resources they want aren't present (like the container netns) they shouldn't fail.  Absolutely critical failures can still be returned.
2) the runtime calls *all* plugins for DEL, regardless of which ones were ADD-ed.  Since plugins should be flexible on DEL, this shouldn't fail even if the plugin was not run.

This approach means the runtime doesn't have to store a specific list of plugins, and also means we don't have to do complex explanations in the spec that could be misunderstood.  I'm a big fan of making things simple and unambiguous rather than potentially convoluted.

Plus I think plugins should be resilient on DEL anyway...

dcbw:
@asridharan @tomdee @containernetworking/cni-maintainers updated the Spec to add examples of how runtimes should process conflists, and to clarify what happens for errors encountered during an ADD/DEL action.  PTAL thanks!

tomdee:
Spec changes LGTM

SPEC.md
```diff
@@ -216,6 +216,164 @@ Plugins may define additional fields that they accept and may generate an error
...
+
+#### Network Configuration List Error Handling
+
+When an error occurs during execution of a plugin list, the runtime MUST stop execution of the list.
+If an ADD action fails, the runtime should execute the DEL action (in reverse order from the ADD of course) for all plugins in the list, even if some were not called during the ADD action. Plugins should be flexible in clean up and generally succeed even though some resources are not present.  For example, an IPAM plugin should generally release an IP allocation and return success even if the container network namespace no longer exists, unless that network namespace is critical for IPAM management. While DHCP may usually send a 'release' message on the container network interface, since DHCP leases have a lifetime this release action would not be considered critical and no error should be returned. For another example, the `bridge` plugin should delegate the DEL action to the IPAM plugin and clean up its own resources (if present) even if the container network namespace and/or container network interface no longer exist.
```
asridharan: Should we add a statement being explicit about the behavior during DEL? That is bail in the middle when an error occurs during DEL processing.

asridharan: Modulo the statement on DEL, this LGTM !!

dcbw: @asridharan spec updated to clarify.

dcbw:
@squeed @steveeJ @rosenhouse any thoughts?  If it's good to go somebody hit the merge button :)

plugins/test/noop/main.go
rosenhouse: Ah, my mistake!

PR #348, opened 2017-01-02
Set RT_SCOPE_LINK when gateway is nil

lstoll:
There's no gateway, so keep the route scoped to the link. It'll only be relevant for it.

This seems like it would be consistent with iproute2's behaviour https://kernel.googlesource.com/pub/scm/linux/kernel/git/shemminger/iproute2/+/master/ip/iproute.c#1258 . I'm unsure of a case where this behaviour would be be wanted.

### DIFF
--- a/pkg/ip/route_test.go
+++ b/pkg/ip/route_test.go
@@ -1,0 +1,123 @@
+// Copyright 2016 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ip_test
+
+import (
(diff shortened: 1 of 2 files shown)


pkg/ip/route.go
```diff
@@ -28,9 +28,13 @@ func AddDefaultRoute(gw net.IP, dev netlink.Link) error {
 
 // AddRoute adds a universally-scoped route to a device.
```
rosenhouse: Mind updating the doc comment?

lstoll:
@rosenhouse updated and added a test. I'm not familiar with this testing framework, so feedback on my usage of it very welcome

squeed:
It looks like this PR fell off the radar. Would you mind resubmitting it under the new plugins repository?

pkg/ip/route_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/vishvananda/netlink"
+	"github.com/vishvananda/netlink/nl"
+)
+
+var _ = Describe("Link", func() {
+	const (
+		ifaceFormatString string = "i%d"
+		mtu               int    = 1400
+		ip4onehwaddr             = "0a:58:01:01:01:01"
```
rosenhouse: i don't think this const is being used.

rosenhouse: sorry for the delay, I didn't realize my github review was "pending" and wasn't public.

lstoll:
I'll freshen it up and open a new PR, cheers!

PR #355, opened 2017-01-23
skel: adds PluginMainWithError which returns a *types.Error

### LINKED ISSUE #354: skel.PluginMain calls os.Exit instead of returning an error, preventing graceful shutdown
CNI plugins that access state outside the immediate process (such as communicating with external processes, accessing shared resources, etc.) have to gracefully release those resources on exit. However, skel.PluginMain calls os.Exit directly on behalf of the plugin any time the ADD or DEL handler returns an error. This prevents the plugin from doing any graceful shutdown.

My understanding is the packages under cni.pkg are libraries, and libraries should not just kill the process in which they are loaded. They should return an error and leave it up to the caller to decide how to handle that error.

One could of course call a general cleanup function directly from inside the ADD and DEL handlers at every failure point, but that results in duplicate code and does not result in an elegant design.

What is the recommended way of doing graceful cleanup and shutdown?
Should we add an alternative PluginMain that returns an error instead of abruptly exiting?

rosenhouse:
Enables plugins to perform custom error-handling.

Fixes #354

### DIFF
--- a/pkg/skel/skel.go
+++ b/pkg/skel/skel.go
@@ -189,25 +189,40 @@ func (t *dispatcher) pluginMain(cmdAdd, cmdDel func(_ *CmdArgs) error, versionIn
 	return nil
 }
 
-// PluginMain is the "main" for a plugin. It accepts
-// two callback functions for add and del commands.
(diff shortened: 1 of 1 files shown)


squeed:
Mmm, makes sense. LGTM.

PR #359, opened 2017-01-28
libcni, pkg\invoke: Use OS-specific list separator when parsing CNI_PATH

### LINKED ISSUE #358: CNI uses the hard-coded ":" character as list separator
Hardcoding the list separator character as ":" causes CNI to fail when parsing CNI_PATH on other operating systems. For example, Windows uses ";" as list separator because ":" can legally appear in paths such as "C:\path\to\file".

Without this PR, a CNI_PATH of value "\path\to\config; c:\another\path" would be incorrectly parsed as "\path\to\config; c" and "\another\path" on Windows.

I am sending a PR that fixes various instances of this issue by making OS-agnostic calls, or replacing the ":" with os.PathListSeparator. This equals ":" on Linux and ";" on Windows, so there is no behavior change on Linux.

The Spec is also updated to refer to OS-agnostic behavior.

ofiliz:
Hardcoding the list separator character as ":" causes CNI to fail when parsing
CNI_PATH on other operating systems. For example, Windows uses ";" as list
separator because ":" can legally appear in paths such as "C:\path\to\file".

Fixes #358

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -88,7 +88,7 @@ It will then look for this executable in a list of predefined directories. Once 
 - `CNI_NETNS`: Path to network namespace file
 - `CNI_IFNAME`: Interface name to set up; plugin must honor this interface name or return an error
 - `CNI_ARGS`: Extra arguments passed in by the user at invocation time. Alphanumeric key-value pairs separated by semicolons; for example, "FOO=BAR;ABC=123"
-- `CNI_PATH`: Colon-separated list of paths to search for CNI plugin executables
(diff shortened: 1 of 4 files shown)


ofiliz:
@rosenhouse thanks! Agreed with the need on CI support for Windows builds. Until then I am running tests on Linux.

I will be adding new test cases with each new PR from now on (I just sent one). This change was just path parsing and existing tests in find_test.go already cover this scenario.

philips:
LGTM

pires:
@tomdee one more maintainer LGTM is needed, please.

tomdee:
:smile: LGTM. Merging.

PR #366, opened 2017-02-03
Do not error if last reserved not found after initial creation

ehazlett:
When using the `host-local` IPAM plugin, after initial network creation the following error is printed due to missing file:

```
2017/02/03 17:50:08 Error retriving last reserved ip: Failed to retrieve last reserved ip: open /var/lib/cni/networks/external/last_reserved_ip: no such file or directory
```

This fixes to not return an error when the file does not exist.

### DIFF
--- a/plugins/ipam/host-local/backend/allocator/allocator.go
+++ b/plugins/ipam/host-local/backend/allocator/allocator.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"log"
 	"net"
+	"os"
 
 	"github.com/containernetworking/cni/pkg/ip"
 	"github.com/containernetworking/cni/pkg/types"
@@ -253,7 +254,7 @@ func (a *IPAllocator) getSearchRange() (net.IP, net.IP) {
 	var endIP net.IP
 	startFromLastReservedIP := false
 	lastReservedIP, err := a.store.LastReservedIP()
-	if err != nil {
+	if err != nil && !os.IsNotExist(err) {
 		log.Printf("Error retriving last reserved ip: %v", err)
 	} else if lastReservedIP != nil {
 		subnet := net.IPNet{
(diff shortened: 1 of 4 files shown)


ehazlett:
The test failure seems like a non code fail:

```
```

squeed:
I'm not sure this is the right choice.
Rather, I think we should probably just return the unadorned error. That way the consumer can just call `os.IsNotExist` themselves.

ehazlett:
Ah yes it seems it would be better to handle it here (https://github.com/containernetworking/cni/blob/master/plugins/ipam/host-local/backend/allocator/allocator.go#L255) and just create the directory instead of letting it log and continue.  I'll update.

squeed:
Ah yes, that's what I meant. I didn't realize we were the consumer :-). Thanks.

dcbw:
@ehazlett were you able to update the fix yet?

ehazlett:
sorry for the delay!

updated -- thx!

squeed:
You'll still need to edit `disk/backend.go#88` to return the original error.

rosenhouse:
@ehazlett See above.  Also, could you add a unit test to cover this change?

ehazlett:
@rosenhouse I've updated.  When trying to add a test it doesn't appear to run (I added a failing case to `master` but the test appears to be ignored as everything passes).  I even tried to run the tests direct based upon the test script but I get `no tests to run`.   I'm not famliar with ginkgo so this will take me a bit to dig through.

ehazlett:
OK I think I found it.  I had to add a new test suite for `host-local`.  It appears none of those tests were running.

ehazlett:
@rosenhouse OK I believe I've finished updating and have enabled / fixed the remaining failing tests.

rosenhouse:
@ehazlett Awesome, thanks for getting this done!  (and amazing that none of those tests were running already).

👍 

other @containernetworking/cni-maintainers want to get this in before the release? (I'm tired of seeing this erroneous error message in my logs)

squeed:
How embarrassing! Anyways, LGTM!

ehazlett:
Thanks!

PR #369, opened 2017-02-08
CONVENTIONS.md: Update details on port-mappings

tomdee:
After discussion with the CNI maintainers and representatives from Mesos and Kubernetes we've settled on this approach for passing the dynamic port mapping information from runtimes to plugins.

Including it in the plugin config allows "operators" to signal that they expect port mapping information to be passed to a specific plugin (or plugins). This follows the model that config passed in the plugin config should be acted upon by a plugin.

Runtimes can optionally return an error to users if they find no plugin in the chain that supports port mappings.

I'm sure we'd all like to see this merged soon so I'll try to turn any feedback quickly (and I'm happy to get any and all feedback - spellings, unclear working, wrong JSON terminology etc..)

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -10,9 +10,11 @@ Establishing these conventions allows plugins to work across multiple runtimes. 
 
 ## Plugins
 * Plugin authors should aim to support these conventions where it makes sense for their plugin. This means they are more likely to "just work" with a wider range of runtimes.
+* Plugins should accept arguments according to these conventions if they implement the same basic functionality as other plugins. If plugins have shared functionality that isn't coverered by these conventions then a PR should be opened against this document.
 
 ## Runtimes
 * Runtime authors should follow these conventions if they want to pass additional information to plugins. This will allow the extra information to be consumed by the widest range of plugins.
+* These conventions serve as an abstraction for the runtime. For example, port forwarding is highly implementation specific, but users should be able to select the plugin of their choice without changing the runtime.
 
 # Current conventions
 Additional conventions can be created by creating PRs which modify this document.
@@ -23,7 +25,9 @@ Additional conventions can be created by creating PRs which modify this document
 
 A plugin can define any additional fields it needs to work properly. It is expected that it will return an error if it can't act on fields that were expected or where the field values were malformed.
 
-This method of passing information to a plugin is recommended when the information has specific meaning to the plugin that it's passed to and when the plugin should act on it or return an error if it can't.
+This method of passing information to a plugin is recommended when the following conditions hold
+* The configuration has specific meaning to the plugin (i.e. it's not just general meta data)
+* the plugin is expected to act on the configuration or return an error if it can't
 
 Dynamic information (i.e. data that a runtime fills out) should be placed in a `runtime_config` section.
 
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -29,14 +29,14 @@ Dynamic information (i.e. data that a runtime fills out) should be placed in a `
 
 | Area  | Purpose| Spec and Example | Runtime implementations | Plugin Implementations |
 | ------ | ------ | ------             | ------  | ------                  | ------                 |  
-| port mappings | Pass mapping from ports on the host to ports in the container network namespace. | Runtimes should receive the follow plugin config <pre>"runtime_config": {port_mappings": []} </pre> Runtimes should fill in the actual port mappings when the config is passed to plugins e.g. <pre>"runtime_config": {<br />  "port_mappings" : [<br />    { "host_port": 8080, "container_port": 80, "protocol": "tcp" },<br />    { "host_port": 8000, "container_port": 8001, "protocol": "udp" }<br />  ]<br />}</pre> | none | none |
+| port mappings | Pass mapping from ports on the host to ports in the container network namespace. | Operators can ask runtimes to pass port mapping information to plugins, by setting the following in the CNI config <pre>"capabilities": {port_mappings": true} </pre> Runtimes should fill in the actual port mappings when the config is passed to plugins. It should be placed in a new section of the config "runtime_config" e.g. <pre>"runtime_config": {<br />  "port_mappings" : [<br />    { "host_port": 8080, "container_port": 80, "protocol": "tcp" },<br />    { "host_port": 8000, "container_port": 8001, "protocol": "udp" }<br />  ]<br />}</pre> | none | none |
 
 For example, the configuration for a port mapping plugin might look like this to an operator (it should be included as part of a [network configuration list](https://github.com/containernetworking/cni/blob/master/SPEC.md#network-configuration-lists).
 ```json
 {
   "name" : "ExamplePlugin",
   "type" : "port-mapper",
-  "runtime_config": {"port_mappings": []}  
+  "capabilities": {"port_mappings": true}  
 }
 ```
 
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -21,13 +21,37 @@ Additional conventions can be created by creating PRs which modify this document
 [Plugin specific fields](https://github.com/containernetworking/cni/blob/master/SPEC.md#network-configuration) formed part of the original CNI spec and have been present since the initial release.
 > Plugins may define additional fields that they accept and may generate an error if called with unknown fields. The exception to this is the args field may be used to pass arbitrary data which may be ignored by plugins.
 
-A plugin can define any additional fields it needs to work properly. It is expected that it will return an error if if can't act on fields that were expected or where the field values were malformed.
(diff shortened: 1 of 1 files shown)


tomdee:
cc @containernetworking/cni-maintainers @thockin @freehan @jieyu @asridharan @philips

CONVENTIONS.md
```diff
@@ -21,13 +21,37 @@ Additional conventions can be created by creating PRs which modify this document
...
+
+Dynamic information (i.e. data that a runtime fills out) should be placed in a `runtime_config` section.
 
 | Area  | Purpose| Spec and Example | Runtime implementations | Plugin Implementations |
 | ------ | ------ | ------             | ------  | ------                  | ------                 |  
-| port mappings | Pass mapping from ports on the host to ports in the container network namespace. | <pre>"port_mappings" : [<br />  { "host_port": 8080, "container_port": 80, "protocol": "tcp" },<br />  { "host_port": 8000, "container_port": 8001, "protocol": "udp" }<br />] </pre> | none | none |
+| port mappings | Pass mapping from ports on the host to ports in the container network namespace. | Runtimes should receive the follow plugin config <pre>"runtime_config": {port_mappings": []} </pre> Runtimes should fill in the actual port mappings when the config is passed to plugins e.g. <pre>"runtime_config": {<br />  "port_mappings" : [<br />    { "host_port": 8080, "container_port": 80, "protocol": "tcp" },<br />    { "host_port": 8000, "container_port": 8001, "protocol": "udp" }<br />  ]<br />}</pre> | none | none |
```
asridharan: This sentence :
> Runtimes should receive the follow plugin config
seems a bit off.  We could be more explicit here. Maybe:
> Operators can ask runtimes to pass port mapping information to plugins, by setting the following in the CNI config ....

CONVENTIONS.md
```diff
@@ -21,13 +21,37 @@ Additional conventions can be created by creating PRs which modify this document
 [Plugin specific fields](https://github.com/containernetworking/cni/blob/master/SPEC.md#network-configuration) formed part of the original CNI spec and have been present since the initial release.
 > Plugins may define additional fields that they accept and may generate an error if called with unknown fields. The exception to this is the args field may be used to pass arbitrary data which may be ignored by plugins.
 
-A plugin can define any additional fields it needs to work properly. It is expected that it will return an error if if can't act on fields that were expected or where the field values were malformed.
+A plugin can define any additional fields it needs to work properly. It is expected that it will return an error if it can't act on fields that were expected or where the field values were malformed.
 
-This method of passing information to a plugin is recommended when the information should be acted on and has specific meaning to that plugin.
+This method of passing information to a plugin is recommended when the information has specific meaning to the plugin that it's passed to and when the plugin should act on it or return an error if it can't.
```
asridharan: This sentence seems a bit convoluted. Can we rephrase this?

asridharan: I believe the main point we want to put accross is that plugin specific fields are immutable by the container run time, and are supposed to be passed as is to the plugin. 

The following paragraph can then deal with mutable information that the runtime needs to add to the configuration during container launch.

freehan:
Port-Mapping is expecting a list. In the future, other fields may expect something else, such as a string, a boolen or a json blob. Those field probably need to use `{}` as a place holder? Maybe worth mentioning `[]` is for a list of objects, `{}` is for a single object. 

It is also worth mentioning if any of the runtime_config fields are not filled by runtime, plugins must not return error but only noop. Reasonable?

thockin:
This looks OK to me.  The only question I have is about how consumers will unmarshal this JSON.  Naive unmarshalling of an empty list is indistinguishable from a non-specified list in some languages.

Languages like Go distinguish a nil slice from an empty slice, but it is non-obvious.  It seems this design intends to avoid a second field indicating the intent, but is that too clever?

Concretely, if I unmarshal to a strict which has a []PortMapping, can I *really* get the right answer every time?  What happens when I marshal that struct - does it omit the slice (nil) or produce an empty slice?

Lastly, can we spec what plugins should do if they are passed this option but do not implement it?  Should they hard fail or just ignore?  What happens if a plugin doesn't implement portMap, but is passed an empty list of portMap?  Syntactically wrong, semantically right.  Fail?

jieyu:
+1 on the unmarshal comment. If one unmarshal this JSON to protobuf (Mesos does), protobuf cannot distinguish an empty list from non-specified list.

asridharan:
@tomdee @jieyu so looks like a better option would be to have `port_mappings` as a boolean? This feels a bit odd since this would be first instance where the CNI config on file and CNI config understood by the plugin (that supports port-mapping) will end having different representations.

asridharan:
@tomdee @jieyu @thockin a follow up thought. Looks like the current convention of `runtime_configs.port_mappings`  trying to achieve two things inform the container run time about the capability of a plugin, and give the container run time a place to fill run time configuration for that capability.

Can we actually split these two objectives? My thought was to introduce a top-level `capabilities` dict, which will be immutable, and the container runtime will fill in the run time configuration in a `runtime_configs` dict which is mutable by the container runtime. So for `port_mappings` we would have an immutable boolean field in the CNI config `capabilities.port_mappings = true|false` and the container runtime would fill up the corresponding field in `runtime_config` which could be `runtime_config.port_mappings=[...]` .

CONVENTIONS.md
tomdee: Yes, it's very convoluted! I think it wasn't clear before and maybe I've just made it worse now.

The goal of this paragraph is to give specific guidance on when this form of passing data to a plugin should be used. It aims to contrast this approach with the other two options. So the point I want to get across, is that this approach is used when the plugin is expected to act on the data. This is in contrast with the "args" approach where the data is much more loosely coupled and the runtime doesn't really know or care if the plugin acts on it.

CONVENTIONS.md
```diff
@@ -29,14 +29,14 @@ Dynamic information (i.e. data that a runtime fills out) should be placed in a `
...
+| port mappings | Pass mapping from ports on the host to ports in the container network namespace. | Operators can ask runtimes to pass port mapping information to plugins, by setting the following in the CNI config <pre>"capabilities": {port_mappings": true} </pre> Runtimes should fill in the actual port mappings when the config is passed to plugins. It should be placed in a new section of the config "runtime_config" e.g. <pre>"runtime_config": {<br />  "port_mappings" : [<br />    { "host_port": 8080, "container_port": 80, "protocol": "tcp" },<br />    { "host_port": 8000, "container_port": 8001, "protocol": "udp" }<br />  ]<br />}</pre> | none | none |
 
 For example, the configuration for a port mapping plugin might look like this to an operator (it should be included as part of a [network configuration list](https://github.com/containernetworking/cni/blob/master/SPEC.md#network-configuration-lists).
 ```json
 {
   "name" : "ExamplePlugin",
   "type" : "port-mapper",
-  "runtime_config": {"port_mappings": []}  
+  "capabilities": {"port_mappings": true}
```
rosenhouse: Do we prefer
```
"capabilities: { "port_mappings": true }
```
over
```
"capabilities: [ "port_mappings" ]
```
?

I don't have a strong opinion, but I'm curious what others think.

freehan: `"capabilities: { "port_mappings": true }`

This disallow duplicates.

freehan:
LGTM

CONVENTIONS.md
lxpollitt: `"capabilities: { "port_mappings": true }` is also extendable to support capabilities beyond boolean types in the future if required.

thockin:
LGTM

squeed:
Does it make sense to add some kind of text explaining that this should be used if it is expected that multiple plugins will implement the same functionality, as a way of abstracting that behavior from runtimes?

Something like:
> Plugins should also accept arguments in this manner if they implement the same basic functionality as other plugins. These conventions serve as an abstraction for the runtime. For example, port forwarding is highly implementation specific, but users should be able to select the plugin of their choice without changing the runtime.

squeed:
One small question, otherwise LGTM! Thanks!

tomdee:
@squeed Yes, I think it's worth clarifying that point. I've tried to work your text into this PR - see the latest commit.

squeed:
As I stare at the wall, one last question: do we want to add a `host_ip` parameter, or accept `ip:port`?

tomdee:
@squeed  I say let's merge this and then add that extra parameter if we need it?

squeed:
👍 
I'm working on a iptables-based port-forwarding plugin in parallel - hence the thought.

thockin:
Here's where it gets fun.  To be really friendly, the iptables PF plugin
should *also* open the real socket and leave a process around that is doing
`while(1) { yield(); }` but ideally that would be in C, not Go (since go's
runtime actually wakes up and wastes time :)

tomdee:
@squeed OK, merging.

PR #377, opened 2017-02-20
plugins/main/ptp: set the Sandbox property on the response

squeed:


dcbw:
LGTM

bboreham:
Please can you provide some context why this is needed.

squeed:
PR 145 added passing back created interfaces, including host-side interfaces. We add the `sandbox` property to distinguish the two. This was missing in the `ptp` response.

dcbw:
@bboreham I think I just missed adding that bit to the ptp plugin in #145.  I added it to the other main plugins, just not ptp.

PR #378, opened 2017-02-21
iptables-based default portmap plugin

squeed:
Not yet ready to merge, needs ipv6 testing. Comments welcome.

This is a bog-standard iptables-based port mapping plugin. It does *not* do port reservation.

### DIFF
--- a/plugins/util/portmap/chain.go
+++ b/plugins/util/portmap/chain.go
@@ -1,0 +1,114 @@
+// Copyright 2017 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"fmt"
+
+	"github.com/coreos/go-iptables/iptables"
+)
+
+type chain struct {
+	table       string
+	name        string
+	entryRule   []string // the rule that enters this chain
+	entryChains []string // the chains to add the entry rule
+}
+
+// setup idempotently creates the chain. It will not error if the chain exists.
+func (c *chain) setup(ipt *iptables.IPTables, rules [][]string) error {
+	// create the chain
+	exists, err := chainExists(ipt, c.table, c.name)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		if err := ipt.NewChain(c.table, c.name); err != nil {
+			return err
+		}
+	}
+
+	// Add the rules to the chain
+	for _, rule := range rules {
+		if err := prependUnique(ipt, c.table, c.name, rule); err != nil {
+			return err
+		}
+	}
+
+	// Add the entry rules
+	entryRule := append(c.entryRule, "-j", c.name)
+	for _, entryChain := range c.entryChains {
+		if err := prependUnique(ipt, c.table, entryChain, entryRule); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// teardown idempotently deletes a chain. It will not error if the chain doesn't exist.
+func (c *chain) teardown(ipt *iptables.IPTables) error {
+	// flush the chain
+	// This will succeed *and create the chain* if it does not exist.
+	// If the chain doesn't exist, the next checks will fail.
+	if err := ipt.ClearChain(c.table, c.name); err != nil {
+		return err
+	}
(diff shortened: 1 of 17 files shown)


squeed:
Update: ipv6 works except for SNAT (so connections to `::1` are not forwarded). I do not think this is sufficient to be a blocker.
TODO: tests.

tomdee:
TODO: documentation? :smile:

plugins/util/portmap/chain.go
```diff
@@ -0,0 +1,89 @@
...
+	entryChains []string // the chains to add the entry rule
+}
+
+func (c *chain) setup(ipt *iptables.IPTables, rules [][]string) error {
+	// Create the chain and add its rules
+	if err := ipt.NewChain("nat", c.name); err != nil {
+		return err
+	}
+	for _, rule := range rules {
+		if err := ipt.AppendUnique("nat", c.name, rule...); err != nil {
```
bboreham: Note it is possible for `iptables` to return a "temporarily unavailable" code if another process has the lock.  See [here](https://github.com/weaveworks/weave/blob/master/net/veth.go#L283) for example of detecting this and looping until it works.

Now I mention it, this should really go down into the library.

squeed: `go-iptables` is a bit simplistic. It always passes `--wait` - so by default, iptables will wait (indefinitely!) until the lock is available. So we probably don't have to restart on this level.

bboreham: No it does not wait indefinitely.

plugins/util/portmap/portmap.go
```diff
@@ -0,0 +1,223 @@
...
+
+	return chain{
+		name: name,
+		entryRule: []string{
+			"-m", "addrtype",
+			"--dst-type", "LOCAL",
+			"-m", "comment",
+			"--comment", comment,
+			"-j", name},
+		entryChains: []string{"PREROUTING", "OUTPUT"},
```
freehan: Two more upside for having a summary chain:
1. The chain can choose to only process `--dst-type=Local`, hence all outbound traffic can skip it. 
2. Easier to flush all the rules

squeed: Good point.
Do you also think it makes sense to use the `multiport` module to further reduce runtime? So, the rules would be:

* `-I PREROUTING --dst-type LOCAL -j CNI-HOSTPORTS`
* `-I CNI-HOSTPORTS --dports 80,443,6000 -j CNI-DN-abc123`
* `-A CNI-DN-abc123 --dport 80 -j DNAT`
* `-A CNI-DN-abc123 --dport 443 -j DNAT`
* `-A CNI-DN-abc123 --dport 6000 -j DNAT`

plugins/util/portmap/portmap.go
```diff
@@ -0,0 +1,223 @@
...
+	if isV6 {
+		return chain{
+			name: name,
+			entryRule: []string{
+				"-s", "::1",
+				"!", "-d", "::1",
+				"-m", "comment",
+				"--comment", comment,
+				"-j", name},
+			entryChains: []string{"POSTROUTING"},
```
squeed: We can't assume that we're in a k8s world with the MARK-MASQ setup. However, it can be provided as an option.

squeed: Also, we can add a single `-s 127.0.0.1 ! -d 127.0.0.1` rule to POSTROUTING, which should be pretty efficient.

freehan: I understand. We need to be careful here to not SNAT connections accidentally.

plugins/util/portmap/portmap.go
freehan: I am not sure whether most OS distros have it by default. Since the number of hostport on a node will not be too high, I think it may not worth adding this dependency.

plugins/util/portmap/chain.go
dcbw: @bboreham looks like it will wait indefinitely if you don't give it a specific number of seconds.  IIRC there was some older version that people still ship that doesn't allow a seconds option, so nobody uses it by default, just infinite wait.

plugins/util/portmap/portmap.go
```diff
@@ -0,0 +1,223 @@
...
+
+// snatRules sets up masquerading for connections to localhost:hostport,
+// rewriting the source so that returning packets are correct.
+func snatRules(entries []PortMapEntry, containerIP net.IP) [][]string {
+	isV6 := (containerIP.To4() == nil)
+	var localhostIP string
+	if isV6 {
+		localhostIP = "::1"
+	} else {
+		localhostIP = "127.0.0.1"
```
dcbw: There's already a couple of these, so why not make a small helper function that takes a bool (isV4 or something) returns the right address?  Then you can use it in genSnatChain() too and save some LoC.

squeed:
As per the discussion in the CNi meeting, I've moved this out of the cni tree into a separate plugins repository: https://github.com/containernetworking/plugins/pull/1

Issue #343, opened 2016-12-13
Multi-home pod support in CNI


rkamudhan:
We developed a CNI called *[Multus](https://github.com/Intel-Corp/multus-cni)* to support Multi-homed pod in Kubernetes, which support multi interface in a pod and support different network in a single pod. Please see the use cases in https://github.com/kubernetes/kubernetes/issues/27398

rosenhouse:
Cool.  We'd welcome a PR to our README to add your plugin to our [list of 3rd party plugins](https://github.com/containernetworking/cni/blob/master/README.md#3rd-party-plugins).

Otherwise, I don't believe this issue is actionable.

rkamudhan:
@rosenhouse. Sure, I will create the PR for the 3rd party plugins.

asridharan:
Just FYI, pods in Mesos already support multi-homing (being attached) to multiple CNI networks. Would be great if you could test `Multus` against Mesos. Just wondering as to why does the plugin need to be aware of multi-homing? In Mesos its the `Framework`/operator who tells the container to join multiple CNI networks, and the same CNI can be invoked with different CNI configs?

rkamudhan:
@asridharan I will test Multus with Mesos. Currently, Kubernetes doesn't support multiple networks in the pod. Please refer the https://github.com/kubernetes/kubernetes/issues/27398 for the use case for the Multi-homed pod. Multus connect multiple CNI plugin net config to a pod.

rosenhouse:
This all looks good, nothing actionable here.

PR #379, opened 2017-02-22
README: List multus as 3rd party plugin

rkamudhan:
refer #343 issue. Blog regarding Multus CNI usage http://dougbtv.com/nfvpe/2017/02/22/multus-cni/


plwhite:
FTR, we at Metaswitch have been using the multus plugin pretty extensively over the past few weeks for a demo (which may turn into production use in due course). So I'm keen to see it get recognised here and more widely used.

PR #380, opened 2017-02-22
Add Bryan Boreham as maintainer

bboreham:
Replacing Michael Bridgen


dcbw:
@bboreham Apparently you need to fix the PPC build before you can be a maintainer :)

Anyway, LGTM

bboreham:
There's nothing in the logs for the failed PPC build, other than it timed out.  I would hit the re-try button if there was one.

dcbw:
@bboreham just joking, for the record :)

PR #382, opened 2017-02-27
Documentation: guidance for upgrading to CNI Spec v0.3.0

rosenhouse:
@containernetworking/cni-maintainers An action-item from last week's meeting.

Big thanks for @dcbw for hammering out a great 1st draft.  I've re-worked it, but left both commits, so feel free to compare / revert my changes :-)

I've set "Allow edits from maintainers" so feel free to pile on more changes.  We can squash at merge-time.

### DIFF
--- a/Documentation/spec-upgrades.md
+++ b/Documentation/spec-upgrades.md
@@ -23,8 +23,8 @@ and Release v0.5.0 supports Spec v0.3.0.
 If you maintain CNI configuration files for a container runtime that uses CNI,
 ensure that the configuration files specify a `cniVersion` field and that the
 version there is supported by your container runtime and CNI plugins.
-Configuration files without the version field should be labeled version 0.2.0
-by default. The CNI spec includes example configuration files for 
+Configuration files without a version field should be given version 0.2.0. 
+The CNI spec includes example configuration files for 
 [single plugins](https://github.com/containernetworking/cni/blob/master/SPEC.md#example-configurations)
 and for [lists of chained plugins](https://github.com/containernetworking/cni/blob/master/SPEC.md#example-configurations).
 
@@ -55,11 +55,11 @@ command with the following JSON data:
 ```
 
 Second, for the `ADD` command, a plugin must respect the `cniVersion` field
-provided in the [network configuration JSON](https://github.com/containernetworking/cni/blob/master/SPEC.md#network-configuration).
+provided in the [network configuration JSON](https://github.com/containernetworking/cni/blob/master/SPEC.md#network-configuration). 
 That field is a request for the plugin to return results of a particular format:
 
 - If the `cniVersion` field is not present, then spec v0.2.0 should be assumed
-	and v0.2.0 format JSON returned.
+	and v0.2.0 format result JSON returned.
 
 - If the plugin doesn't support the version, the plugin must error.
 
--- a/Documentation/spec-upgrades.md
+++ b/Documentation/spec-upgrades.md
@@ -217,16 +217,20 @@ other packages, such as the high-level `libcni` package, have been updated to us
 this interface.  Concrete types are now per-version subpackages. The `types/current`
 subpackage contains the latest (spec v0.3.0) types.
 
-The versioned types understand how to convert between most versions. However, converting
-to a higher version will mean some fields are missing.
+When up-converting older result types to spec v0.3.0, fields new in
+spec v0.3.0 (like `interfaces`) may be empty.  Conversely, when
+down-converting v0.3.0 results to an older version, any data in those fields
+will be lost.
 
 | From   | 0.1 | 0.2 | 0.3 |
 |--------|-----|-----|-----|
-| To 0.1 |  ✔  | ✔   |     |
-| To 0.2 |  ✴  |  ✔  |  ✔  |
-| To 0.3 |     |  ✴  |  ✔  | 
+| To 0.1 |  ✔  |  ✔  |  x  |
+| To 0.2 |  ✔  |  ✔  |  x  |
+| To 0.3 |  ✴  |  ✴  |  ✔  |
 
-*(pairs marked with ✴ will have some missing data)*
+✔ : lossless conversion
+✴ : higher-version output may have empty fields
+x : lower-version output is missing some data
 
 A container runtime should use `current.NewResultFromResult()` to convert the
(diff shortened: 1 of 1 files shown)


Documentation/spec-upgrades.md
```diff
@@ -0,0 +1,253 @@
...
+for this repo.  For example, Release v0.4.0 supports Spec version v0.2.0,
+and Release v0.5.0 supports Spec v0.3.0.
+
+----
+
+## For CNI Users
+If you maintain CNI configuration files for a container runtime that uses CNI,
+ensure that the configuration files specify a `cniVersion` field and that the
+version there is supported by your container runtime and CNI plugins.
+Configuration files without the version field should be labeled version 0.2.0
```
dcbw: Perhaps clarify to "Existing configuration files without a version field should be given version 0.2.0".  Maybe also note that the CNI specification requires that runtimes and plugins treat configuration without an explicit version tag as 0.2.0.

Documentation/spec-upgrades.md
```diff
@@ -0,0 +1,253 @@
...
+  "supportedVersions": [ "0.1.0", "0.2.0", "0.3.0" ]
+}
+```
+
+Second, for the `ADD` command, a plugin must respect the `cniVersion` field
+provided in the [network configuration JSON](https://github.com/containernetworking/cni/blob/master/SPEC.md#network-configuration).
+That field is a request for the plugin to return results of a particular format:
+
+- If the `cniVersion` field is not present, then spec v0.2.0 should be assumed
+	and v0.2.0 format JSON returned.
```
dcbw: Perhaps clarify to "v0.2.0 format IPAM structure JSON returned."

Documentation/spec-upgrades.md
```diff
@@ -0,0 +1,253 @@
...
+func main() {
+	skel.PluginMain(cmdAdd, cmdDel, version.PluginSupports("0.1.0", "0.2.0", "0.3.0"))
+}
+```
+
+Alternately, to use the result from a delegated IPAM plugin, the `result`
+value might be formed like this:
+
+```go
+ipamResult, err := ipam.ExecAdd(netConf.IPAM.Type, args.StdinData)
```
dcbw: perhaps ipamResult -> resultInterface to be consistent with below and make it clear the returned value is an interface?

Documentation/spec-upgrades.md
```diff
@@ -0,0 +1,253 @@
...
+In particular, runtimes with existing installed bases should add support for CNI
+spec version 0.3.0 while maintaining compatibility with older versions.
+
+To support multiple versions of the CNI spec, runtimes should be able to
+call both new and legacy plugins, and handle the results from either.
+
+When calling a plugin, the runtime must request that the plugin respond in a
+particular format by specifying the `cniVersion` field in the
+[Network Configuration](https://github.com/containernetworking/cni/blob/master/SPEC.md#network-configuration)
+JSON block.  The plugin will then respond with
```
dcbw: Might be worth noting that we expect the container runtime to inject its expected `cniVersion` into the JSON when sending to the plugin, if it supports CNI spec >= 0.3.0.  And note that the runtime cannot rely on any on-disk or pre-supplied config JSON that includes a `cniVersion` unless the runtime guarantees that it can read that `cniVersion` Result JSON in some way.

Documentation/spec-upgrades.md
```diff
@@ -0,0 +1,253 @@
...
+of a successful response to `VERSION`, assume that the plugin only supports
+CNI spec v0.1.0.
+
+#### Handle missing data in v0.3.0 results
+The Result for the `ADD` command in CNI spec version 0.3.0 includes a new field
+`interfaces`.  An IP address in the `ip` field may describe which interface
+it is assigned to, by placing a numeric index in the `interface` subfield.
+
+However, some plugins which are v0.3.0 compatible may nonetheless omit the
+`interfaces` field and/or set the `interface` index value to `-1`.  Runtimes
```
dcbw: Could probably use some clarification.  I think the cases are:

1. IPAM plugins: can legally return just 'ips', 'routes', and 'dns' without 'interfaces'
2. non-IPAM plugins that don't create interfaces or do any IPAM (eg, tuning): return "{}" and don't include any Result dict keys
3. non-IPAM plugins that do touch interfaces: must return at least 'interfaces{}'; other keys (ips, routes, dns) are optional but if any 'ips' are returned, the plugin must set interface indexes in each IP address

Since Runtimes cannot exec IPAM plugins separately, case #1 doesn't apply to runtimes.  Thus, we are left with:

*0.3.0 plugins may return either empty Result JSON (eg, tuning) when they do not create interfaces or do IPAM, or will include Result JSON with a valid `interfaces` key and optional 'ips'.  If 'ips' are given, each IP address will have a valid interface index or the Result is invalid.*

I should add some checks to the current.Result unmarshalling to enforce that...

rosenhouse: > If ips are given, each IP address will have a valid interface index or the Result is invalid.

Is that consistent with [this up-convert code](https://github.com/containernetworking/cni/blob/master/pkg/types/current/types.go#L70-L87)?  What if a 0.3.0 plugin delegates to a 0.2.0 plugin, and wants to up-convert the result to 0.3.0 and return that?  Is that possible?

dcbw: @rosenhouse I did not consider a 0.3.0 plugin delegating to a 0.2.0 plugin, and assumed that all plugins it the chain are required to be the same version.  That seems simpler to require that, though less flexible.  But hey, if you have a hard dep on old sh*t maybe you deserve to be stuck on an older version.  What do you think?

Documentation/spec-upgrades.md
squeed: This is a fun complection - we've tied together config version and results version. Should runtimes be overwriting the config file version?

Documentation/spec-upgrades.md
dcbw: @squeed @tomdee @bboreham what do you think?

tomdee:
LGTM

tomdee:
The CI failure shouldn't be anything to do with this PR - it's just a docs change. It only failed on one go version and one pltfor (AMD64 on Go 1.7) - and it seemed to be a problem reporting to coveralls.
```
```

I'm going to assume it's just a flake and merge anyway.

PR #373, opened 2017-02-17
spec,libcni: add support for injecting runtimeConfig into plugin stdin data

dcbw:
And "runtime_config"->"runtimeConfig" to make capitalizatoin consistent
with other CNI config keys like "cniVersion" and such.

@containernetworking/cni-maintainers

Issue #251, opened 2016-06-16
Proposal: Support hypervisor container runtime for CNI


feiskyer:
## Overview

CNI is designed to provide a generic plugin-based networking solution for application containers on Linux. There are many container runtime implementations today. To isolate container networking from host, some of them are based on linux network namespace (e.g., [docker](http://github.com/docker/docker), [rkt](https://github.com/coreos/rkt) without lkvm), while others are based on hypervisor (e.g., [rkt with lkvm](https://github.com/coreos/rkt/blob/master/Documentation/running-lkvm-stage1.md), [HyperContainer](https://github.com/hyperhq/hyperd)). Current CNI design already supports runtimes based on network namespace, but not hypervisors. So this proposal amis to add hypervisor runtime support to CNI.
## Considerations

There many differences between hypervisor and linux network namespace on networking:
- Hypervisor has its own network stack, there is no need to create a network namespace for hypervisor. Instead, CNI should create a tap device for hypervisor.
- Since the network is configured inside hypervisor, CNI network plugin couldn't configure ip address directly outside. Instead, CNI should prepare the network device (e.g., create bridge and connect a tap device), then pass the network information (including the device, IP address, gateway, routes, etc.) back to the runtime, and let runtime do remaining configuration work.
## Design

To support hypervisor-based runtimes, a few changes should be made in CNI:
- Let CNI be aware of runtime type (hypervisor or netns)
- Configure tap device instead of veth pair for hypervisor-based runtime
- Pass network information back to runtime
### Aware of runtime type

Add a new field `UsesTapDevice` in `RuntimeConf`, e.g.:

``` go
type RuntimeConf struct {
    // The ID of the container
    ContainerID string
    // If set, a tap device instead of veth pair should be created
    UsesTapDevice bool
    // The network namespace path. 
    // Optional. Only need if IsHypervisor is not set.
    NetNS       string
    // Name of the interface inside the container. 
    IfName      string
    // Extra arguments
    Args        [][2]string
}
```
### Configure tap device instead of veth pair

This will be the job of network plugins. Since there are no network namespace, network plugin should create a tap device directly on host. Existing plugins would be refactored to create tap device for hypervisor runtime. 

The network configuration and IPAM could keep as is, so container's ip address could be allocated as before.
### Pass network info back to runtime

Dan has already started this on [#145](https://github.com/containernetworking/cni/pull/145). 

CC @dcbw @rosenhouse @squaremo @steveeJ @tomdee

[...]

PR #384, opened 2017-02-28
RFC: CNI Runtime hint: Provide hints regarding the runtime in use

mcastelino:
All main CNI plugins assume they have to deal with namespaces as the primary form of isolation in the case of containers. However with CRI-O adding support for multiple runtimes including virtual machines, it makes sense to provide an option to support these runtimes without breaking backward compatibility.

https://github.com/kubernetes/kubernetes/pull/40662

In the case of runtimes that use virtual machines as the isolation implementation, this information can be passed into CNI as an ***optional*** parameter.  

The VM runtimes have tried to address this by using integration bridges in some cases. However this approach does not always work (e.g macvlan)

When the runtime hint is present the network plugin can choose an interface setup that is optimal for that runtime type. The plugin is free to completely ignore the hint and assume it only supports namespace centric runtimes. This also means that runtimes should not assume that the hint will result in a interface that is optimal, and will need to handle the cases where a namespace friendly interface is created.

This is by no means mandatory and the runtime should be able to detect and support the case where an namespace centic interface is created.

This PR attempts to seek feedback on this approach using the bridge plugin.

If this runtime hint CNI_RUNTIME_TYPE is acceptable I will add support for the same for all other plugins.

For a quick test to observe the default behavior vs the behavior when the hint is accepted.

```
sudo CNI_PATH=$CNI_PATH CNI_RUNTIME_TYPE="vm" ./priv-net-run.sh ip -d l
sudo CNI_PATH=$CNI_PATH ./priv-net-run.sh ip -d l

```

cc @sameo @mrunalp @thockin

### DIFF
--- a/pkg/ip/link.go
+++ b/pkg/ip/link.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"syscall"
 
 	"github.com/containernetworking/cni/pkg/ns"
 	"github.com/containernetworking/cni/pkg/utils/hwaddr"
@@ -44,6 +45,22 @@ func makeVethPair(name, peer string, mtu int) (netlink.Link, error) {
 	}
 
 	return veth, nil
+}
+
+func makeTap(name string, mtu int) (netlink.Link, error) {
+	tap := &netlink.Tuntap{
+		LinkAttrs: netlink.LinkAttrs{
+			Name:  name,
+			Flags: net.FlagUp,
+			MTU:   mtu,
+		},
+		Mode: syscall.IFF_TAP,
+	}
+	if err := netlink.LinkAdd(tap); err != nil {
+		return nil, err
+	}
+
+	return tap, nil
 }
 
 func peerExists(name string) bool {
@@ -152,6 +169,46 @@ func SetupVeth(contVethName string, mtu int, hostNS ns.NetNS) (net.Interface, ne
 		return net.Interface{}, net.Interface{}, err
 	}
 	return ifaceFromNetlinkLink(hostVeth), ifaceFromNetlinkLink(contVeth), nil
+}
+
+// SetupTap sets up a vm friendly tap interface and connects it to specified bridge
+func SetupTap(tapName string, mtu int, containerNS ns.NetNS, br *netlink.Bridge) (tap netlink.Link, err error) {
+	tap, err = makeTap(tapName, mtu)
+	if err != nil {
+		return
+	}
+
+	if err = netlink.LinkSetUp(tap); err != nil {
+		err = fmt.Errorf("failed to set %q up: %v", tapName, err)
+		return
+	}
+
+	tap, err = netlink.LinkByName(tapName)
+	if err != nil {
+		err = fmt.Errorf("failed to lookup %q: %v", tapName, err)
+		return
+	}
+
+	// connect tap to the bridge
+	err = netlink.LinkSetMaster(tap, br)
+	if err != nil {
+		err = fmt.Errorf("failed to connect %q to bridge %v: %v", tap.Attrs().Name, br.Attrs().Name, err)
+		return
+	}
+
+	err = netlink.LinkSetUp(tap)
+	if err != nil {
+		err = fmt.Errorf("failed to set %q up: %v", tapName, err)
+		return
+	}
+
+	err = netlink.LinkSetNsFd(tap, int(containerNS.Fd()))
+	if err != nil {
+		err = fmt.Errorf("failed to move tap to the container netns: %v", err)
+		return
+	}
+
+	return
 }
 
 // DelLinkByName removes an interface link.
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -37,13 +37,14 @@ const defaultBrName = "cni0"
 
 type NetConf struct {
 	types.NetConf
-	BrName       string `json:"bridge"`
-	IsGW         bool   `json:"isGateway"`
-	IsDefaultGW  bool   `json:"isDefaultGateway"`
-	ForceAddress bool   `json:"forceAddress"`
-	IPMasq       bool   `json:"ipMasq"`
-	MTU          int    `json:"mtu"`
-	HairpinMode  bool   `json:"hairpinMode"`
+	BrName        string                 `json:"bridge"`
+	IsGW          bool                   `json:"isGateway"`
+	IsDefaultGW   bool                   `json:"isDefaultGateway"`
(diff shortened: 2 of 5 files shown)


dcbw:
We're attempting to move away from the environment variables generally and towards the "runtimeConfig" option you'll find in CONVENTIONS.md.  So one option here is to use that instead of creating a new top-level option.  This would involve specifying a new capability that indicated that the plugin could handle VM-type requests, which the runtime would set when calling the plugin.  Per https://github.com/containernetworking/cni/pull/373 this runtime-set option would then get injected into the plugin's 'runtimeConfig' key in the JSON passed via stdin.

There's another question though, and that's whether the same CNI configuration JSON can/should be used for both container and VM-based operation.  Perhaps that's the right way forward, one config to rule them all and a runtimeConfig key to indicate VM vs. container.  But maybe that's attempting to simplify things too much as well, when these cases are pretty different at the low level of sandbox configuration.

Here's another thought, which involves recently added plugin config lists and capabilities:

1) we create new plugins 'veth' and 'tap', where all they do is create the specified interface type in the given sandbox (eg, for the VM or inside the netns)
2) the config json for the both veth and tap plugins specifies 'configureVM=true'
3) the ordered config list is [ veth, tap, bridge ]
4) when veth is called, if the runtime passes the "configureVM" capability arg, veth simply does nothing
5) control passes to the 'tap' plugin which recognizes the "configureVM" capability arg, and makes a tap interface
6) the Bridge plugin sees the "prevResult" of the tap plugin, and adds the tap to the bridge, and runs IPAM for the tap interface
7) the Bridge plugin fills in the addresses and returns the Result to the runtime
8) the runtime knows this is for a VM, since it was responsible for passing "configureVM=true" as a CapabilityArg.  Thus, it knows that the returned sandboxed interface should now be made available to the VM.

Just a thought, using some of the mechanisms we already have in CNI.  I hesitate to keep adding interface types to the bridge plugin, when perhaps we should just make separate plugins for each interface.  eg more composability, now that we have config lists.

@containernetworking/cni-maintainers what do you think?

mcastelino:
@dcbw 

> We're attempting to move away from the environment variables generally and towards the "runtimeConfig" option you'll find in CONVENTIONS.md. So one option here is to use that instead of creating a new top-level option.

I will switch to using the runtimeConfig as you indicated.

> Here's another thought, which involves recently added plugin config lists and capabilities:

This plugin chaining method is indeed modular and makes sense. I will implement it for a few of the common network plugins and send a follow up RFC PR.

>There's another question though, and that's whether the same CNI configuration JSON can/should be 
> used for both container and VM-based operation. Perhaps that's the right way forward, one config to rule them 

Ideally the input CNI configuration should remain the same.  But as you indicated the sandbox type is different and in some cases the configuration needs to be performed within the VM vs in the containing namespace. e.g the IPAM configuration. 

So the runtime would need to consume the result (https://github.com/containernetworking/cni/blob/master/SPEC.md#parameters) of the CNI plugin and and in some cases apply *some* of the configuration within the VM.  In that case part of the result will be indicative of desired configuration for some parts of the result. Also performing this IPAM configuration on the host namespace is not required.

squeed:
First, I think it would be awesome if the bridging plugin (and flannel, eventually) were the second step in a chained invocation, putting to rest the current pseudo-chaining we have now.

However, as you all have said, the semantics between configuring vm interfaces and ns interfaces are really far too great. For example, every existing interface plugin expects to be able to enter the namespace and set the IP address and routing table. Other plugins further down the chain may also manipulate the container-side IF.

While the CNI invocation paradigm definitely seems useful in a VM environment, I don't see how an arbitrary (for some definition of arbitrary) CNI configuration can realistically work in both scenarios.

mcastelino:
@squeed 

>While the CNI invocation paradigm definitely seems useful in a VM environment, I don't see how an arbitrary (for some definition of arbitrary) CNI configuration can realistically work in both scenarios.

Yes, this will be a challenges. Today we handle that in [the case of clear containers](https://github.com/containers/virtcontainers#cni) by actually scanning the entire network ns to figure out what has really happened. Yes, we can rely on the information returned by the CNI plugin, but we scan the network ns as a failsafe.

However I would still like to attempt supporting this for the default and common plugins. This will at least get us an interfaces that we can direct plugin into the VM.

feiskyer:
@mcastelino @dcbw @bboreham This is actually same idea with proposal #251. But we didn't get an  agreement there. Should we revisit the proposal?

bboreham:
@feiskyer This general topic was discussed at length in the Community Sync meeting; I hope I am summarising correctly in saying we are happy to look at dropping restrictions that make something impossible within the current CNI spec, but unlikely to add, piece by piece, every specific type of networking option to the spec.

mcastelino:
@dcbw @bboreham @feiskyer we are working with the upstream linux community to close the gaps for some interfaces. So providing the optional hint to the plugin, allows the plugin to make a choices around the type of interface to create. We do not mandate a tap interface per-se, but request for a VM friendly interface. However the VM runtime should not expect to always get a VM friendly interface

Here is an upstream patch set we are working on right now https://lkml.org/lkml/2017/1/25/665
This includes ipvlan tap: https://lkml.org/lkml/2017/1/6/844
And another part of the same patch set to makes tap a separate module for other types of virtual interface to use https://lkml.org/lkml/2017/1/25/664

Our goal is to provide the network plugins to ability make the right choices.

/cc @sainath14

bboreham:
I'm afraid that kernel patches make the situation harder for me to understand.

Why can't the tap interface be created by a separate plugin?

mcastelino:
@bboreham I am working on using plugin chaining to setup the tap interfaces in a separate plugin. However in some cases like macvlan, ipvlan it may not be possible to split the creation of the tap interface using the plugin chaining mechanism. Hence the kernel patches.

mcastelino:
@dcbw reworked the current PR to use "runtimeConfig" and  "configureVM" capability, Also added the capability to macvlan to demonstrate the case where chaining may not make sense.

The change to the bridg plugin is just temporary. I will further rework this patch set to split the bridge plugin into bridge, veth and tap

These can also be exercised from the command line

```
/etc/cni/net.d/10-mynet.conf
{
        "cniVersion": "0.3.0",
        "name": "mynet",
        "type": "bridge",
        "bridge": "cni0",
        "isGateway": true,
        "ipMasq": true,
        "ipam": {
                "type": "host-local",
                "subnet": "10.22.0.0/16",
                "routes": [
                        { "dst": "0.0.0.0/0" }
                ]
        },
        "runtimeConfig": {
                "configureVM": true
        }
}

```

bboreham:
Closing stale PR.

PR #385, opened 2017-02-28
Add option to disable default route inside container namespace

richardpen:
### Summary
Adding a new option to remove the default route in the container namespace when using the bridge plugin. For security concerns, users may only want specific, white-listed routes where even having the default route is a security vulnerability for their application. Enable this option will remove both IPV4 and IPV6 default routes and left only the routes specified in the configuration file.

### Test done
Added new test in `bridge_test.go` to cover this change.
`./test` ran successfully, apart from the `macvlan` Operations suite. I tested this on Amazon Linux and there's no support for `macvlan` on that platform. The failure log is pasted next:
```
• Failure [0.053 seconds]
macvlan Operations
/home/ec2-user/cni/gopath/src/github.com/containernetworking/cni/plugins/main/macvlan/macvlan_test.go:192
  creates an macvlan link in a non-default namespace [It]
  /home/ec2-user/cni/gopath/src/github.com/containernetworking/cni/plugins/main/macvlan/macvlan_test.go:102

  Expected error:
      <*errors.errorString | 0xc82000bf70>: {
          s: "failed to create macvlan: operation not supported",
      }
      failed to create macvlan: operation not supported
  not to have occurred

  /home/ec2-user/cni/gopath/src/github.com/containernetworking/cni/plugins/main/macvlan/macvlan_test.go:87
```

### DIFF
--- a/Documentation/bridge.md
+++ b/Documentation/bridge.md
@@ -40,3 +40,4 @@ If the bridge is missing, the plugin will create one on first use and, if gatewa
 * `mtu` (integer, optional): explicitly set MTU to the specified value. Defaults to the value chosen by the kernel.
 * `hairpinMode` (boolean, optional): set hairpin mode for interfaces on the bridge. Defaults to false.
 * `ipam` (dictionary, required): IPAM configuration to be used for this network.
+* `noVethDefaultRt`(boolean, optional): Indicate if the default route inside the veth namespace should be removed. Defaults to false.
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -37,13 +37,14 @@ const defaultBrName = "cni0"
 
 type NetConf struct {
 	types.NetConf
-	BrName       string `json:"bridge"`
-	IsGW         bool   `json:"isGateway"`
-	IsDefaultGW  bool   `json:"isDefaultGateway"`
-	ForceAddress bool   `json:"forceAddress"`
-	IPMasq       bool   `json:"ipMasq"`
-	MTU          int    `json:"mtu"`
-	HairpinMode  bool   `json:"hairpinMode"`
+	BrName          string `json:"bridge"`
+	IsGW            bool   `json:"isGateway"`
+	IsDefaultGW     bool   `json:"isDefaultGateway"`
+	ForceAddress    bool   `json:"forceAddress"`
+	IPMasq          bool   `json:"ipMasq"`
+	MTU             int    `json:"mtu"`
+	HairpinMode     bool   `json:"hairpinMode"`
+	NoVethDefaultRt bool   `json:"noVethDefaultRt"`
 }
 
 func init() {
@@ -310,11 +311,28 @@ func cmdAdd(args *skel.CmdArgs) error {
 			return err
 		}
 
-		// Refetch the veth since its MAC address may changed
 		link, err := netlink.LinkByName(args.IfName)
 		if err != nil {
 			return fmt.Errorf("could not lookup %q: %v", args.IfName, err)
 		}
+
+		// Remove the default route for both ipv4 and ipv6 if this option was set
+		if n.NoVethDefaultRt {
+			// Get all the routes
+			routes, err := netlink.RouteList(link, netlink.FAMILY_ALL)
+			if err != nil {
+				return fmt.Errorf("Failed to get the route list inside container network namespace")
+			}
+			for _, r := range routes {
+				if r.Gw == nil {
+					if err = netlink.RouteDel(&r); err != nil {
+						return fmt.Errorf("Failed to delete the default route inside container network namespace")
+					}
+				}
+			}
+		}
+
+		// Refetch the veth since its MAC address may changed
 		containerInterface.Mac = link.Attrs().HardwareAddr.String()
 
 		return nil
--- a/plugins/main/bridge/bridge_test.go
+++ b/plugins/main/bridge/bridge_test.go
@@ -123,7 +123,7 @@ var _ = Describe("bridge Operations", func() {
 		Expect(err).NotTo(HaveOccurred())
 	})
 
-	It("configures and deconfigures a bridge and veth with default route with ADD/DEL", func() {
+	It("configures and deconfigures a bridge and veth with default route and gateway with ADD/DEL", func() {
 		const BRNAME = "cni0"
 		const IFNAME = "eth0"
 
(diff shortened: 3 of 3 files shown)


dcbw:
The `isDefaultGateway` bridge plugin option is supposed to control whether or not the bridge plugin adds a default route, and it is false by default.  Only when that option is 'true' will the bridge plugin add a default route by itself.  Does your configuration set that option to true?

richardpen:
@dcbw Thanks for your response, here is the test I did:
Without setting `isDefaultGateway`:
```
{
     "cniVersion": "0.2.0",
     "name": "mynet",
     "type": "bridge",
     "bridge": "cni0",
     "isGateway": true,
     "ipMasq": false,
     "ipam": {
         "type": "host-local",
         "subnet": "192.168.3.0/16",
         "routes":[
         {"dst": "10.0.0.2/32"}
 		]
     }
 }
```
And here is the output from `ip route`:
```
$ ip netns exec container3 ip route
10.0.0.2 via 192.168.3.1 dev eth0  
192.168.0.0/16 dev eth0  proto kernel  scope link  src 192.168.3.8 # This line is not intended

$ ip netns exec container3 ip -6 route
fe80::/64 dev eth0  proto kernel  metric 256  pref medium # This line is not intended
```

Setting `isDefaultGateway` to be true:
```
{
    "cniVersion": "0.2.0",
    "name": "mynet",
    "type": "bridge",
    "bridge": "cni0",
    "isGateway": true,
    "isDefaultGateway": true,
    "ipMasq": false,
    "ipam": {
        "type": "host-local",
        "subnet": "192.168.3.0/16",
        "routes":[
        {"dst": "10.0.0.2/32"}
]
    }
}
```
And the output from `ip route`:
```
$ ip netns exec container3 ip route
default via 192.168.3.1 dev eth0   # This is the one not intended
10.0.0.2 via 192.168.3.1 dev eth0
192.168.0.0/16 dev eth0  proto kernel  scope link  src 192.168.3.9 # This line is not intended

$ ip netns exec container3 ip -6 route
fe80::/64 dev eth0  proto kernel  metric 256  pref medium # This line is not intended
```

From the test above, I think the one `192.168.0.0/16 dev eth0  proto kernel  scope link  src 192.168.3.8 ` is not controllable anywhere, but `default via 192.168.3.1 dev eth0` is configured by `isDefaultGateway`. Please suggest if there is other option that can remove all of the routes not specified in the configuration file. Thanks.

dcbw:
`192.168.0.0/16 dev eth0 proto kernel scope link src 192.168.3.8` is not a default route, but a "prefix" route added by the kernel whenever you add an IP address to an interface. It is not added by CNI.  It indicates that any other 192.168.x.x/16 IP address is reachable from eth0 and thus any packet destined for 192.168.x.x/16 will simply be dumped onto eth0, and whatever is on the other end of eth0 has responsibility for filtering those packets.

So if what you'd like to do is prevent addition of the prefix route, and thus cause an EHOSTUNREACH error whenever the containerized application attempts to send something to another 192.168.x.x/16 address, a new option to the bridge plugin is probably the right way for now.  But this option would be something like "suppressPrefixRoute" and when set to TRUE would tell pkg/ipam to set the IFA_F_NOPREFIXROUTE flag when adding the IP address to the interface.

Obviously, the containerized app can send whatever packets it wants on the interface, including open a raw socket and construct its own IP packets regardless of what the container's routing tables say.  So preventing a prefix route from being created really doesn't increase security of the container.  The restrictions really need to be enforced on the other end of the 'eth0' interface, which in this case is the host's network stack.

richardpen:
@dcbw Thanks for explaining this. In our use case we want to start multiple containers within their own network namespace with a bridge connected. And by default these containers shouldn't be able to communicate with each other, but can reach to the routes we specified, so by default we don't want the prefix route to be added.

I tested with the option `IFA_F_NOPREFIXROUTE(512)`, which won't set the prefix route for ipv4, but the ipv6 prefix route is still there. Did I miss something, can you suggest?
```
-               addr := &netlink.Addr{IPNet: &ipc.Address, Label: ""}
+               addr := &netlink.Addr{IPNet: &ipc.Address, Label: "", Flags: 512}
```

Could you kindly suggest about enforcing on the host's network, will it work by adding route to drop packets that src and dst are both in the subnet?

Thanks,
Richard

dcbw:
It might depend on the kernel you have.  The IPv6 handling of NOPREFIXROUTE was added to the 3.14 kernel version.  But if you have it for IPv4 (added in kernel 4.4) then you should have it for IPv6 already too.  Which kernel version do you have?

squeed:
This smells like something that should be a chained plugin rather than part of the initial interface creation?

dcbw:
> This smells like something that should be a chained plugin rather than part of the initial interface creation?

@squeed do you mean a plugin to remove routes rather than preventing them from being added in the first place?

richardpen:
@dcbw I tested both on the ubuntu with kernel 4.8.0 and amazon linux ami with kernel 4.4.41, both has the same result: the ipv6 "prefix" route is added while ipv4 "prefix" route is not added.

squeed:
@dcbw indeed, that's what I mean. I propose it as an alternative to modifying every interface plugin.

My justification for this is basically just that this sort of behavior (prefix route removal) seems well outside standard network story.

richardpen:
@dcbw @squeed The flag `IFA_F_NOPREFIXROUTE` doesn't work for IPV6 in my test, can you take a look? For the approach to implement this, it may be better to prevent the "prefix" route to be added rather than remove it after added, because in that way the "add and remove" seems duplicate operation. But since the "prefix" route is added by the kernel, I'm not an expert of this, so I don't know if it's configurable. I think it would be good if we can make `IFA_F_NOPREFIXROUTE` work for IPV6.

> My justification for this is basically just that this sort of behavior (prefix route removal) seems well outside standard network story.

I agree with this is not a standard network story, which may not be most customers' use case. But definitely it will benefit some of use case like us to make it more configurable.

Please let me know what else I can do to make this feature implemented in the plugin.

Thanks,
Richard

richardpen:
@dcbw @squeed any suggestion on this?

richardpen:
Hi, 
@dcbw @squeed can anyone take a look at this, let me know if there is something I can help?

Thanks,
Richard

bboreham:
Where is the IPv6 address coming from?  I don't think the CNI bridge plugin adds one.

bboreham:
I don't think CNI should depend on something as new as `IFA_F_NOPREFIXROUTE`, if it was added in kernel 4.4.

I'm ok with the idea that removing the default route goes in a chained plugin.  Not 100% convinced, but it sounds plausible.

squeed:
This needs to be closed, and either refiled on the plugins repository, or adapted to a chained plugin.

PR #386, opened 2017-03-01
spec,plugins,core: convert Route gateway to an array of NextHops

dcbw:
Multiple route next hops can be used for multipath routing or for
gateway redundancy.  Instead of allowing only one gateway for each
route, allow one or more via the NextHops array of gateway IP
addresses.

@containernetworking/cni-maintainers @tgraf


dcbw:
Just going to drop this one for now...  Can resurrect later if it's necessary to do so.

PR #391, opened 2017-03-06
Validate rangeStart and rangeEnd specified in conf

tangle329:
If we run plugin like this, 10.1.0.0 is allocated for use. which is the wrong way：
`
{
    "cniVersion": "0.2.0",
    "name": "mynet",
    "type": "bridge",
    "bridge": "cni10",
    "isGateway": true,
    "ipMasq": true,
    "ipam": {
        "type": "host-local",
        "subnet": "10.1.0.0/16",
        "rangeStart": "10.1.0.0",
        "rangeEnd": "10.1.255.255",
        "gateway": "10.1.0.1"
    }
}

 export CNI_PATH=$GOPATH/src/github.com/containernetworking/cni/bin/
 export CNI_COMMAND=ADD
 export CNI_NETNS=/var/run/test
 export CNI_IFNAME=eth0
 export CNI_ARGS="FOO=BAR;ABC=123"
 $GOPATH/src/github.com/containernetworking/cni/bin/host-local < /etc/cni/net.d/10-mynet.conf
2017/03/06 16:44:55 Error retriving last reserved ip: Failed to retrieve last reserved ip: open /var/lib/cni/networks/mynet/last_reserved_ip: no such file or directory
{
    "ip4": {
        "ip": "10.1.0.0/16",                      <== We should skip this address.
        "gateway": "10.1.0.1"
    },
    "dns": {}
}
`

Signed-off-by: Tang Le <tangle3@wanda.cn>

### DIFF
--- a/plugins/ipam/host-local/backend/allocator/allocator.go
+++ b/plugins/ipam/host-local/backend/allocator/allocator.go
@@ -56,13 +56,13 @@ func NewIPAllocator(conf *IPAMConfig, store backend.Store) (*IPAllocator, error)
 	start = ip.NextIP(start)
 
 	if conf.RangeStart != nil {
-		if err := validateRangeIP(conf.RangeStart, (*net.IPNet)(&conf.Subnet), nil, nil); err != nil {
+		if err := validateRangeIP(conf.RangeStart, (*net.IPNet)(&conf.Subnet), start, end); err != nil {
 			return nil, err
 		}
 		start = conf.RangeStart
 	}
 	if conf.RangeEnd != nil {
-		if err := validateRangeIP(conf.RangeEnd, (*net.IPNet)(&conf.Subnet), start, nil); err != nil {
+		if err := validateRangeIP(conf.RangeEnd, (*net.IPNet)(&conf.Subnet), start, end); err != nil {
 			return nil, err
 		}
 		end = conf.RangeEnd
--- a/plugins/ipam/host-local/backend/allocator/allocator_test.go
+++ b/plugins/ipam/host-local/backend/allocator/allocator_test.go
@@ -263,33 +263,55 @@ var _ = Describe("host-local ip allocator", func() {
 		})
 
 		It("RangeStart must be in the given subnet", func() {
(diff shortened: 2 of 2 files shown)


bboreham:
Your description speaks about skipping 10.1.0.0, but your code doesn't actually touch the path where the address is chosen.

I think what you are trying to do is make the plugin validate `rangeStart` and `rangeEnd`.  Is that correct?

tangle329:
@bboreham You are right. The description has been corrected.

bboreham:
The code looks fine to me; however there are now 3 commits in this branch.

If you can, @tangle329, please squash those commits into one.  Otherwise I'll cherry-pick fc5fd822f3aa4d72df5f00b2a50e6d42378a9fc3 which seems to be the meat.

rosenhouse:
Would it be possible to get test coverage of this change?

dcbw:
@tangle329 here's a testcase, if you want to replace the existing "RangeStart must be in the given subnet" testcase with this code:

```
		It("RangeStart must be in the given subnet", func() {
			testcases := []struct{
				name  string
				ipnet string
				start string
			}{
				{"outside-subnet", "192.168.1.0/24", "10.0.0.1"},
				{"zero-ip", "10.1.0.0/16", "10.1.0.0"},
			}

			for _, tc := range testcases {
				subnet, err := types.ParseCIDR(tc.ipnet)
				Expect(err).ToNot(HaveOccurred())

				conf := IPAMConfig{
					Name:       tc.name,
					Type:       "host-local",
					Subnet:     types.IPNet{IP: subnet.IP, Mask: subnet.Mask},
					RangeStart: net.ParseIP(tc.start),
				}
				store := fakestore.NewFakeStore(map[string]string{}, net.ParseIP(""))
				_, err = NewIPAllocator(&conf, store)
			}
		})
```

tangle329:
@dcbw  Thanks.

tangle329:
Any update on this PR?

dcbw:
@tangle329 we require two maintainer approvals; and this PR seems to have slipped through the cracks.  Sorry about that!  So LGTM too.

PR #394, opened 2017-03-09
hwaddr: Generate MAC address for IPv6-only pods

### LINKED ISSUE #370: generate hwaddr (mac address) from an IPv6 address
Currently, an IPv4 address is needed to generate a MAC address. The [GenerateHardwareAddr4 function](https://github.com/containernetworking/cni/blob/master/pkg/utils/hwaddr/hwaddr.go#L48-L49) should be updated or replicated to support IPv6 addresses for generating a  mac address of a container.

leblancd:
This change adds logic to set the MAC (hardware) address for IPv6-only
containers.

Currently, the SetHWAddrByIP depends on an IPv4 IP address being
configured, and it generates the 48-bit MAC address by wrapping the
4-byte IPv4 address with a special, hard-coded 2-byte prefix:
    [0a:58] + [IPv4-address]
The IPv6-only case is not being handled.

What's being proposed with this change set is to modify SetHWAddrByIP
behavior as follows:
**- IPv4-only or Dual-Stack (IPv4 + IPv6):**
  No change in behavior, MAC address is based on IPv4 address:
    [0a:58] + [IPv4-address]
**- IPv6-only:**
  Generate a MAC address using a different hard-coded prefix
  (to make the V6 MAC address space separate from the
  V4 MAC address space):
    [6a:58] + [four-random-octets]
  Here, the four random octets are derived by taking the last 4 bytes
  of the original MAC address as assigned by the OS when the intf/link
  is created. These are assumed to be random.

In both cases, the IPv6 address is assigned independently (uncoupled)
from the IPv4 and MAC address assignments.

There was some concern expressed about whether the IPv6 Neighbor Discovery (ND) cache
will be subject to some "thrashing" or collisions as containers are rapidly cycled
and IPv6 addresses eventually get recycled. After all, this is why, in the IPv4 case,
the MAC address was generated based on the IPv4 address: see squeed's
explanation in the comments below. However, the IPv6 ND cache
will be resilient to IPv6 unicast addresses getting recycled/reused by containers
with different MAC addresses. The source of this resiliency is that when a node/container
sends an IPv6 Neighbor Advertisement (NA) claiming that it "owns" an IPv6 unicast address,
it will set the "ovr" flag (or "o-bit"), indicating that any existing ND cache entries should be
overwritten based on this NA message (see [tshark trace](https://gist.github.com/leblancd/e550d41cad5a1c0ecb81837dfbc93cc5)).

Fixes #370

### DIFF
--- a/pkg/ip/link.go
+++ b/pkg/ip/link.go
@@ -172,28 +172,52 @@ func DelLinkByNameAddr(ifName string, family int) (*net.IPNet, error) {
 	return addrs[0].IPNet, nil
 }
 
+// SetHWAddrByIP sets the hardware address of an interface/link. If an IPv4
+// address is provided (for IPv4-only or dual-stack operation), then the
+// that IPv4 address is assumed to be locally unique, and the hardware
+// address that is configured comprises the 4 bytes of IPv4 address preceded
+// by a hard-coded 2-byte prefix (0a:58). If only an IPv6 address is
+// provided, then the last 4 bytes of the interface/link's current hardware
+// address are assumed to be random, and the hardware address that is
+// configured is derived by overwriting the first two bytes of the
+// current hardware address with a different hard-coded prefix (6a:58).
 func SetHWAddrByIP(ifName string, ip4 net.IP, ip6 net.IP) error {
 	iface, err := netlink.LinkByName(ifName)
 	if err != nil {
 		return fmt.Errorf("failed to lookup %q: %v", ifName, err)
 	}
+	startHWAddr := iface.Attrs().HardwareAddr.String()
 
+	var hwAddr net.HardwareAddr
 	switch {
-	case ip4 == nil && ip6 == nil:
-		return fmt.Errorf("neither ip4 or ip6 specified")
-
 	case ip4 != nil:
-		{
-			hwAddr, err := hwaddr.GenerateHardwareAddr4(ip4, hwaddr.PrivateMACPrefix)
-			if err != nil {
-				return fmt.Errorf("failed to generate hardware addr: %v", err)
-			}
-			if err = netlink.LinkSetHardwareAddr(iface, hwAddr); err != nil {
-				return fmt.Errorf("failed to add hardware addr to %q: %v", ifName, err)
-			}
+		hwAddr, err = hwaddr.GenerateHardwareAddr4(ip4, hwaddr.PrivateMACPrefix)
+		if err != nil {
+			return fmt.Errorf("failed to generate hardware addr: %v", err)
 		}
 	case ip6 != nil:
-		// TODO: IPv6
+		attrs := iface.Attrs()
+		hwAddr, err = hwaddr.GenerateHardwareAddr6(attrs.HardwareAddr,
+			hwaddr.PrivateMACPrefix6)
+		if err != nil {
+			return fmt.Errorf("failed to generate hardware addr: %v", err)
+		}
+	default:
+		return fmt.Errorf("unable to generate hardware address as neither IPv4 nor IPv6 address specified")
+	}
+
+	if hwAddr.String() != startHWAddr {
+		// Toggle the interface while changing the hardware address
+		// so that a new IPv6 link local address gets generated.
+		if err := netlink.LinkSetDown(iface); err != nil {
+			return fmt.Errorf("failed to set down %q: %v", ifName, err)
+		}
+		if err = netlink.LinkSetHardwareAddr(iface, hwAddr); err != nil {
+			return fmt.Errorf("failed to add hardware addr to %q: %v", ifName, err)
+		}
+		if err := netlink.LinkSetUp(iface); err != nil {
+			return fmt.Errorf("failed to set up %q: %v", ifName, err)
+		}
 	}
 
 	return nil
--- a/pkg/utils/hwaddr/hwaddr.go
+++ b/pkg/utils/hwaddr/hwaddr.go
@@ -20,37 +20,38 @@ import (
 )
 
 const (
-	ipRelevantByteLen      = 4
-	PrivateMACPrefixString = "0a:58"
+	ipRelevantByteLen       = 4
+	PrivateMACPrefixString  = "0a:58"
+	PrivateMACPrefixString6 = "6a:58"
 )
 
 var (
 	// private mac prefix safe to use
-	PrivateMACPrefix = []byte{0x0a, 0x58}
+	PrivateMACPrefix  = []byte{0x0a, 0x58}
+	PrivateMACPrefix6 = []byte{0x6a, 0x58}
 )
 
-type SupportIp4OnlyErr struct{ msg string }
+type InvalidIP4Err struct{ msg string }
 
-func (e SupportIp4OnlyErr) Error() string { return e.msg }
-
-type MacParseErr struct{ msg string }
-
-func (e MacParseErr) Error() string { return e.msg }
+func (e InvalidIP4Err) Error() string { return e.msg }
 
 type InvalidPrefixLengthErr struct{ msg string }
 
 func (e InvalidPrefixLengthErr) Error() string { return e.msg }
 
-// GenerateHardwareAddr4 generates 48 bit virtual mac addresses based on the IP4 input.
+// GenerateHardwareAddr4 generates 48 bit virtual mac addresses based on
+// an IPv4 address.
 func GenerateHardwareAddr4(ip net.IP, prefix []byte) (net.HardwareAddr, error) {
 	switch {
 
 	case ip.To4() == nil:
-		return nil, SupportIp4OnlyErr{msg: "GenerateHardwareAddr4 only supports valid IPv4 address as input"}
+		return nil, InvalidIP4Err{msg: fmt.Sprintf(
+			"Invalid IPv4 address: %s", ip.String()),
+		}
 
 	case len(prefix) != len(PrivateMACPrefix):
 		return nil, InvalidPrefixLengthErr{msg: fmt.Sprintf(
-			"Prefix has length %d instead  of %d", len(prefix), len(PrivateMACPrefix)),
+			"Prefix has length %d instead of %d", len(prefix), len(PrivateMACPrefix)),
 		}
 	}
 
@@ -60,4 +61,20 @@ func GenerateHardwareAddr4(ip net.IP, prefix []byte) (net.HardwareAddr, error) {
 			prefix,
 			ip[ipByteLen-ipRelevantByteLen:ipByteLen]...),
 	), nil
+}
+
+// GenerateHardwareAddr6 generates 48 bit virtual mac address for an
+// interface by taking the existing, presumably random mac address
+// and overwriting the first 2 octets with an easily discernible
+// 2-byte hard-coded prefix (6a:58).
+func GenerateHardwareAddr6(mac net.HardwareAddr, prefix []byte) (net.HardwareAddr, error) {
+
+	if len(prefix) != len(PrivateMACPrefix) {
+		return nil, InvalidPrefixLengthErr{msg: fmt.Sprintf(
+			"Prefix has length %d instead of %d", len(prefix),
+			len(PrivateMACPrefix6)),
+		}
+	}
+
+	return (net.HardwareAddr)(append(prefix, mac[len(prefix):]...)), nil
 }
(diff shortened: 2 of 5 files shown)


pmichali:
Would another alternative be to hash the IPv6 address into a 32 bit value?

pkg/ip/link.go
```diff
@@ -172,28 +172,41 @@ func DelLinkByNameAddr(ifName string, family int) (*net.IPNet, error) {
...
-			hwAddr, err := hwaddr.GenerateHardwareAddr4(ip4, hwaddr.PrivateMACPrefix)
-			if err != nil {
-				return fmt.Errorf("failed to generate hardware addr: %v", err)
-			}
-			if err = netlink.LinkSetHardwareAddr(iface, hwAddr); err != nil {
-				return fmt.Errorf("failed to add hardware addr to %q: %v", ifName, err)
-			}
+		hwAddr, err = hwaddr.GenerateHardwareAddr4(ip4, hwaddr.PrivateMACPrefix)
+		if err != nil {
+			return fmt.Errorf("failed to generate hardware addr: %v", err)
```
pmichali: Do we want to provide more info "failed to generate hardware address from IPv4 address: %v"? Similar for L202

pkg/ip/link.go
```diff
@@ -172,28 +172,41 @@ func DelLinkByNameAddr(ifName string, family int) (*net.IPNet, error) {
...
 	case ip6 != nil:
-		// TODO: IPv6
+		attrs := iface.Attrs()
+		hwAddr, err = hwaddr.GenerateHardwareAddr6(attrs.HardwareAddr,
+			hwaddr.PrivateMACPrefix6)
+		if err != nil {
+			return fmt.Errorf("failed to generate hardware addr: %v", err)
+		}
+	default:
+		return fmt.Errorf("neither ip4 or ip6 specified")
```
pmichali: "unable to generate hardware address as neither IPv4 nor IPv6 address specified"?

pkg/utils/hwaddr/hwaddr.go
```diff
@@ -20,37 +20,39 @@ import (
...
-		return nil, SupportIp4OnlyErr{msg: "GenerateHardwareAddr4 only supports valid IPv4 address as input"}
+		return nil, InvalidIP4Err{msg: fmt.Sprintf(
+			"Invalid IPv4 address: %s", ip.String()),
+		}
 
 	case len(prefix) != len(PrivateMACPrefix):
 		return nil, InvalidPrefixLengthErr{msg: fmt.Sprintf(
-			"Prefix has length %d instead  of %d", len(prefix), len(PrivateMACPrefix)),
+			"Prefix has length %d instead  of %d", len(prefix),
+			len(PrivateMACPrefix)),
```
pmichali: Doesn't look like there is a change here (other than return added).

pkg/utils/hwaddr/hwaddr.go
```diff
@@ -61,3 +63,19 @@ func GenerateHardwareAddr4(ip net.IP, prefix []byte) (net.HardwareAddr, error) {
...
+
+// GenerateHardwareAddr6 generates 48 bit virtual mac address for an
+// interface by taking the existing, presumably random mac address
+// and overwriting the first 2 octets with an easily discernible
+// 2-byte hard-coded prefix (6a:58).
+func GenerateHardwareAddr6(mac net.HardwareAddr, prefix []byte) (net.HardwareAddr, error) {
+
+	if len(prefix) != len(PrivateMACPrefix) {
+		return nil, InvalidPrefixLengthErr{msg: fmt.Sprintf(
+			"Prefix has length %d instead  of %d", len(prefix),
```
pmichali: nit: extra space "instead  of"

pkg/utils/hwaddr/hwaddr.go
```diff
@@ -20,37 +20,39 @@ import (
...
 
 var (
 	// private mac prefix safe to use
-	PrivateMACPrefix = []byte{0x0a, 0x58}
+	PrivateMACPrefix  = []byte{0x0a, 0x58}
+	PrivateMACPrefix6 = []byte{0x6a, 0x58}
 )
 
-type SupportIp4OnlyErr struct{ msg string }
+type InvalidIP4Err struct{ msg string }
```
danehans: Should the name of line 34 and 36 be general for v4 and v6 or should a separate type and func be added for v6?

pkg/ip/link.go
leblancd: The %v in this case will expand to the specific error that was encountered in GenerateHardwareAddr4(). Same with L202.

pkg/utils/hwaddr/hwaddr.go
leblancd: The invalid IPv4 error only applies to IPv4, since the GenerateHardwareAddress6() function for IPv6 operates on the current MAC address, not the IPv6 address.

pkg/ip/link.go
leblancd: Yes, that would be clearer. Will change the error string.

pkg/utils/hwaddr/hwaddr.go
leblancd: Yes, looks like I got too ambitious with limiting line length. I'll revert this.

pkg/utils/hwaddr/hwaddr.go
leblancd: Will fix. There's another double space above that this was copied from as well.

pkg/ip/link.go
leblancd: The %v will expand to the specific error that was encountered in the call to GenerateHardwareAddr4().

squeed:
Hm. This will take some thought. The original justification for generating the MAC address from the IP address (as reported in #194 / #196 / #254) was that the arp cache interacted poorly with constantly cycling mac for a given IP (which is not unlikely when containers are rolling). Specifically, this is a problem with the bridge plugin. I'm curious if the nd-cache is any different in this respect.

One advantage of hashing the ipv6 address is that the mapping would be stable, preserving the nd-cache.

Incidentally, I'm not sure how SLAAC fits in the CNI world just yet. It doesn't work for every type of interface; only those such as macvlan would really work. However, this is a side discussion for now.

leblancd:
@squeed: Thank you for your explanation of why V4 MAC addresses are generated
based on V4 IP addresses! Who knew that assigning a MAC address would be this complicated? :) As strange as this might sound, I believe that an implication of this issue is that we'll need to impose a minimum subnet space (i.e. maximum prefix length) for IPv6 subnets that are allocated to a node (as I'll explain). My strong preference would be /64 as the minimum IPv6 subnet space (i.e. maximum prefix length of 64 bits), for reasons outlined in RFC7421. I believe that this would have an impact on PR #279.

As you described, I expect that that for IPv6, the ND cache would be subject to some thrashing/collision over time (similar to IPv4 case) as IPv6 addresses are recycled, but the degree or chance of collisions would depend on the size of each node's IPv6 subnet space. For example, if a node is assigned a /120 subnet, there will be only 255 possible container/pod IPv6 addresses that are being cycled through, so that IPv6 addresses would be recycled fairly quickly in a fast-paced, short-lived container environment. The problem is actually worse for dual-stack V4/V6 than it is for IPv6-only, especially considering the changes I've proposed in this pull request. In the dual-stack case, the MAC will be based on IPv4 address, so MAC<->IPv4 is deterministic, BUT the IPv6 addresses are assigned uncoupled/independently, so MAC<->IPv6 relationship is non-deterministic. For example, let's take a hypothetical case where, at some point in time, a container/pod is assigned the following
IPs/MAC:
    IPv4: 10.1.2.3
    IPv6: fd00::3
    MAC: 0a:58:0a:01:02:03
But at a later point in time, after IPv6 addresses start to get recycled and this container/pod gets deleted, another container/pod could be assigned:
    IPv4: 10.1.2.7
    IPv6: fd00::3                     < = = = Same IPv6 address as before
    MAC: 0a:58:0a:01:02:07   < = = = Different MAC address, based on a different IPv4 address
(Of course, IPv4 & IPv6 assignment might be "phase-locked" if their space sizes are identical, but we can't depend on that in all cases.) Depending on how fast or frequently the IPv6 container/pod addresses are recycled (i.e how big the IPv6 subnet space is), we could be inducing collisions or thrashing in IPv6 ND caches. For a /120 subnet space, this might happen fairly quickly.

So, how do we solve this dilemma for dual-stack? I can think of 2 main ways to tackle the problem:
- **Force the IPv6 subnet space to be sufficiently large**: If we make the IPv6 subnet space that is assigned to each node sufficiently large (e.g. /80, or better yet, /64), AND continue to assign container/pods IPv6 addresses in a round-robin fashion (independent from IPv4/MAC address assignments), then the chance of collisions should be reduced. However, one potential problem with this approach is that once IPv6 container/pod addresses have wrapped and we're reusing IPv6 addresses, then the risk of collision is heightened, especially if we have a mix of long-lived and short-lived containers/pods. For example, let's say we have a long-lived container/pod that's using the IP address fd00::3, and that container/pod gets deleted right as we're cycling through the beginning of the subnet space, then the next container/pod could be assigned that same IPv6 address with a different IPv4/MAC. Of course, for a /64 subnet, it would take a long time for wrapping to occur, so it's sort of a "ticking time bomb" with a long fuse.
- **Generate dual-stack IPv6 addresses based on the IPv4/MAC**: For the dual-stack case, if we take the MAC that has been generated based on an IPv4 address, and then generate an IPv6 address from that, then the IPv4 <-> MAC <-> IPv6 will all be deterministic. HOWEVER, what this means is that the interface ID used for the IPv6 subnet must be 64 bits (and therefore IPv6 subnet must be at least as large as a /64). According to RFC7136:
      "For all unicast addresses, except those that start with the binary
      value 000, Interface IDs are required to be 64 bits long.  If
      derived from an IEEE MAC-layer address, they must be constructed
      in Modified EUI-64 format."
As stated in RFC7421:
    "...using an IID shorter than 64 bits and a subnet prefix longer than 64 bits is outside the current
     IPv6 specifications, so results may vary."
So in order to conform to IPv6 standards, we would need to impose a minimum subnet space size of /64 if we generate IPv6 addresses based on a MAC that's derived from an IPv4 address.

Even considering the IPv6-only case separately, we'd still need to consider imposing a minimum subnet size (maximum subnet bit length) of /64. If the choice is to generate a MAC address by hashing an IPv6 address, then I *think* we'd want to go with the hash algorithm specified in RFC5535, and this depends on the interface identifier being 64 bits long.

So what I would propose is this:
- Impose a minimum subnet space of /64 (maximum 64 bits of prefix) for IPv6 subnets assigned to nodes. (If we have to impose a limit anyway, let's bite the bullet and make it 64 bits).
- For  dual-stack V4/V6, continue to assign MAC addresses based on the IPv4 address, and then generate/derive an IPv6 address from that MAC address using a 64-bit EUI-64 interface identifier (since EUI-64 seems to be the defacto standard for generating interface ID from a MAC).
- For IPv6-only, first assign a MAC randomly, and then generate an IPv6 address within the node's IPv6 subnet space using a 64-bit EUI-64 interface identifier.

Using 64 bits for interface identifier might sound wasteful, but it's really not in the IPv6 world where you have so many bits to use to carve a node subnet out of a cluster subnet space (especially for private, or ULA addresses).
One benefit to imposing this restriction would be that in the future, we would be able to implement the IPv6 features listed in Section 4.1 of RFC7421 that depend on a 64-bit interface ID, e.g.:
- Teredo (RFC4380)
- 6to4 (RFC3056)
- NPTv6 (RFC6296)
I don't know if these specific features would be of interest or value, but the bottom line is that supporting a 64-bit interface identifier would make the code more future-proof in IPv6 land.

leblancd:
@squeed: I now believe that we will NOT have any issue with IPv6 Neighbor Discovery (ND) cache "thrashing" due to IPv6 unicast addresses getting recycled (as we see with IPv4 ARP cache with IPv4 address recycling). I've discussed this with one of our IPv6 experts, and I've done some experimenting to convince myself that IPv6 address recycling will not be an issue. The reason that IPv6 ND cache is resilient to IPv6 unicast address re-use is that when a node sends a Neighbor Advertisement (NA) packet, saying that it owns a certain IPv6 unicast address, it always sets the "ovr" (or "O-bit", see RFC4861) override flag. This flag indicates that this NA should override any existing ND cache entry. To see this in action, check out this Gist: [tshark output](https://gist.github.com/leblancd/e550d41cad5a1c0ecb81837dfbc93cc5).

rosenhouse:
The plugins (and many of their supporting packages like `ip` and `utils`) have moved into a [separate repo here](https://github.com/containernetworking/plugins).  If you'd like to continue development, please open a new PR there.

Sorry for the churn.

Issue #395, opened 2017-03-09
Vendoring cni and netlink


jaydunk:
We're writing a CNI plugin and running into a few issues around vendoring `containernetworking/cni` and netlink.

First is that running the normal `go build` fails due to multiple definitions of `vishvananda/netlink`:
```
/usr/local/go/pkg/tool/linux_amd64/link: running gcc failed: exit status 1
/tmp/go-link-088250716/000002.o: In function `_cgo_f36fc5609ae4_C2func_load_simple_bpf':
veth/go/src/github.com/cloudfoundry-incubator/silk/vendor/github.com/containernetworking/cni/vendor/github.com/vishvananda/netlink/bpf_linux.go:77: multiple definition of `_cgo_f36fc5609ae4_C2func_load_simple_bpf'
veth/tmp/go-link-088250716/000000.o:/go/src/github.com/cloudfoundry-incubator/silk/vendor/github.com/vishvananda/netlink/bpf_linux.go:77: first defined here
```

We are able to work around this by adding `-ldflags="-extldflags=-Wl,--allow-multiple-definition"` to our build commands.

A bigger issue comes up when trying to use the existing cni libraries while also vendoring in netlink. I have a function for creating a veth pair and returns the pair back to the caller. I would like to leverage the `SetupVeth` function in `pkg/ip`. For example:

```
	var hostVeth, containerVeth netlink.Link
	err = containerNS.Do(func(_ ns.NetNS) error {
		hostLink, containerLink, err := ip.SetupVeth(ifname, mtu, hostNS)
		if err != nil {
			return err
		}
		return nil
	})

```

This fails to compile since `hostVeth` and `containerVeth` are of type `netlink.Link` as defined in `vendor/github.com/vishvananda/netlink` but `SetupVeth` returns `netlink.Link` as defined in `vendor/github.com/containernetworking/cni/vendor/github.com/vishvananda/netlink`.

cc: @rosenhouse

bboreham:
You really have to use a vendoring technique that flattens dependencies such that there is only one copy of each library, and they're all at top level. 

Tools like `dep`, `gvt` and `glide` will do this for you. 

Ideally no library repo would vendor any dependencies, but this repo is both a library and a source of binaries.

dcbw:
[...]

PR #396, opened 2017-03-09
pkg/ip: do not leak types from vendored netlink package

jaydunk:
The exported function SetupVeth now returns a package-defined type.

Addresses #395.

cc: @rosenhouse

### DIFF
--- a/pkg/ip/link.go
+++ b/pkg/ip/link.go
@@ -98,30 +98,49 @@ func RenameLink(curName, newName string) error {
 	return err
 }
 
+type LinkAttrs struct {
+	Name         string
+	HardwareAddr net.HardwareAddr
+	Index        int
+}
+
+type link struct {
+	netlink.Link
+}
+
+func (l *link) Attrs() LinkAttrs {
+	a := l.Link.Attrs()
+	return LinkAttrs{
+		Name:         a.Name,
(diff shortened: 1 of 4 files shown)


bboreham:
It seems to me you could use `net.Interface` instead of defining another local type.

rosenhouse:
@bboreham Updated!

rosenhouse:
@containernetworking/cni-maintainers thoughts?

bboreham:
I guess it's fine; lines like `return net.Interface{}, net.Interface{}, err` seem ugly but not overwhelmingly so.

How do we feel about the impact on people who were already using this interface?

rosenhouse:
They might need to change a few lines, like [we did for the PtP plugin](https://github.com/containernetworking/cni/pull/396/files#diff-1c0526e7387af989eda0ab95196ccfd7L70).    It's pretty small.

I'd love for us to work towards extracting a clean, stable, public API for plugin developers.   What's in the `pkg` directory *today* is not there yet.

rosenhouse:
@bboreham During the community sync meeting today, there didn't seem to be any objections to these types of changes to the `pkg` packages.  Are you ok with a merge?

dcbw:
Looks OK to me.

bboreham:
No objection from me

PR #400, opened 2017-03-15
Rename build script to avoid conflict with bazel

lstoll:
It's difficult to include this repository using bazel, because
the file named "build" conflicts with new_go_repository generation
on case-insensitive filesystems (ref
https://github.com/bazelbuild/rules_go/issues/234). This change
renames the file to something that doesn't conflict, and also
renames the test script for consistency.

I felt like I'd open this as a proposal to see how people felt, I wouldn't be concerned if it was rejected. I felt with kubernetes adopting bazel this might come up for others at some point.


dcbw:
Seems fine to me.

PR #402, opened 2017-03-15
readme: remove community sync meeting notice

rosenhouse:
will add back when we schedule the next one


jonboulle:
how about keeping a message with the intention and then just saying people should follow the mailing list for date announcements?

PR #406, opened 2017-03-16
Return cni.type.Error instead of plain error

GheRivero:
This way, we can use Error.Code attribute to check error type without
needing to regexp the message string

kubernetes/kubernetes#43014

### DIFF
--- a/pkg/invoke/raw_exec.go
+++ b/pkg/invoke/raw_exec.go
@@ -50,13 +50,9 @@ func pluginErr(err error, output []byte) error {
 	if _, ok := err.(*exec.ExitError); ok {
 		emsg := types.Error{}
 		if perr := json.Unmarshal(output, &emsg); perr != nil {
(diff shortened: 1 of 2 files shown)


pkg/invoke/raw_exec.go
```diff
@@ -52,11 +52,7 @@ func pluginErr(err error, output []byte) error {
 		if perr := json.Unmarshal(output, &emsg); perr != nil {
 			return fmt.Errorf("netplugin failed but error parsing its diagnostic message %q: %v", string(output), perr)
```
squeed: Do we want to generate a `types.Error` for this condition as well? Then we can tighten up the return types.

GheRivero: Sure!

dcbw:
LGTM after the change Casey requested.

GheRivero:
Commit updated with suggestion

squeed:
LGTM

PR #407, opened 2017-03-17
plugins/meta/flannel: If net config is missing do not return err on DEL

aaronlevy:
Based on the CNI spec, if resources are missing we should not always return error (https://github.com/containernetworking/cni/blob/master/SPEC.md#network-configuration-list-error-handling).

An open question on my end would be if that delegates would still need to be called. My thinking is that if we can't read the config - all information that would have been passed down doesn't exist - so we may as well exit early.

Additionally, other plugins may need to be checked for the same conformance against DEL spec expectations.

### DIFF
--- a/plugins/meta/flannel/flannel.go
+++ b/plugins/meta/flannel/flannel.go
@@ -142,6 +142,7 @@ func saveScratchNetConf(containerID, dataDir string, netconf []byte) error {
 
 func consumeScratchNetConf(containerID, dataDir string) ([]byte, error) {
 	path := filepath.Join(dataDir, containerID)
+	// Ignore errors when removing - Per spec safe to continue during DEL
 	defer os.Remove(path)
 
 	return ioutil.ReadFile(path)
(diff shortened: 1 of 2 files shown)


squeed:
Hmm. I'd like to see the file deleted only if deletion was sucessful. It looks like `consumeScratchNetConf` *always* deletes the file, which seems a bit odd...

GheRivero:
/lgtm
Could we check this at a higher level? That way, all plugins will follow the spec (related #406)

aaronlevy:
Also to address your comment @squeed - spoke with @tomdee offline - and moving the deletion until after successfully decoding the json (only step after `consumeScratchNetConf`) would mean we could get into failure loops because the bad json is never cleaned up - and that seemed less ideal. 

Left behavior as-is for now (with a comment regarding any errors from deletion).

dcbw:
LGTM; I guess the plugin could re-render the scratch netconf on cmdDel too since all the information is still there (like the flannel .env file), except that info could have changed no longer generate the same scratch netconf.

PR #408, opened 2017-03-17
Added current cni spec version to error code

GheRivero:
According to the CNI 3.0 spec:

'''
Errors are indicated by a non-zero return code and the following JSON being printed to stdout:

{
  "cniVersion": "0.2.0",
  "code": <numeric-error-code>,
  "msg": <short-error-message>,
  "details": <long-error-message> (optional)
}
cniVersion specifies a Semantic Version 2.0 of CNI specification used by the plugin.
'''

We were omitting it. Use the current version and fix the required tests to pas.

### DIFF
--- a/pkg/skel/skel.go
+++ b/pkg/skel/skel.go
@@ -141,8 +141,9 @@ func (t *dispatcher) getCmdArgsFromEnv() (string, *CmdArgs, error) {
 
 func createTypedError(f string, args ...interface{}) *types.Error {
 	return &types.Error{
-		Code: 100,
-		Msg:  fmt.Sprintf(f, args...),
+		CniVersion: version.Current(),
+		Code:       100,
+		Msg:        fmt.Sprintf(f, args...),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


dcbw:
The spec says that plugins should return a Result in the same version as their input JSON, which I guess I'd expect from errors too.  But this PR looks like it would always return the latest spec version that the binaries were built with.

Would it be too hard to use the input spec version instead?

pkg/skel/skel.go
```diff
@@ -180,11 +181,21 @@ func (t *dispatcher) pluginMain(cmdAdd, cmdDel func(_ *CmdArgs) error, versionIn
...
 		if e, ok := err.(*types.Error); ok {
 			// don't wrap Error in Error
+			if e.CniVersion == "" {
+				e.CniVersion = configVersion
+			}
 			return e
 		}
-		return createTypedError(err.Error())
+
+		e := createTypedError(err.Error())
```
dcbw: Could we just have createTypedError take a version string which it then sets in the error?

pkg/skel/skel_test.go
```diff
@@ -93,8 +93,9 @@ var _ = Describe("dispatching to the correct callback", func() {
 		err := dispatch.pluginMain(cmdAdd.Func, cmdDel.Func, versionInfo)
 		if isRequired {
 			Expect(err).To(Equal(&types.Error{
-				Code: 100,
-				Msg:  "required env variables missing",
+				CniVersion: version.Current(),
+				Code:       100,
+				Msg:        "required env variables missing",
```
dcbw: Could we also add a testcase that sends an old CNI spec version and ensure that the returned error reports that old version?

bboreham:
Closing stale PR.

PR #409, opened 2017-03-18
Ignore errors when deleting a non existant resource

GheRivero:
Plugins should generally complete a DEL action without error even if some resources are missing. This is plugin agnostic.


GheRivero:
do not commit

PR #411, opened 2017-03-20
vlan: add VLAN plugin

dcbw:
Add a pretty simple plugin that just creates a VLAN interface in the given namespace off a given master.

### DIFF
--- a/plugins/main/vlan/vlan.go
+++ b/plugins/main/vlan/vlan.go
@@ -1,0 +1,190 @@
+// Copyright 2015 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


dcbw:

bboreham:
Should this go in https://github.com/containernetworking/plugins ?

dcbw:
@bboreham I must have missed that, are we generally moving plugins over there now?

bboreham:
It was discussed at the Community Sync meeting; we have three things that ought to release on different schedules: the spec, the library and the plugins.  I don't think anyone's fired up to move the existing plugins, but it seems like a good idea to put new ones there.

Maybe we could just discuss whether we think this particular plugin should be forever maintained by the CNI maintainers before you go ahead and open that PR, however.

plugins/main/vlan/vlan.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	if err := netlink.LinkAdd(v); err != nil {
+		return nil, fmt.Errorf("failed to create vlan: %v", err)
+	}
+
+	err = netns.Do(func(_ ns.NetNS) error {
+		err := ip.RenameLink(tmpName, ifName)
+		if err != nil {
+			return fmt.Errorf("failed to rename vlan to %q: %v", ifName, err)
```
squeed: Probably worth trying to clean up the failed link here too.

dcbw:
Closed in favor of https://github.com/containernetworking/plugins/pull/2

PR #412, opened 2017-03-21
plugins/*: Don't error if the device doesn't exist

tomdee:
I wasn't able to test or updat the dhcp plugin. All the others are
tested and fixed

### DIFF
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -386,25 +386,30 @@ func cmdDel(args *skel.CmdArgs) error {
 		return nil
 	}
 
+	// There is a netns so try to clean up. Delete can be called multiple times
+	// so don't return an error if the device is already removed.
+	// If the device isn't there then don't try to clean up IP masq either.
 	var ipn *net.IPNet
 	err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
 		var err error
 		ipn, err = ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
+		if err != nil && err == ip.ErrLinkNotFound {
+			return nil
+		}
 		return err
 	})
+
 	if err != nil {
 		return err
 	}
 
-	if n.IPMasq {
+	if ipn != nil && n.IPMasq {
 		chain := utils.FormatChainName(n.Name, args.ContainerID)
 		comment := utils.FormatComment(n.Name, args.ContainerID)
-		if err = ip.TeardownIPMasq(ipn, chain, comment); err != nil {
-			return err
-		}
+		err = ip.TeardownIPMasq(ipn, chain, comment)
 	}
 
-	return nil
+	return err
 }
 
 func main() {
--- a/plugins/main/ipvlan/ipvlan.go
+++ b/plugins/main/ipvlan/ipvlan.go
@@ -191,9 +191,18 @@ func cmdDel(args *skel.CmdArgs) error {
 		return nil
 	}
 
-	return ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
-		return ip.DelLinkByName(args.IfName)
+	// There is a netns so try to clean up. Delete can be called multiple times
+	// so don't return an error if the device is already removed.
+	err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+		if _, err := ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4); err != nil {
+			if err != ip.ErrLinkNotFound {
+				return err
+			}
+		}
+		return nil
 	})
+
+	return err
 }
 
 func main() {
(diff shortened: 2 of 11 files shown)


plugins/main/bridge/bridge.go
```diff
@@ -386,14 +386,23 @@ func cmdDel(args *skel.CmdArgs) error {
...
+			var err error
+			ipn, err = ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
+			return err
+		})
+		if err != nil {
+			return err
+		}
 	}
 
 	if n.IPMasq {
```
aaronlevy: If the iptables operations below also error - I would assume in some cases we would want to also ignore (e.g. chain does not exist?)

plugins/main/ptp/ptp.go
```diff
@@ -268,14 +268,24 @@ func cmdDel(args *skel.CmdArgs) error {
...
+			var err error
+			ipn, err = ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
+			return err
+		})
+		if err != nil {
+			return err
+		}
 	}
 
 	if conf.IPMasq {
```
aaronlevy: Same question about errors here.

aaronlevy:
Does host-local/ipam also need changes or is that unaffected?

Unfortunately - I'm not super familiar with the CNI plugins - so could only give a quick scan. But only questions, no objections to these changes.

rosenhouse:
Is this something we want to have test coverage for?

plugins/main/ipvlan/ipvlan.go
```diff
@@ -191,9 +191,20 @@ func cmdDel(args *skel.CmdArgs) error {
...
-	return ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
-		return ip.DelLinkByName(args.IfName)
+	// Check whether the device exists before trying to delete it
+	devErr := ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
+		_, err := netlink.LinkByName(args.IfName)
+		return err
 	})
+
+	// Only try to delete the device if it could be found in the first place.
+	if devErr == nil {
```
calebamiles: just a silly question is there any reason to not bail out if `devErr != nil`  rather than continuing if `devErr == nil`. maybe with a comment about the device not existing/not being found is not an exceptional error which should be returned

plugins/main/bridge/bridge.go
```diff
@@ -386,25 +387,25 @@ func cmdDel(args *skel.CmdArgs) error {
...
-		return err
-	}
-
-	if n.IPMasq {
-		chain := utils.FormatChainName(n.Name, args.ContainerID)
-		comment := utils.FormatComment(n.Name, args.ContainerID)
-		if err = ip.TeardownIPMasq(ipn, chain, comment); err != nil {
-			return err
+		if ipn, err := ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4); err != nil {
+			if !strings.Contains(err.Error(), "Link not found") {
```
liggitt: is this text stable across versions and machine locales?

calebamiles: you beat me to the :sob:

plugins/main/bridge/bridge.go
```diff
@@ -386,25 +387,32 @@ func cmdDel(args *skel.CmdArgs) error {
...
 
+	// There is a netns so try to clean up. Delete can be called multiple times
+	// so don't return an error if the delete is already removed.
+	// If the device isn't there then don't try to clean up IP masq either.
 	var ipn *net.IPNet
 	err = ns.WithNetNSPath(args.Netns, func(_ ns.NetNS) error {
 		var err error
 		ipn, err = ip.DelLinkByNameAddr(args.IfName, netlink.FAMILY_V4)
+		if err != nil && strings.Contains(err.Error(), "Link not found") {
+			// Ignore the error if link isn't found.
```
calebamiles: possibly unnecessary given apparent conformance to return error or value but not both paradigm of [`DelLinkByNameAddr`](https://github.com/containernetworking/cni/blob/master/pkg/ip/link.go#L168)

calebamiles: you'll probably want to add a unit test to address the fact that you're relying on a string returned by a dependency

calebamiles:
@tomdee looking at the changes I'm wondering, again naively, if you might get more mileage from making a change to `DelLinkByNameAddr` to not return an error if the name doesn't exist rather than duplicating the same check across multiple files. Or maybe creating a new utility function which does that if you're concerned about other consumers of `DelLinkByNameAddr`

tomdee:
OK, it should now have test coverage for all changes and a test has been added for host-local too (even though there were no changes there). I also added a test for the for the error string that we're relying on.

I don't feel too strongly about the code duplication - I'm calling it for the night so won't be able to fix it today. I think it's probably OK to go in as-is, but I'm happy to change it tomorrow if desired?

tomdee:
OK, tests are passing. @rosenhouse has approved and I've address all of @dcbw's feedback so merging.

dcbw:
LGTM FWIW

PR #413, opened 2017-03-22
spec/plugins: fix 'ip'->'ips' in the spec, bump to 0.3.1

dcbw:
Fix a spec omission in https://github.com/containernetworking/cni/pull/145 that forgot to change on instance of 'ip' to 'ips', bump the spec version and update all the plugins to the new spec.  Adds a single testcase for the bridge plugin to ensure compatibility between 0.3.0 and 0.3.1.


bboreham:
Should there be some comment about the difference between 0.3.0 and 0.3.1?

dcbw:
> Should there be some comment about the difference between 0.3.0 and 0.3.1?

@likely there should be, yes

dcbw:
@tomdee @rosenhouse @bboreham does the change to VERSIONING.md look OK?  We don't really have a changelog anywhere else so I stuck it there.

PR #414, opened 2017-03-22
Enable s390x build

kad:
This will help potentially to use cni releases directly in Kubernetes.


kad:
travis check failed for first version. 1.6.x doesn’t have s390x build target. will fix it.

squeed:
LGTM

Issue #10, opened 2015-05-06
Change the plugin model to be job agnostic


eyakubovich:
I am creating this issue to summarize @thockin's comments on https://github.com/appc/cni/pull/2. In particular, Tim's proposed to make the spec less opinionated. Current spec defines a network as a group of containers that are able to talk to each other. The user defines which networks a given container needs to join. The container runtime executes a plugin for each network and the job of said plugin is to join the container to the network. As such, there's an assumption that a plugin will create an interface, assign it an IP, etc.

Tim proposes to have the container runtime call out to a plugin (or maybe multiple) and the plugin can do whatever it wants. For example, you can have a plugin to setup IP Masquerade or to block some traffic or change an IP of an existing interface.

Tim's proposal is quite flexible but also a fundamental shift from the current spec. I'd like to discuss this further and also schedule a hangout for Fri or next week: http://doodle.com/arqu4y9rgt9ybfp3

steveeJ:
The proposal seems very vague and personally for me not enough to form an opinion about it. What would help is a comparison of the current situation and the changes the proposal would make to it.

eyakubovich:
Hangout on Tue May 12th, at 2:00PM Pacific Time. 	https://plus.google.com/hangouts/_/coreos.com/cni-model

jainvipin:
@eyakubovich - I will be there in the hangout. While it might seem vague to some, I think if you are allowing a plugin to do whatever it likes for network, I like @thokin's idea, leave it to the plugin with the right hooks (call them before container is launched, etc.) and then there full liberty/flexibility given to the plugin - as the network intent is driven from it. This is what was proposed as hook for Kubernetes and works fine with for my use cases.

eyakubovich:
Thank @jonboulle for taking minutes: https://docs.google.com/document/d/1RRgRhSM2_r68XZuJfOhU3sUR99lnOzRWyCAE8-HTVBA

eyakubovich:
I think the basic takeaways from this meeting were:
- There's much conceptual overlap with CNM/libnetwork -- we should explore collaboration
- Having a plugin do a particular job of plugging container into a network is OK
- Prescribing interface name to the plugin is excessive
- There's a desire to introduce Network Create event. However, it is still unclear if this event needs to be triggered on a single node in the cluster or on every node.
- @jainvipin will file an issue with proposed policy based straw man.

Symmetric:
[...]

PR #416, opened 2017-03-23
bridge: Add support for IPv6 to bridge plugin

### LINKED ISSUE #415: Bridge plugin needs support for IPv6 addresses
Support needs to be added to the bridge plugin in order for it to support dual-stack and
IPv6-only network configurations.

Changes to the bridge plugin will also depend upon changes proposed in PR #279 and PR #394
in order to exercise/test IPv6 functionality.

leblancd:
This change adds support for IPv6 container/pod addresses to the
CNI bridge plugin, both for dual-stack (IPv4 + IPv6) and for IPv6-only
network configurations.

This change could potentially be committed independently from other CNI
IPv6 changes, however the IPv6 functionality that is enabled by this change
can't be exercised/tested without the commits of PR #279 and PR #394.

There are some IPv6 unit test cases that are currently commented out
in the proposed changes because these test cases will fail without the
prior commits of PR #279 and PR #394, however, these test cases have been
tested with patches for PR#279 and PR #394 applied, so they can be
uncommented (enabled) after those dependent PRs are committed.

The proposed changes support multiple IPv6 addresses on a container
interface. If isGW is configured, the bridge will also be configured with
gateway addresses for each IPv6 subnet.

The proposed changes do not support the case where a dual-stack
network configuration specifies different IPAM plugins for IPv4 vs. IPv6.
Supporting such a configuration would require separately marshaling
the IPv4 and IPv6 IPAM config into their own stdin (byte stream) formats,
then calling the IPAM plugins in sequence, and then merging the results.
This wouldn't be terribly complicated, but it's left for a followup PR.

This pull request includes a temporary workaround for Kubernetes
Issue #32291 (Container IPv6 address is marked as duplicate (dadfailed).
The problem is that kubelet enables hairpin mode on bridge veth
interfaces. Hairpin mode causes the container/pod to see echos of its
IPv6 neighbor solicitation packets, so that it declares duplicate address
detection (DAD) failure. The long-term fix is to use enhanced-DAD
when that feature is readily available in kernels. The short-term fix is
to disable IPv6 DAD in the container. Unfortunately, this has to be done
unconditionally (i.e. without a check for whether hairpin mode is enabled)
because hairpin mode is turned on by kubelet after the CNI bridge plugin
has completed cmdAdd processing. Disabling DAD should be okay if
IPv6 addresses are guaranteed to be unique (which is the case for
host-local IPAM plugin).

Fixes #415

### DIFF
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -66,6 +66,9 @@ type NetConf struct {
 	IPAM         struct {
 		Type string `json:"type,omitempty"`
 	} `json:"ipam,omitempty"`
+	IPAM6 struct {
+		Type string `json:"type,omitempty"`
+	} `json:"ipam6,omitempty"`
 	DNS DNS `json:"dns"`
 }
 
--- a/plugins/main/bridge/bridge.go
+++ b/plugins/main/bridge/bridge.go
@@ -22,6 +22,8 @@ import (
 	"runtime"
 	"syscall"
 
+	"io/ioutil"
+
 	"github.com/containernetworking/cni/pkg/ip"
 	"github.com/containernetworking/cni/pkg/ipam"
 	"github.com/containernetworking/cni/pkg/ns"
@@ -46,6 +48,12 @@ type NetConf struct {
 	HairpinMode  bool   `json:"hairpinMode"`
 }
 
+type addrGWs struct {
+	addr   net.IP
+	gws    []net.IPNet
+	family int
+}
+
 func init() {
 	// this ensures that main runs only on main thread (thread group leader).
 	// since namespace ops (unshare, setns) are done for a single thread, we
@@ -63,8 +71,86 @@ func loadNetConf(bytes []byte) (*NetConf, string, error) {
 	return n, n.CNIVersion, nil
 }
 
-func ensureBridgeAddr(br *netlink.Bridge, ipn *net.IPNet, forceAddress bool) error {
-	addrs, err := netlink.AddrList(br, syscall.AF_INET)
+// firstAddrGWs processes the results from the IPAM plugin, and
+// determines the following:
+//    - The first IPv4 and IPv6 container addresses. These addresses are
+//      used to determine the hardware address for the container interface
+//    - Default route(s)
+//    - Gateway address(es)
+func firstAddrGWs(result *current.Result, n *NetConf) (*addrGWs,
+	*addrGWs, error) {
+
+	aGWV4 := &addrGWs{}
+	aGWV6 := &addrGWs{}
+
+	for _, ipc := range result.IPs {
+
+		// Determine if this config is IPv4 or IPv6
+		var aGW *addrGWs
+		defaultNet := &net.IPNet{}
+		switch {
+		case ipc.Address.IP.To4() != nil:
+			aGW = aGWV4
+			aGW.family = netlink.FAMILY_V4
+			defaultNet.IP = net.IPv4zero
+		case len(ipc.Address.IP) == net.IPv6len:
+			aGW = aGWV6
+			aGW.family = netlink.FAMILY_V6
+			defaultNet.IP = net.IPv6zero
+		default:
+			return nil, nil, fmt.Errorf("Unknown IP object: %v", ipc)
+		}
+
+		// If this is the first configured address for this family, save it.
+		// This will be used to generate a MAC address.
+		if aGW.addr == nil {
+			aGW.addr = ipc.Address.IP
+		}
+
+		// Calculate gateway address corresponding to the selected IP address
+		ipc.Interface = 2
+		if ipc.Gateway == nil && n.IsGW {
+			ipc.Gateway = calcGatewayIP(&ipc.Address)
+		}
+
+		// Add the default route for this family if requested
+		if n.IsDefaultGW {
+			defaultNet.Mask = net.IPMask(defaultNet.IP)
+			defaultRouteFound := false
+			for _, route := range result.Routes {
+				if route.GW != nil && defaultNet.String() == route.Dst.String() {
+					defaultRouteFound = true
+				}
+			}
+			if !defaultRouteFound {
+				result.Routes = append(
+					result.Routes,
+					&types.Route{Dst: *defaultNet, GW: ipc.Gateway},
+				)
+			}
+		}
+
+		// Add the gateway address if requested
+		if n.IsGW {
+			gw := net.IPNet{
+				IP:   ipc.Gateway,
+				Mask: ipc.Address.Mask,
+			}
+			aGW.gws = append(aGW.gws, gw)
+		}
+	}
+	return aGWV4, aGWV6, nil
+}
+
+func firstGW(aGW *addrGWs) net.IP {
+	if aGW.gws != nil {
+		return aGW.gws[0].IP
+	}
+	return nil
+}
+
+func ensureBridgeAddr(br *netlink.Bridge, family int, ipn *net.IPNet, forceAddress bool) error {
+	addrs, err := netlink.AddrList(br, family)
 	if err != nil && err != syscall.ENOENT {
 		return fmt.Errorf("could not get list of IP addresses: %v", err)
 	}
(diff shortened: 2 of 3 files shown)


pkg/types/types.go
```diff
@@ -66,6 +66,9 @@ type NetConf struct {
 	IPAM         struct {
 		Type string `json:"type,omitempty"`
 	} `json:"ipam,omitempty"`
+	IPAM6 struct {
+		Type string `json:"type,omitempty"`
+	} `json:"ipam6,omitempty"`
```
dcbw: Seems OK, but I wonder if the same ipam plugin might want to coordinate IPv4 and IPv6 addresses for a container, and having separate blocks precludes that.

plugins/main/bridge/bridge.go
```diff
@@ -96,19 +103,23 @@ func ensureBridgeAddr(br *netlink.Bridge, ipn *net.IPNet, forceAddress bool) err
...
-func deleteBridgeAddr(br *netlink.Bridge, ipn *net.IPNet) error {
+func deleteBridgeAddr(br *netlink.Bridge, ipn *net.IPNet, isIPV4 bool) error {
 	addr := &netlink.Addr{IPNet: ipn, Label: ""}
 
-	if err := netlink.LinkSetDown(br); err != nil {
-		return fmt.Errorf("could not set down bridge %q: %v", br.Name, err)
+	if isIPV4 {
+		if err := netlink.LinkSetDown(br); err != nil {
+			return fmt.Errorf("could not set down bridge %q: %v", br.Name, err)
+		}
```
dcbw: Didn't see this before.  I know you're not adding it, but I really wonder why we bother with up/down.  I'm pretty sure you can remove an address without having to down the bridge...  @steveeJ any idea?

plugins/main/bridge/bridge.go
```diff
@@ -242,8 +253,21 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+	// same for IPv4 vs. IPv6. To allow for different plugin types,
+	// the IPv4 and IPv6 IPAM config will need to be split up and
+	// passed to each plugin separately, with results being combined.
+	var plugin_type string
+	switch {
+	case n.IPAM.Type != "":
+		plugin_type = n.IPAM.Type
+	default:
+		plugin_type = n.IPAM6.Type
+	}
```
dcbw: Hmm, why wouldn't we want to just run both IPv4 and IPv6 IPAM in sequence and then merge the results?

plugins/main/bridge/bridge.go
```diff
@@ -260,12 +284,20 @@ func cmdAdd(args *skel.CmdArgs) error {
...
 	for _, ipc := range result.IPs {
 		// All IPs currently refer to the container interface
 		ipc.Interface = 2
 		if ipc.Gateway == nil && n.IsGW {
 			ipc.Gateway = calcGatewayIP(&ipc.Address)
 		}
+		switch {
+		case firstV4Addr == nil && ipc.Address.IP.To4() != nil:
+			firstV4Addr = ipc.Address.IP
+		case firstV6Addr == nil && len(ipc.Address.IP) == net.IPv6len:
```
dcbw: I'm not sure you can use length as a check here, since a V4 address can be 16 bytes long in Go too, hence the need for To4() everywhere.  Perhaps grab and use canonicalizeIP() from plugins/ipam/host-local/backend/allocator/allocator.go and then you can be sure?

plugins/main/bridge/bridge.go
```diff
@@ -323,30 +355,40 @@ func cmdAdd(args *skel.CmdArgs) error {
...
+				}
+			case len(ipc.Gateway) == net.IPv6len && firstV6GWAddr == nil:
+				firstV6GWAddr = ipc.Gateway
+				err = ensureBridgeAddr(br, netlink.FAMILY_V6, gwn, n.ForceAddress)
+				if err != nil {
+					return fmt.Errorf("failed to set bridge addr: %v", err)
+				}
+				if err := ip.EnableIP6Forward(); err != nil {
+					return fmt.Errorf("failed to enable forwarding: %v", err)
+				}
```
dcbw: Ugh, I feel like there's gotta be a way to genericize these two cases.  I didn't look, but perhaps we can make ensureBridgeAddr() detect the family based on length, as long as we ensure addresses are canonicalized to 4 or 16 byte sizes?

plugins/main/bridge/bridge.go
```diff
@@ -378,7 +420,18 @@ func cmdDel(args *skel.CmdArgs) error {
...
+	// Determine the plugin type
+	// TODO: Add logic to support dual-stack config with different
+	// plugin types for IPv4 vs. IPv6.
+	var plugin_type string
+	switch {
+	case n.IPAM.Type != "":
+		plugin_type = n.IPAM.Type
+	default:
+		plugin_type = n.IPAM6.Type
+	}
```
dcbw: Same as above, why aren't we running these in sequence?

pkg/types/types.go
squeed: If we're going to change the spec, I was thinking of making IPAM a list...

leblancd: I agree, it would be better to have one ipam config block that contains a list of addresses. The changes that I've proposed are based on PR #279 (see this line: https://github.com/containernetworking/cni/pull/279/files#diff-b257c660535eb4a7279ff8eb1cc07214R46). PR #279 needs to be modified to reflect this, if I'm not mistaken. Having a list of addresses would help for multiple IPv6 addresses, e.g. when we want to display both an LLA and another address (ULA or GUA) for a pod interface.

plugins/main/bridge/bridge.go
leblancd: I was wondering the same thing, and was hoping someone had some historical context for why this the link flapping is there. I'll remove this in next update.

plugins/main/bridge/bridge.go
leblancd: I was trying to minimize the complexity of the current review, and add the handling for this special case (dual-stack, but with different ipam plugins for v4 vs v6) in a followup pull request.

plugins/main/bridge/bridge.go
leblancd: Same response. I can either add this handling into this current pull request, or add it as a followup. My preference would be to do staged pull requests.

plugins/main/bridge/bridge.go
leblancd: Ah, yes, good catch. I prefer just adding a To4() check here, rather than cononicalizing. There is a line below (280 old, 312 new) where I think it does make sense to remove the call to To4(), but would appreciate a double-check on the logic.

plugins/main/bridge/bridge.go
leblancd: At the moment there are 3 places where it's looping through all IPs in the IPAM results. What I could try to do is combine these 3 loops, and calculate first-V4 / first-V6 pod addresses, default routes, and gateway addresses all in this one loop. If that works, then there would only be one place where were detecting the V4/V6 family. If this is possible, I'll include this in the next uploaded patch.

plugins/main/bridge/bridge.go
leblancd: To add more detail... to handle the case where the config calls for a different IPAM plugin for V4 vs. V6, we would have to take the IPAM / IPAM6 config objects, separately marshal these objects into stdin (byte stream) format, call the separate IPAM plugins, and then merge the results. In the simpler case where the V4 and V6 use the same IPAM plugin, the net config stdin can be used as is. It's not terribly complicated, but I was going to add this in later.

leblancd:
This pull request has been ported to the containernetworking/plugins repo:
[plugins PR #10](https://github.com/containernetworking/plugins/pull/10)
Hopefully, changes requested here are sufficiently addressed in the new PR.

leblancd:
This PR has been moved to the containernetworking/plugins: 
[plugins PR #10](https://github.com/containernetworking/plugins/pull/10)

Issue #388, opened 2017-03-02
Should the Bridge plugin be chained?


squeed:
(credt to @dcbw for [dreaming this up](#384#issuecomment-283186620))
Right now, the bridge plugin sets up a veth pair, ensuring the host-side is in a bridge. Theoretically, it should be able to add any reasonable host-side interface to a bridge.

As a motivating example, consider a tap interface created by a previous step in the plugin chain.

Currently, the container-side of the veth is not configured particularly differently from the standard veth plugin.

Perhaps, then, the bridge plugin should accept a PreviousResult if supplied, and create a veth if not?

rosenhouse:
👍 This makes a lot of sense, at least in the abstract.

tomdee:
I like this too

feiskyer:
+1. Makes a lot of sense for taped (and other) devices.

PR #420, opened 2017-03-31
RFC: Split the bridge plugin functionality across bridge, veth and tap

mcastelino:
# Split the functionality of the bridge plugin into three seperate plugins.
- bridge: Handling just the bridge setup
- veth and tap: Which handle the setup of veth/tap interfaces
  and also invoke IPAM to setup the networking configuration
  of the interface as well as the bridge
- `exec-plugins.sh` enhanced to support plugin chaining

This builds on top of earlier proposal
https://github.com/containernetworking/cni/pull/384

This will also need the merge of capability args support to cnitool
https://github.com/containernetworking/cni/pull/440

# OPENs:
- The bridge should ideally not rely on the veth/tap plugins to setup the IP configuration of the bridge.
    - The bridge plugin should be the last in the chain allowing it to be configured using the results from the previous IPAM invocation
- Add veth functionality back into the bridge plugin for backward compatibility

# TODO:
- The unit test cases will have to re written to account for this split

# Testing with cnitool

cnitool already supports plugin chaining. 

This can be tested with the updated plugin chaining script using a conflist file shown below

## /etc/cni/net.d/10-mynet.conflist
```
{
        "cniVersion": "0.3.0",
        "name": "mynet",
        "plugins": [
        {
                "type": "bridge",
                "bridge": "cni0"
        },
        {
                "type": "veth",
                "bridge": "cni0",
                "isGateway": true,
                "ipMasq": true,
                "capabilities": {"configureVM": true},
                "ipam": {
                        "type": "host-local",
                        "subnet": "10.22.0.0/16",
                        "routes": [
                                { "dst": "0.0.0.0/0" }
                        ]
                }
        },
        {
                "type": "tap",
                "bridge": "cni0",
                "isGateway": true,
                "ipMasq": true,
                "capabilities": {"configureVM": true},
                "ipam": {
                        "type": "host-local",
                        "subnet": "10.22.0.0/16",
                        "routes": [
                                { "dst": "0.0.0.0/0" }
                        ]
                }
        }
        ]
}
```

## To create namespace friendly interfaces

```
sudo -E ./cnitool  add mynet /var/run/netns/net0
```

## To create VM friendly interfaces

```
export CAP_ARGS='{
   "configureVM": true
}'
sudo -E ./cnitool  add mynet /var/run/netns/net0
```

/cc @dcbw @squeed


squeed:
I think the right way to approach this is #388 - the Bridge plugin should work as a chained plugin.

mcastelino:
@squeed this PR was a initial attempt to implement https://github.com/containernetworking/cni/issues/388 based on the feedback I got for https://github.com/containernetworking/cni/pull/384

The example configuration files are just temporary till I add support for plugin chaining invocation to exec-plugins.sh. I am working on that right now along with some code refactoring.

However if the current split out does not meet the needs for plugin chaining as specified in https://github.com/containernetworking/cni/issues/388 please do let me know, so that I can implement this properly.

mcastelino:
@squeed reworked this a little bit and added plugin chaining support to the scripts. But I am still not sure how we should chain the plugins veth->tap->bridge or bridge->veth->tap. Hopefully the addresses your initial concern.

@dcbw Also placing the IPAM within both veth and tap does not appear to be most natural. So looking for feedback on how IPAM should interact with chained plugins.

squeed:
(Also answered in another PR, but reproducing here for completeness)
The chaining should be `(veth|tap) -> bridge`, with the administrator or runtime supplying the correct configuration w.r.t. the isolation environment.

The IPAM should be part of the first plugin.

feiskyer:
@mcastelino Both veth and tap are provided in the examples. Why veth is required for tap+bridge settings?

mcastelino:
@feiskyer The tap implementation is not complete yet. But the goal of the chaining is to ensure that you can use the same plugin chain for a host that can run VM and namespace based containers at the same time.

So in this case a namespace based runtime would invoke the plugin chain with configureVM not specified (or specify it as configureVM=false). This will ensure that in this case the tap plugin would be effectively a passthro and would just pass the prevResult it got as its result. This also ensures that runtimes that are not aware of this runtime property do not have to specify it.

However if the configureVM=true, then the veth plugin would be a passthro and the tap plugin would perform the interface setup. All VM based runtimes will need to be aware of this property and specify it to get VM friendly interfaces.

This idea was proposed by @dcbw https://github.com/containernetworking/cni/pull/384#issuecomment-283186620

feiskyer:
@mcastelino Thanks. Seems the order is wrong in the examples:

> the ordered config list is [ veth, tap, bridge ]

thockin:
I feel like a larger number of more end-to-end plugins might be better than expecting people to chain these together in a sane way.  e.g. `bridge_veth` vs `bridge_tap` vs `veth` would all reuse code but not be network assembly language for sysadmins..

mcastelino:
> @mcastelino Thanks. Seems the order is wrong in the examples:

@feiskyer yes. The conflist I used just reflects what is working today. The code I have in the PR today does not as yet support chaining the bridge to the end. I am working moving the bridge plugin to the end of the chain. So the chain can be [ veth, tap, bridge ] or [ tap, veth, bridge ]. The ordering of tap or veth will not matter.

mcastelino:
> I feel like a larger number of more end-to-end plugins might be better than expecting people to chain these together in a sane way. e.g. bridge_veth vs bridge_tap vs veth would all reuse code but not be network assembly language for sysadmins..

@thockin bridge_veth and bridge_tap would be simpler from an user perspective. 

@dcbw had suggested [splitting](https://github.com/containernetworking/cni/pull/384#issuecomment-283186620) the veth out of the bridge plugin. 

@squeed suggested [retaining ](https://github.com/containernetworking/cni/pull/427#issuecomment-295240106) the veth logic within the bridge plugin making it effectively bridge_veth. Based on that what we will end up with; is a chain that looks like [tap, bridge] when you need to support VM runtimes. And just [bridge] when you do not need that functionality.

So the question is should we support 

- [tap, bridge]

or 

- [bridge_tap, bridge]

Here bridge == bridge_veth so that we do not break any existing deployments.

bridge_tap will probably never be a standalone plugin as VM based runtimes will always need a namespace based runtime present to support [host privileged PODS](https://github.com/kubernetes-incubator/cri-o/blob/master/server/config.go#L77).

feiskyer:
@mcastelino @dcbw Any updates on this? When is this expected to be in?

mcastelino:
@feiskyer Sorry got sidetracked. I hope to get all the changes in within the next two weeks.

feiskyer:
@mcastelino Thanks. 👍

rosenhouse:
All plugins have moved to https://github.com/containernetworking/plugins

If you'd still like to contribute these changes, please open a new PR there.

Sorry for the churn.

PR #424, opened 2017-04-05
scripts: Add support for invoking plugin chains

mcastelino:
Add support to the command line test script to support invoking
plugin configurations which have plugin chains and those that
do not use plugin chains.

This is useful for exercising plugin chains from the command line.
This is also needed to be able to test 
https://github.com/containernetworking/cni/issues/388

To test the chaining implementation implemented in  https://github.com/containernetworking/cni/pull/420 the following configuration file can be used.

Longer term we should consider adding capability enabling as part of the script invocation. 

```
 /etc/cni/net.d/10-mynet.conf
{
        "cniVersion": "0.3.0",
        "name": "mynet",
        "plugins": [
        {
                "type": "bridge",
                "bridge": "cni0"
        },
        {
                "type": "veth",
                "bridge": "cni0",
                "isGateway": true,
                "ipMasq": true,
                "ipam": {
                        "type": "host-local",
                        "subnet": "10.22.0.0/16",
                        "routes": [
                                { "dst": "0.0.0.0/0" }
                        ]
                }
        },
        {
                "type": "tap",
                "bridge": "cni0",
                "isGateway": true,
                "ipMasq": true,
                "ipam": {
                        "type": "host-local",
                        "subnet": "10.22.0.0/16",
                        "routes": [
                                { "dst": "0.0.0.0/0" }
                        ]
                }
        }
        ]
}

```

/cc @dcbw 

Signed-off-by: Manohar Castelino <manohar.r.castelino@intel.com>

### DIFF
--- a/scripts/exec-plugins.sh
+++ b/scripts/exec-plugins.sh
@@ -4,6 +4,16 @@ if [[ ${DEBUG} -gt 0 ]]; then set -x; fi
 
 NETCONFPATH=${NETCONFPATH-/etc/cni/net.d}
 
+function process_error() {
+	errmsg=$(echo $1 | jq -r '.msg')
+	if [ -z "$errmsg" ]; then
+		errmsg=$res
+	fi
+
+	echo "${name} : error executing $CNI_COMMAND: $errmsg"
+	exit 1
+}
+
 function exec_plugins() {
 	i=0
 	contid=$2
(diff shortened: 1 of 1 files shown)


squeed:
Unfortunately, this doesn't set the PrevResult field.

Given how complicated actually doing chaining and capability arguments is, should we even bother trying? I suspect we should standardize on cnitool.

mcastelino:
@squeed agreed, I have already switched to using cnitool.

mcastelino:
Closing this PR. cnitool has support for plugin chains.

PR #425, opened 2017-04-07
scripts/release-with-rkt.sh Minor improvements

tomdee:
- Use a new Fedora to get a more up to date Go version
- Don't build so many archives, they are not used.
- Handle the new build.sh script
- Just use host networking with rkt


squeed:
👍

PR #430, opened 2017-04-16
Wondering if this is still a proposal?

kris-nova:
I was reading through the spec and noticed that it was labeled as a proposal. Is that the case?

Since we are already a few versions in, this feels more like a "specification" to me at this point. Somebody let me know if this was done intentionally - I am just here learning 😄


bboreham:
LGTM

kris-nova:
Yey! Thanks @bboreham and @rosenhouse!

PR #431, opened 2017-04-16
Can we specify with network interface here?

kris-nova:
Sorry, I was just reading through the spec. So my question here is:

Am I reading the spec correctly? Is this in fact the *container's interface* being referenced here?

If so, can we merge this in so we explicitly call this out.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -408,7 +408,7 @@ Also note that plugins are executed in reverse order from the ADD action.
 
 ### IP Allocation
 
-As part of its operation, a CNI plugin is expected to assign (and maintain) an IP address to the interface and install any necessary routes relevant for that interface. This gives the CNI plugin great flexibility but also places a large burden on it. Many CNI plugins would need to have the same code to support several IP management schemes that users may desire (e.g. dhcp, host-local). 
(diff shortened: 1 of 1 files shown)


bboreham:
It's the network interface which the CNI plugin has just created, or is just about to create.  Nothing makes it "the container's interface" other than the CNI plugin itself.

I mean, your change is not wrong, but it doesn't seem a particularly key emphasis to me.

Now I'm frustrated that the thing we want to talk about, the "container network interface", is also the name of the API.

dcbw:
It could also be a VM/hypervisor interface, since we just updated the spec to help accommodate those too.

qianzhangxa:
@dcbw So we extend the scope of CNI to not only handle the network of containers but also VM/hypervisor?

dcbw:
@qianzhangxa the recent 0.3.x API change which added "sandbox" to the Result structure was one of the first steps in that direction, yes.

bboreham:
How about something like "As part of its operation in creating an interface for a container, a CNI plugin that is expected to assign (and maintain) an IP address to that interface and ..."

bboreham:
@kris-nova do you want to have another go at the wording?

bboreham:
Closing due to lack of activity; please re-open if you want to take it further

PR #433, opened 2017-04-21
scripts: cleanup() should rm net container in docker-run.sh

dunjut:
Hi, if users do `Running a Docker container with network namespace set up by CNI plugins` by following steps in README, then after this, they'll get a dead net container (busybox) left on their environments.

I'm thinking that it would be better if we clean that dead container up and leave the users' environments as clean as before. See `git diff`.

### DIFF
--- a/scripts/docker-run.sh
+++ b/scripts/docker-run.sh
@@ -13,7 +13,7 @@ netnspath=/proc/$pid/ns/net
 
 function cleanup() {
 	./exec-plugins.sh del $contid $netnspath
-	docker kill $contid >/dev/null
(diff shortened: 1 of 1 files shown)


dunjut:
@squeed You just reviewed my last PR 2 hours ago, maybe you'd like to take a look at this one either. 😄

bboreham:
LGTM

squeed:
Ah yes, makes sense. LGTM.

dunjut:
Thank you guys. @bboreham @squeed

dcbw:
LGTM

PR #434, opened 2017-04-21
Fix unmarshall of results for interface index 0

### LINKED ISSUE #404: CNI result `ips[].interface` is omitted when set to 0
The following code returns the `interface` in the `ips` array:

code:

```go
	cniResult := &current.Result{}
	cniResult.Interfaces = append(cniResult.Interfaces,
		&current.Interface{Name: "foo"},
		&current.Interface{Name: "bar"},
	)
	cniResult.IPs = []*current.IPConfig{
		&current.IPConfig{
			Version:   "4",
			Interface: 1,
			Address: net.IPNet{
				IP:   net.IPv4(10, 255, 4, 5),
				Mask: net.IPv4Mask(255, 255, 255, 255),
			},
		},
	}
	return types.PrintResult(cniResult, "0.3.0")
```

result:

```json
      {
        "interfaces": [
          {
            "name": "foo"
          },
          {
            "name": "bar"
          }
        ],
        "ips": [
          {
            "version": "4",
            "interface": 1,
            "address": "10.255.4.5/32"
          }
        ],
        "dns": {}
      }
```

But when you set `interface` to `0`, the field is omitted from the result:

code:

```go
	cniResult := &current.Result{}
	cniResult.Interfaces = append(cniResult.Interfaces,
		&current.Interface{Name: "foo"},
		&current.Interface{Name: "bar"},
	)
	cniResult.IPs = []*current.IPConfig{
		&current.IPConfig{
			Version:   "4",
			Interface: 0,
			Address: net.IPNet{
				IP:   net.IPv4(10, 255, 4, 5),
				Mask: net.IPv4Mask(255, 255, 255, 255),
			},
		},
	}
	return types.PrintResult(cniResult, "0.3.0")
```

result:

```json
      {
        "interfaces": [
          {
            "name": "foo"
          },
          {
            "name": "bar"
          }
        ],
        "ips": [
          {
            "version": "4",
            "address": "10.255.4.5/32"
          }
        ],
        "dns": {}
      }
```

vikaschoudhary16:
In the "ips" struct, if the plugin passes interface index as 0 to
cni unmarshalling api, index index gets omitted in the result.

fixes #404

Signed-off-by: vikaschoudhary16 <vichoudh@redhat.com>

### DIFF
--- a/pkg/types/current/types.go
+++ b/pkg/types/current/types.go
@@ -261,7 +261,7 @@ func (i *IPConfig) String() string {
 // JSON (un)marshallable types
 type ipConfig struct {
 	Version   string      `json:"version"`
-	Interface int         `json:"interface,omitempty"`
+	Interface int         `json:"interface"`
 	Address   types.IPNet `json:"address"`
 	Gateway   net.IP      `json:"gateway,omitempty"`
 }


dcbw:
I think this might be a problem for IPAM plugins which use the same structure, but don't pass anything about interface indexes.  Their JSON would now include 0 where before it was empty, and their results do not always correspond to any specific interface index.

vikaschoudhary16:
@dcbw , IPAM plugins are not supposed to pass anything about interface indexes as per the [cni spec](https://github.com/containernetworking/cni/blob/master/SPEC.md#result). May be i misunderstood your point. Would you mind please reword it again for me?

squeed:
@dcbw I thought you were gonna replace that with an `*int`...

dcbw:
CLosing in favor of https://github.com/containernetworking/plugins/pull/28

PR #437, opened 2017-04-21
Added version command to cnitool

### LINKED ISSUE #436: Add VERSION command to cnitool plugin
After reading the spec, it seems like `cnitool` should also handle the `version` command. I'd be happy to add this.

micahhausler:
cnitool: added version command

fixes #436

### DIFF
--- a/cnitool/cni.go
+++ b/cnitool/cni.go
@@ -18,21 +18,34 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/containernetworking/cni/libcni"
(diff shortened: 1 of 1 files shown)


squeed:
IMHO, it would be interesting if `cnitool` reported the VERSION of all the configured plugins.

`cnitool` isn't a plugin, it's a trivial runtime.

dcbw:
Yeah, agree with @squeed.  The functionality this PR adds should really be triggered by a --version argument rather than the version command.  But having the version command as @squeed suggests would be awesome too!

PR #446, opened 2017-05-04
Log changes

zdj6373:
Join CNI, starting with the modified log


squeed:
Hi there, welcome to CNI.

In golang projects, you should not change vendored code directly. Instead, you should change the "upstream" code, then pull in those changes separately.

Let me know if you have any questions, thanks.

Issue #438, opened 2017-04-24
Investigate replacing mac-address-changing with gratuitous arp


squeed:
This would make SLAAC much simpler.

squeed:
This was done in containernetworking/plugins#16

PR #448, opened 2017-05-05
 macvlan, bridge, ptp: send gratuitous arp, don't change HWAddr

squeed:
We used to set the HWAddr based on the assigned IP in case we re-use IPs in a short timescale, so we don't have to wait for the arp table to time out. Since this makes ipv6 harder, don't do it. Send a gratuitous ARP instead.

Fixes: #438


dcbw:
LGTM, though major bonus points if you use gopacket to detect and verify that the ARP gets sent :)

http://www.devdungeon.com/content/packet-capture-injection-and-analysis-gopacket

squeed:
Also for the record: we added round-robin to the `host-local` ipam plugin, so the chance of quickly re-using an IP is a lot lower than before.

squeed:
Ask and ye shall receive, @dcbw  :-)

dcbw:
Neato, bonus approved and in the mail.

bboreham:
Wondering whether "github.com/j-keck/arping" would be a better dependency; it seems to have a smaller footprint.

squeed:
@bboreham I think you're right - I'll swap it out. Shouldn't change much.

squeed:
@bboreham Swapped libraries and updated. Ahh, much better.

dcbw:
LGTM

squeed:
This probably shouldn't be merged - Once it gets another LGTM, I'll move it to the plugins repository.

squeed:
Nevermind, we should merge this and then snarf that commit in to plugins.

We should also delete the plugins here soon :-)

bboreham:
> Also, added some test deps to godeps.json, but didn't change any actual vendor files

Can you explain this a bit more?  Why do we want to list things in godeps.json that we don't actually vendor in?

pkg/ip/arp.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"net"
+
+	"github.com/j-keck/arping"
+)
+
+// GratuitousArp sends an arp for the supplied IP address on the interface ifName
+func GratuitousArp(iface *net.Interface, addr net.IP) error {
+	return arping.GratuitousArpOverIface(addr, *iface)
```
bboreham: I question the value of a 26-line file and 99-line test to swap two arguments round

bboreham:
Aside from complaining about the function that swaps two arguments round, the basic functionality lgtm.

squeed:
Closing in favor of containernetworking/plugins#16

PR #450, opened 2017-05-05
cniVersion and config structure mismatch

gunjan5:
`cniVersion` is specified `0.3.1` but the IP structure is from older CNI spec versions [`"ip4"` instead of `"ips"` with its list, etc.]


dcbw:
LGTM

squeed:
LGTM, thanks!

PR #451, opened 2017-05-11
ipam/host-local: support multiple IP ranges

squeed:
This change allows the host-local allocator to allocate multiple IPs. This is intended to enable dual-stack, but is not limited to only two subnets.

Not actually a large change; mostly improved testing.

### DIFF
--- a/Documentation/host-local.md
+++ b/Documentation/host-local.md
@@ -5,30 +5,41 @@ it can include a DNS configuration from a `resolv.conf` file on the host.
 
 ## Overview
 
-host-local IPAM plugin allocates IPv4 addresses out of a specified address range.
+host-local IPAM plugin allocates ip addresses out of a set of address ranges.
 It stores the state locally on the host filesystem, therefore ensuring uniqueness of IP addresses on a single host.
 
 ## Example configurations
 
-IPv4:
 ```json
 {
 	"ipam": {
 		"type": "host-local",
-		"subnet": "10.10.0.0/16",
-		"rangeStart": "10.10.1.20",
-		"rangeEnd": "10.10.3.50",
-		"gateway": "10.10.0.254",
+		"ranges": [
+			{
+				"subnet": "10.10.0.0/16",
+				"rangeStart": "10.10.1.20",
+				"rangeEnd": "10.10.3.50",
+				"gateway": "10.10.0.254"
+			},
+			{
+				"subnet": "3ffe:ffff:0:01ff::/64",
+				"rangeStart": "3ffe:ffff:0:01ff::0010",
+				"rangeEnd": "3ffe:ffff:0:01ff::0020",
+			}
+		],
 		"routes": [
 			{ "dst": "0.0.0.0/0" },
-			{ "dst": "192.168.0.0/16", "gw": "10.10.5.1" }
+			{ "dst": "192.168.0.0/16", "gw": "10.10.5.1" },
+			{ "dst": "3ffe:ffff:0:01ff::1/64" }
 		],
-		"dataDir": "/var/my-orchestrator/container-ipam-state"
+		"dataDir": "/run/my-orchestrator/container-ipam-state"
 	}
 }
 ```
 
-IPv6:
+Previous versions of the `host-local` allocator did not support the `ranges`
+property, and instead expected a single range on the top level. This is
+deprecated but still supported.
 ```json
 {
   "ipam": {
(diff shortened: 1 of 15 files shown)


plugins/ipam/host-local/backend/disk/backend.go
```diff
@@ -71,7 +74,7 @@ func (s *Store) Reserve(id string, ip net.IP) (bool, error) {
 		return false, err
 	}
 	// store the reserved ip in lastIPFile
-	ipfile := filepath.Join(s.dataDir, lastIPFile)
+	ipfile := filepath.Join(s.dataDir, lastIPFileName(rangeIdx))
 	err = ioutil.WriteFile(ipfile, []byte(ip.String()), 0644)
 	if err != nil {
```
dcbw: Hmm, we don't os.Remove() the file in this error case?

plugins/ipam/host-local/backend/allocator/range.go
```diff
@@ -0,0 +1,145 @@
...
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/containernetworking/cni/pkg/ip"
+	"github.com/containernetworking/cni/pkg/types"
+)
+
+// Validate takes a given range and ensures that all information is consistent,
```
leblancd: "Validate" should be "Canonicalize".

plugins/ipam/host-local/backend/allocator/range.go
```diff
@@ -0,0 +1,145 @@
...
+	if r.RangeStart != nil {
+		if err := canonicalizeIP(&r.RangeStart); err != nil {
+			return err
+		}
+
+		if err := r.IPInRange(r.RangeStart); err != nil {
+			return err
+		}
+	} else {
+		r.RangeStart = ip.NextIP(ip.NextIP(r.Subnet.IP))
```
leblancd: If the gateway is set to something other than the .1 address, should the RangeStart be initialized to .1? Or maybe simpler: unconditionally set the RangeStart to .1 here, since there's a check in Get() so that you don't allocate the gateway address anyway.

plugins/ipam/host-local/backend/allocator/range.go
```diff
@@ -0,0 +1,145 @@
...
+	if len(addr) != len(r.Subnet.IP) {
+		return fmt.Errorf("IP %s is not the same protocol as subnet %s",
+			addr, subnet.String())
+	}
+
+	if !subnet.Contains(addr) {
+		return fmt.Errorf("%s not in network %s", addr, subnet.String())
+	}
+
+	// We ignore nils here so we can use this function as we initialize the range.
```
leblancd: Out of curiosity: Is this comment still accurate re. "ignore nils here so we can use this function as we initialize"? (Regardless of the comment, we probably still want the nil checks as a defensive measure.)

plugins/ipam/host-local/backend/allocator/config_test.go
```diff
@@ -0,0 +1,235 @@
...
+	It("Should parse an old-style config", func() {
+		input := `{
+    "cniVersion": "0.3.1",
+    "name": "mynet",
+    "type": "ipvlan",
+    "master": "foo0",
+    "ipam": {
+        "type": "host-local",
+        "subnet": "10.1.2.0/24",
+		"rangeStart": "10.1.2.9",
```
leblancd: Nit: Looks like some extra tabs sneaked into these 3 lines of config.

plugins/ipam/host-local/backend/allocator/config.go
```diff
@@ -58,27 +73,49 @@ func LoadIPAMConfig(bytes []byte, args string) (*IPAMConfig, string, error) {
...
+	}
+
+	// If a single range (old-style config) is specified, move it to
+	// the Ranges array
+	if n.IPAM.Range != nil && n.IPAM.Range.Subnet.IP != nil {
+		n.IPAM.Ranges = append([]Range{*n.IPAM.Range}, n.IPAM.Ranges...)
+	}
+	n.IPAM.Range = nil
+
+	if len(n.IPAM.Ranges) == 0 {
```
squeed: Good point. Older versions support one v4 and one v6 result, so we should allow that.

Documentation/host-local.md
```diff
@@ -5,30 +5,41 @@ it can include a DNS configuration from a `resolv.conf` file on the host.
...
-IPv4:
 ```json
 {
 	"ipam": {
 		"type": "host-local",
-		"subnet": "10.10.0.0/16",
-		"rangeStart": "10.10.1.20",
-		"rangeEnd": "10.10.3.50",
-		"gateway": "10.10.0.254",
+		"ranges": [
```
squeed: Hah! Funny.
I thought about the same thing, but ultimately decided on the keeping the same names for ease of transition.

plugins/ipam/host-local/backend/allocator/range.go
squeed: Yeah - though this could be clearer. What I meant was that this function is called before those properties are set.

plugins/ipam/host-local/backend/allocator/allocator.go
```diff
@@ -21,187 +21,101 @@ import (
...
+		ipRange:  c.Ranges[rangeIdx],
+		store:    store,
+		rangeIdx: rangeIdx,
 	}
-	return nil
 }
 
-// Returns newly allocated IP along with its config
-func (a *IPAllocator) Get(id string) (*current.IPConfig, []*types.Route, error) {
+// Returns newly allocated IP along with its ipRangeig
```
squeed: search and replace strikes again.

squeed:
Moved to containernetworking/plugins#12.

PR #458, opened 2017-05-18
Allow setting promiscuous mode for bridge

bakins:
Currently, one must set promiscuous mode using a script or some other out-of-band means.  This adds an option to the configuration to set it.

This is to address concerns with the issue described in https://github.com/moby/moby/issues/5618#issuecomment-280105097

There is a flag on [netlink.LinkAttrs](https://godoc.org/github.com/vishvananda/netlink#LinkAttrs), but the code presented here works and required no change to the surrounding code.


dnardo:
I believe #441 takes care of this?

bakins:
@dnardo yes it does. I didn't even check the existing PRs. consider this a +1 to that PR :)

PR #461, opened 2017-05-30
pkg/invoke: ensure that custom env vars are prepended to the env

### LINKED ISSUE #410: invoke.Args.AsEnv() does not override existing environment
[The code](https://github.com/containernetworking/cni/blob/b87126377a207a00ce32bcf67ea059ab8b9f6ba6/pkg/invoke/args.go#L60) takes the current process' environment and appends some strings `CNI_COMMAND=ADD`, etc., to it.

Sadly, Go's `Getenv()` function on Unix only takes the first match, so anything previously in the environment will be taken as the parameter, not the ones appended.

ericchiang:
Ensure that custom values take priority over values defined in the
environment by pre-pending them to the env, not appending.

closes #410

### DIFF
--- a/pkg/invoke/args.go
+++ b/pkg/invoke/args.go
@@ -57,13 +57,16 @@ func (args *Args) AsEnv() []string {
 		pluginArgsStr = stringify(args.PluginArgs)
 	}
 
-	env = append(env,
-		"CNI_COMMAND="+args.Command,
-		"CNI_CONTAINERID="+args.ContainerID,
-		"CNI_NETNS="+args.NetNS,
-		"CNI_ARGS="+pluginArgsStr,
-		"CNI_IFNAME="+args.IfName,
-		"CNI_PATH="+args.Path)
+	// Ensure that the custom values are first, so any value present in
+	// the process environment won't override them.
+	env = append([]string{
+		"CNI_COMMAND=" + args.Command,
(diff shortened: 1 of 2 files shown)


squeed:
LGTM.

ericchiang:
bump. what's the process for merging?

rosenhouse:
The process for merging is described in our [Acceptance policy](https://github.com/containernetworking/cni/blob/master/CONTRIBUTING.md#acceptance-policy).

I've added a unit test to cover this change.  I think it's good to go.

rosenhouse:
From the [release notes for Go 1.9 beta](https://tip.golang.org/doc/go1.9#os/exec):

> The `os/exec` package now prevents child processes from being created with any duplicate environment variables. If `Cmd.Env` contains duplicate environment keys, only the last value in the slice for each duplicate key is used.

That would seem to break the intended behavior of this PR.

bboreham:
Ha ha, don'tcha love it when two groups fix a problem in conflicting ways...

rosenhouse:
Just hit this.  Had run `export CNI_COMMAND=VERSION` in my shell and then tried to run some tests that used `invoke`.  Fun times debugging.

Once we figure out how to get #538 green and merge, we should flip this back to appending.

PR #464, opened 2017-06-02
pkg: move packages from cni to plugins

squeed:
Removes the following packages:
* ip
* ipam
* ns
* testutils
* utils


timchenxiaoyu:
why remove this ,i see calico cni plugin error because no ip ns ...  packages

rosenhouse:
@timchenxiaoyu These packages have moved into the [new plugins repo](https://github.com/containernetworking/plugins).

PR #467, opened 2017-06-03
Clarify first released spec with 'args' was 0.2.0

### LINKED ISSUE #429: Mistake in history of "args"
https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#args-in-network-config says the `args` field was introduced in spec version 0.1.0.  As best I can tell, it was actually introduced in spec version 0.2.  The "Spec changes" listed for https://github.com/containernetworking/cni/releases/tag/v0.4.0, which says it corresponds to spec version 0.2.0, includes 'introduce "args" field and new error code'.

bboreham:
Fixes #429 

Note I'm using the phrase "release of the spec" to try to accommodate the fact that changes to the spec were made between 0.1.0 and 0.2.0, but the first spec you should consider "released" with `args` in it is the 0.4.0 tag in this repo.

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -58,7 +58,7 @@ But the runtime would fill in the mappings so the plugin itself would receive so
 ```
 
 ## "args" in network config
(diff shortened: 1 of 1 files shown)


squeed:
lgtm

PR #471, opened 2017-06-07
Clean up vendor

rosenhouse:
With #464 done, we can remove some unused dependencies.

cc: @squeed


squeed:
I still think we should just get rid of godeps / vendor entirely...

rosenhouse:
I agree, I just figured I'd start by removing the mostly clearly useless bits :-)

tomdee:
LGTM - I agree with @squeed on getting rid of godeps or using glide.

PR #477, opened 2017-06-15
types: fix marshalling of omitted "interfaces" key in IPConfig JSON

dcbw:
Plugins that don't have knowledge of interfaces, like host-local or
other IPAM plugins, should not set the 'interfaces' key of their
returned "Result" JSON.  This should then not be translated into
an interface index of 0, which it was due to the int marshaling and
omitempty.

Instead, ensure that an omitted 'interface' in JSON ends up being
-1 in the IPConfig structure, and that a -1 ensures that no 'interfaces'
key is present in the JSON.

Yes, this means that you must set Interface:-1 in IPConfig in your
plugin, but if you weren't already doing that and your plugin didn't
deal with interface indexes, you already had a problem.

A bit ugly, but we can't also change the IPConfig.Interface field to a *int because then it's really, really icky to actually set that member, since you can't &0 and such.  Thoughts?

Will require a plugins revendor and update as well.

Supercedes: https://github.com/containernetworking/cni/pull/434
Fixes: https://github.com/containernetworking/cni/issues/404

### DIFF
--- a/pkg/types/current/types.go
+++ b/pkg/types/current/types.go
@@ -70,10 +70,9 @@ func convertFrom020(result types.Result) (*Result, error) {
 
 	if oldResult.IP4 != nil {
 		newResult.IPs = append(newResult.IPs, &IPConfig{
-			Version:   "4",
-			Interface: -1,
-			Address:   oldResult.IP4.IP,
-			Gateway:   oldResult.IP4.Gateway,
+			Version: "4",
+			Address: oldResult.IP4.IP,
+			Gateway: oldResult.IP4.Gateway,
 		})
 		for _, route := range oldResult.IP4.Routes {
 			gw := route.GW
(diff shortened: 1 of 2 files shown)


rosenhouse:
> but we can't also change the IPConfig.Interface field to a *int because then it's really, really icky to actually set that member, since you can't &0 and such.

Some libraries address this by providing a helper function, e.g.
```go
// Int returns a pointer to the int value passed in.
func Int(v int) *int {
	return &v
}
```
[ref [`aws-sdk-go`](https://github.com/aws/aws-sdk-go/blob/1a4946ffe3/aws/convert_types.go#L124)]

I don't love it.  But it's worth considering.

squeed:
We really have no other choice. It's an optional field.

(something something type unions something rust something)

dcbw:
@rosenhouse yeah, that's the least bad option.  I disliked having to ensure Interface: was -1, I feel like that would often be forgotten.  Your way is better.  PTAL thanks!

pkg/types/current/types.go
```diff
@@ -249,12 +247,18 @@ func (i *Interface) String() string {
...
+	return &v
+}
+
 // IPConfig contains values necessary to configure an IP address on an interface
 type IPConfig struct {
 	// IP version, either "4" or "6"
 	Version string
-	// Index into Result structs Interfaces list
-	Interface int
+	// Index into Result structs Interfaces list; -1 if no interface is given
```
squeed: remove this doc comment.

dcbw: @squeed done

PR #479, opened 2017-06-22
docs: adds linux dep for building bins

danehans:
Previously, the linux dep for building cni ref plugin bins was
unclear. This can cause problems with new devs that use a Mac for
development. This PR adds the Linux dep to the Requirements section
of the readme.

Fixes: https://github.com/containernetworking/plugins/issues/27

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ If you intend to contribute to code or documentation, please read [CONTRIBUTING.
 
 The CNI spec is language agnostic.  To use the Go language libraries in this repository, you'll need a recent version of Go.  Our [automated tests](https://travis-ci.org/containernetworking/cni/builds) cover Go versions 1.7 and 1.8.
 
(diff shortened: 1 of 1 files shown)


rosenhouse:
Thanks!

squeed:
Not sure this is right - you should be able to compile the plugins for linux using goxc. They just don't *run* on non-linux platforms.

dcbw:
@squeed but *does* anyone compile them with goxc?

bboreham:
I agree we should say something.  If we don't actually provide the mechanism to cross-compile (e.g. in `build.sh`) then we may as well say it doesn't work apart from on Linux.

aaronjwood:
Relating to @bboreham's comment how about just doing this? https://github.com/containernetworking/plugins/pull/44

No build containers/VMs needed! :)

rosenhouse:
@danehans would you like to update this PR to respond to the feedback or close it?

danehans:
Closing since the above PR covers the issue.

PR #483, opened 2017-07-13
add `100` to well-known error codes

flyer103:
According to [this code](https://github.com/containernetworking/cni/blob/master/pkg/skel/skel.go#L144), it may be good to specify that `100` is a well-known error code in the SPEC. 
In this way, it could make developers less puzzled when the cni-plugin doesn't return an error which type is not [types.Error](https://github.com/containernetworking/cni/blob/master/pkg/types/types.go#L132).

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -533,3 +533,4 @@ The `dns` field contains a dictionary consisting of common DNS information.
 ## Well-known Error Codes
 - `1` - Incompatible CNI version
 - `2` - Unsupported field in network configuration. The error message must contain the key and value of the unsupported field.
+- `100` - Error during invocation of cni plugin. Plugin didn't specify the error code.


SPEC.md
```diff
@@ -533,3 +533,4 @@ The `dns` field contains a dictionary consisting of common DNS information.
 ## Well-known Error Codes
 - `1` - Incompatible CNI version
 - `2` - Unsupported field in network configuration. The error message must contain the key and value of the unsupported field.
+- `100` - Error happened when invoking cni plugin and the cni pulgin didn't specify the error code
```
jellonek: Typo in `pulgin`

flyer103: fixed

jellonek: Also this can be rephrased as "Error during invocation of cni plugin. Plugin didn't specified the error code."

flyer103: Yes, it's more consistent with the above description 👍 
 Have changed the description.

flyer103:
cc @squeed @dcbw

SPEC.md
```diff
@@ -533,3 +533,4 @@ The `dns` field contains a dictionary consisting of common DNS information.
 ## Well-known Error Codes
 - `1` - Incompatible CNI version
 - `2` - Unsupported field in network configuration. The error message must contain the key and value of the unsupported field.
+- `100` - Error during invocation of cni plugin. Plugin didn't specify the error code.
```
dcbw: I would say something like:

`Generic error during invocation of a plugin. Used when the plugin does not specify the error code.  Can also be used by frameworks and libraries that implement the CNI specification to indicate errors that occur before the plugin is actually invoked.`

flyer103: Thanks for your reply and I have changed the commit log.
I think it's more proper for you or other members to add the above description to the SPEC. My understanding is limited to how it is used in the code and `100` may be used in other libs that implement the CNI specification :D

PR #484, opened 2017-07-13
check the range of error status code after invoking cni-plugin

flyer103:
According to the [SPEC](https://github.com/containernetworking/cni/blob/master/SPEC.md#result), it may be good to check the range of error status code returned by cni-plugin.

### DIFF
--- a/pkg/skel/skel.go
+++ b/pkg/skel/skel.go
@@ -159,7 +159,16 @@ func (t *dispatcher) checkVersionAndCall(cmdArgs *CmdArgs, pluginVersionInfo ver
 			Details: verErr.Details(),
 		}
 	}
-	return toCall(cmdArgs)
+
+	err = toCall(cmdArgs)
+	if e, ok := err.(*types.Error); ok {
+		if e.Code < 100 {
+			e.Code = 100
+		}
+		return e
+	}
+
+	return err
 }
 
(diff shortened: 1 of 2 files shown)


flyer103:
I have written a cni plugin for k8s recently and found that cni skel does not check the range of error code returned by cni plugin. 
So if there's a error-code-check in cni skel, it may be helpful to help developers to comply with the cni spec when writing cni plugin.

flyer103:
cc @dcbw

rosenhouse:
I don't understand why this PR is valuable.

As a user, I expect the runtime to return me the same error code that the plugin used.  This PR breaks that expectation.

For example, suppose the plugin detects that it has been called with an unsupported CNI version field.  The [spec](https://github.com/containernetworking/cni/blob/master/SPEC.md#well-known-error-codes) says the plugin should return code `1`.  But with this PR, the user would not see that code.

flyer103:
I thought that `well-known error codes` were reserved for cni plugin skel to use other than just the part that was developed by cni plugin developers.
I will rethink the relationships between `well-known error codes` and `plugin specific error codes` in the SPEC. Thanks.

rosenhouse:
Ok, in the meantime I'm going to close this.  If you see another improvement, feel free to open a new PR.

PR #487, opened 2017-07-27
README plugins list: add Linen CNI plugin

John-Lin:
Hi, I built this plugin for overlay networks with Open vSwitch. It mainly focuses on 
SDN/OpenFlow network and supports for adding SDN controller such as ONOS, Ryu and 
OpenDaylight by configuring network configuration.


bboreham:
Thanks!

PR #490, opened 2017-08-15
Add a GET action to the CNI spec, packages, and sample plugin

dcbw:


dcbw:

SPEC.md
```diff
@@ -70,6 +70,19 @@ The operations that the CNI plugin needs to support are:
...
+    - **Network configuration**. This is a JSON document describing a network to which a container can be joined. The schema is described below.
+    - **Extra arguments**. This provides an alternative mechanism to allow simple configuration of CNI plugins on a per-container basis.
+    - **Name of the interface inside the container**. This is the name that should be assigned to the interface created inside the container (network namespace); consequently it must comply with the standard Linux restrictions on interface names.
+  - Result:
+    - **Interfaces list**. Depending on the plugin, this can include the sandbox (eg, container or hypervisor) interface name and/or the host interface name, the hardware addresses of each interface, and details about the sandbox (if any) the interface is in.
+    - **IP configuration assigned to each interface**. The IPv4 and/or IPv6 addresses, gateways, and routes assigned to sandbox and/or host interfaces.
+    - **DNS information**. Dictionary that includes DNS information for nameservers, domain, search domains and options.
+
```
matthewdupre: Second paragraph in the result structure talks about ADDs but should also mention GETs.  It would be a good place to specify how similar the result must be to that from an ADD (I see the previous maintainer's meeting we said the same, although that's not SLAAC / IP change friendly).

matthewdupre: (Line 113 below)

squeed:
Does the spec want to take a stand on how divergent the results between GET and ADD can be? Is this a MUST? A SHOULD? A "you MUST document exceptions?"

squeed:
Need to edit lines 273+ - where ConfigLists are specified.

SPEC.md
chinangela: It would be cleaner to not copy-paste the parameters descriptions from add. Instead, the parameters can be noted "as defined above", similar to how it is for delete.

pkg/version/plugin.go
```diff
@@ -79,3 +81,46 @@ func (*PluginDecoder) Decode(jsonBytes []byte) (PluginInfo, error) {
...
+		minor, err = strconv.Atoi(parts[1])
+		if err != nil {
+			return -1, -1, -1, fmt.Errorf("failed to convert minor version part %q: %v", parts[1], err)
+		}
+	}
+
+	if len(parts) >= 3 {
+		micro, err = strconv.Atoi(parts[2])
+		if err != nil {
+			return -1, -1, -1, fmt.Errorf("failed to convert minor version part %q: %v", parts[2], err)
```
chinangela: The error message for the micro version says `failed to convert minor version`. It should say `failed to convert patch version`.

SPEC.md
```diff
@@ -70,6 +70,19 @@ The operations that the CNI plugin needs to support are:
     - **Extra arguments**, as defined above.
```
knobunc: For some reason github won't let me comment on the right line.  Anyway, can we add something to line 61 the "IP configuration assigned to each interface" result to explain that you may not get the IP information immediately.

Can we also describe what the expectations are about a GET?  Is the container runtime expected to periodically call GET throughout the lifecycle of the container?  (And how often?)

Can interfaces be added while the container is running?

Once an interface has an IP address... can it change?  If it has a v4 address, can it later get a v6 address?

Is there a way to know that the network plugin is "done" and that the config is final and doesn't need to be polled any more?

dcbw:
@containernetworking/cni-maintainers and now with more testing...

dcbw:
@containernetworking/cni-maintainers any further comments?  Nobody from the last Community Meeting added any comments, so perhaps this is good to go?

squeed:
@dcbw I was going to try and implement the PR as-mentioned for one or two of the plugins, but haven't done that yet. I think the spec is good as-is.

dcbw:
@squeed I do have an implementation for most plugins at https://github.com/containernetworking/plugins/pull/55 if that helps.

squeed:
I've started to add the "check" logic, but I'm a bit concerned about validating routes. Consider the scenario:

1. The first plugin creates a non-default route
2. A separate chained plugin removes or adjusts it.

These would be completely valid within the spec, but it wouldn't be possible for the plugin to know beforehand what was expected. So, either we should better define what validation means, or scrap it.

dcbw:
@squeed @containernetworking/cni-maintainers while updating the spec, I was left with this question.  How does an overall plugin chain determine that the container is in an error state, when each individual plugin is not supposed to return an error or validate its config?

For example, say that a plugin adds an interface, IP address, and a default route.  Something causes the default route to go away, or the IP address to change, or whatever.  The container is now no longer able to do much of anything.

If the plugin is not able to validate that the expected IP address and default route are present on the interface, because they might have been changed by a later plugin in the chain (which is a reasonable expectation), how do we tell that the container is in an error state if this is the last plugin?

squeed:
I think we have to abandon our goal of "GET does some kind of validation." GET was originally intended to be along the lines of "I forgot what IP address I handed this container," not "I crashed while setting up the network and want to know if it is ready."

That being said, I think there is some room for validation. Specifically, there are invariants around the `CNI_$FOO` arguments:

- An interface in `CNI_NETNS` called `CNI_IFNAME` exists and is up
- That interface is of a type created by the interface-creating plugin

As a motivating example, consider the magic plugin that wraps a namespace in a trivial bridge with a tap device. It has to remove the device's address (and therefore the routes) to do so.

dcbw:
@squeed Yeah, I mostly agree.  But I'm not sure we can depend on the interface type being the same, no?

After a bunch of thought and also discussion in the CNI maintainers meeting this week , I changed the spec to say that plugins should fill in as much of their Result as they can, and once you're done with a ConfigList you should arrive at the same Result as was returned from ADD.  Also added a note about erroring out on general state problems, not interface/address/etc problems.  I still think it's useful for long-running plugins to be able to return errors like that.

dcbw:
@bboreham suggested updating the plugins implementation in https://github.com/containernetworking/plugins/pull/55 to see if any further issues fall out, and to implement testcases there to validate the behavior and see if it makes sense.

dcbw:
@matthewdupre suggested that the requirement that GET only able to be called between ADD/DEL is a bit too strict.  Maybe we could specify a well-known error for "container doesn't exist".

bboreham:
LGTM but I suggest waiting until https://github.com/containernetworking/plugins/pull/55 is done, to be sure we don't need any more changes to the spec before merging.

dcbw:
@containernetworking/cni-maintainers here's an update with caching of the ADD result, which then gets passed to plugins as prevResult for GET and DEL.  The tests aren't all there yet, but looking for feedback on this approach.

dcbw:
@containernetworking/cni-maintainers @squeed  PTAL thanks!

pkg/version/plugin.go
```diff
@@ -79,3 +81,46 @@ func (*PluginDecoder) Decode(jsonBytes []byte) (PluginInfo, error) {
...
+	if err != nil {
+		return false, err
+	}
+
+	secondMajor, secondMinor, secondMicro, err := ParseVersion(otherVersion)
+	if err != nil {
+		return false, err
+	}
+
+	return firstMajor >= secondMajor && firstMinor >= secondMinor && firstMicro >= secondMicro, nil
```
rosenhouse: What about `0.4.0 > 0.3.1`?  Will that evaluate `true`?
Might be worth considering using a semver library?

dcbw: @rosenhouse fixed...

SPEC.md
```diff
@@ -85,20 +85,40 @@ The operations that CNI plugins must support are:
...
+  - The plugin should not return an error if its expected sandbox state (eg interfaces, IP addresses, routes, etc) is not found, as subsequent plugins in the chain may alter this state.
+  - The plugin will be passed a `prevResult` field in its configuration JSON on stdin which contains the `Result` object which the plugin (or the chain of which the plugin is part) returned to the runtime 
+  regenerate as much of its `Result` object as it can, which may be modified by plugins in the chain. For example, if the plugin-created interface exists and has the expected IP address, the plugin should add the interface and address to its `Result`. If the plugin-created interface does not have the expected IP address, the plugin should add the interface (but not the expected IP address) to its `Result`.
+  - A runtime may call `GET` at any time; but if `GET` is called for a given container before an `ADD` or after a `DEL` for that container, the plugin should return error 3 to indicate the container is unknown (see [Well-known Error Codes](#well-known-error-codes) section).
```
matthewdupre: Should we suggest that runtimes call DEL then ADD on a different error, but may proceed straight to ADD in the error 3 case?

SPEC.md
```diff
@@ -85,20 +85,40 @@ The operations that CNI plugins must support are:
...
+  - This action should return the same result as an `ADD` action for the same inputs. The result should not change over the lifetime of the container.
+  - The plugin should return an error if any general internal state is unexpected. For example, if the plugin's data storage is missing or corrupt, or it cannot talk to a controller (if it has one) controller, it should return an error.
+  - The plugin should not return an error if its expected sandbox state (eg interfaces, IP addresses, routes, etc) is not found, as subsequent plugins in the chain may alter this state.
+  - The plugin will be passed a `prevResult` field in its configuration JSON on stdin which contains the `Result` object which the plugin (or the chain of which the plugin is part) returned to the runtime 
+  regenerate as much of its `Result` object as it can, which may be modified by plugins in the chain. For example, if the plugin-created interface exists and has the expected IP address, the plugin should add the interface and address to its `Result`. If the plugin-created interface does not have the expected IP address, the plugin should add the interface (but not the expected IP address) to its `Result`.
```
matthewdupre: Text is garbled between these two lines "returned to the runtime\nregenerate as much"; perhaps a mis-edit.

Having said that I still don't love this part (I know we've gone round and round on it).

Is this statement of what the prevResult contains valid from the perspective of the second(+) plugin in the chain?  Can we provide instructions for what a plugin should do that are independent of its chain position?

Is it fair to say that a chain of plugins should return the prevResult, or an error?  There's a "should" in the first line, but this leaves some wiggle room, so I think it's important to close down whether and how a runtime needs to deal with getting a different non-error response to the ADD or if it's prohibited.

If it's prohibited then it might still make sense to allow plugins to modify it, and just turn chain result != chain prevResult into an error in libcni - that would allow a plugin to say "this isn't the IP I gave it" by changing that field, and then a later plugin that modified the IP can change it back.  Essentially, we're using the result structure to pass per field unexpected value errors through a chain; and then testing for them at the end.  This is quite complex though, so if we do want this we need some better guidance.

If we want chains to sometimes return a different result that's not an error; then it becomes quite hard I think.  Plugins need to pass everything through by default, because they can't rely on a later plugin being able to regenerate any specific field.

I think it all works, but I do think we need text clarifications.  I'm happy to suggest some text if we can discuss what the meaning is.

SPEC.md
```diff
@@ -85,20 +85,40 @@ The operations that CNI plugins must support are:
...
+  - A runtime may call `GET` at any time; but if `GET` is called for a given container before an `ADD` or after a `DEL` for that container, the plugin should return error 3 to indicate the container is unknown (see [Well-known Error Codes](#well-known-error-codes) section).
+  - 
+  - Parameters:
+    - **Container ID**, as defined for `ADD`.
+    - **Network namespace path**, as defined for `ADD`.
+    - **Network configuration**, as defined for `ADD`.
```
matthewdupre: Is it worth mentioning that prevResult is also set for the first plugin in the chain?

SPEC.md
```diff
@@ -273,8 +293,8 @@ The list is described in JSON form, and can be stored on disk or generated from
...
 
-For the ADD action, the runtime MUST also add a `prevResult` field to the configuration JSON of any plugin after the first one, which MUST be the Result of the previous plugin (if any) in JSON format ([see below](#network-configuration-list-runtime-examples)).
-For the ADD action, plugins SHOULD echo the contents of the `prevResult` field to their stdout to allow subsequent plugins (and the runtime) to receive the result, unless they wish to modify or suppress a previous result.
+For the `ADD` and `GET` actions, the runtime MUST also add a `prevResult` field to the configuration JSON of any plugin after the first one, which MUST be the Result of the previous plugin (if any) in JSON format ([see below](#network-configuration-list-runtime-examples)).
```
matthewdupre: Should mention the chain prevResult to first plugin behavior on GET here.

SPEC.md
dcbw: note: matt suggests stating that error 3 means the runtime doesn't need to do anything to clean up/get rid of the container, while other errors may require the runtime to call DEL to clean things up.

SPEC.md
dcbw: @matthewdupre yes, it would be.

libcni/api.go
```diff
@@ -142,22 +159,104 @@ func (c *CNIConfig) AddNetworkList(list *NetworkConfigList, rt *RuntimeConf) (ty
...
+	if err = setCachedResult(result, list.Name, rt); err != nil {
+		return nil, err
+	}
+
+	return result, nil
+}
+
+// GetNetworkList executes a sequence of plugins with the GET command
+func (c *CNIConfig) GetNetworkList(list *NetworkConfigList, rt *RuntimeConf) (types.Result, error) {
+	cachedResult, err := getCachedResult(list.Name, list.CNIVersion, rt)
```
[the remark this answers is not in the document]
dcbw: @squeed done, for Get, GetList, Del, and DelList with testcases too.

dcbw:
@containernetworking/cni-maintainers PTAL thanks!

dcbw:
@containernetworking/cni-maintainers updated SPEC language.

jellonek:
Request for change is only for docstring update. Addition of docstrings in `invoke` is only nice to have.

squeed:
LGTM, provided:
- docstrings are updated (can be in a separate PR)
- CI green

pkg/invoke/delegate.go
```diff
@@ -37,6 +32,20 @@ func DelegateAdd(delegatePlugin string, netconf []byte) (types.Result, error) {
 	return ExecPluginWithResult(pluginPath, netconf, ArgsFromEnv())
 }
 
+func DelegateAdd(delegatePlugin string, netconf []byte) (types.Result, error) {
```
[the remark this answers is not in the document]
dcbw: @jellonek done

pkg/skel/skel.go
```diff
@@ -227,25 +262,25 @@ func (t *dispatcher) pluginMain(cmdAdd, cmdDel func(_ *CmdArgs) error, versionIn
 //
 // To let this package automatically handle errors and call os.Exit(1) for you,
 // use PluginMain() instead.
-func PluginMainWithError(cmdAdd, cmdDel func(_ *CmdArgs) error, versionInfo version.PluginInfo) *types.Error {
+func PluginMainWithError(cmdAdd, cmdGet, cmdDel func(_ *CmdArgs) error, versionInfo version.PluginInfo) *types.Error {
```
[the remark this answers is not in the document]
dcbw: @jellonek done

dcbw:
@containernetworking/cni-maintainers docstrings added to new public functions

SPEC.md
```diff
@@ -552,3 +736,4 @@ Error codes 1-99 must not be used other than as specified here.
 
 - `1` - Incompatible CNI version
 - `2` - Unsupported field in network configuration. The error message must contain the key and value of the unsupported field.
+- `3` - Container unknown or does not exist. This error implies the runtimes does not need to perform any container network cleanup (for example, calling the `DEL` action on the container).
```
bboreham: s/runtimes/runtime/

libcni/api.go
```diff
@@ -142,57 +159,181 @@ func (c *CNIConfig) AddNetworkList(list *NetworkConfigList, rt *RuntimeConf) (ty
...
+
+// AddNetworkList executes a sequence of plugins with the ADD command
+func (c *CNIConfig) AddNetworkList(list *NetworkConfigList, rt *RuntimeConf) (types.Result, error) {
+	result, err := c.addOrGetNetworkList("ADD", nil, list, rt)
+	if err != nil {
+		return nil, err
+	}
+
+	if err = setCachedResult(result, list.Name, rt); err != nil {
+		return nil, err
```
rosenhouse: tiny nit: could you add breadcrumbs to the error here, like you have on line 214.  e.g.
```go
return nil, fmt.Errorf("setting cached result: %s", err)
```
This way the error message that bubbles up to the runtime provides a bit more context on where it came from.

I don't think you need to do it in all the private function calls, just at the top-level exported functions which now have this new behavior.

libcni/api.go
```diff
@@ -142,57 +159,181 @@ func (c *CNIConfig) AddNetworkList(list *NetworkConfigList, rt *RuntimeConf) (ty
...
+	// GET was added in CNI spec version 0.4.0 and higher
+	if gtet, err := version.GreaterThanOrEqualTo(list.CNIVersion, "0.4.0"); err != nil {
+		return nil, err
+	} else if !gtet {
+		return nil, fmt.Errorf("configuration version %q does not support the GET command", list.CNIVersion)
+	}
+
+	cachedResult, err := getCachedResult(list.Name, list.CNIVersion, rt)
+	if err != nil {
+		return nil, err
```
rosenhouse: Again, a bit of `fmt.Errorf("get cached result: %s", err)` could go a long way when debugging from the runtime level.

jellonek:
So the last thing is this s/runtimes/runtime/.

SPEC.md
dcbw: @bboreham fixed

libcni/api.go
dcbw: @rosenhouse added

libcni/api.go
dcbw: @rosenhouse done

dcbw:
@containernetworking/cni-maintainers reposted with updates for review comments

PR #494, opened 2017-08-18
libcni: return error if Type is empty

dcbw:
A network config missing Type is pretty useless since it means we can't
find the plugin to execute.  So return an error if Type isn't given.

### DIFF
--- a/libcni/conf_test.go
+++ b/libcni/conf_test.go
@@ -37,7 +37,7 @@ var _ = Describe("Loading configuration from disk", func() {
 			configDir, err = ioutil.TempDir("", "plugin-conf")
 			Expect(err).NotTo(HaveOccurred())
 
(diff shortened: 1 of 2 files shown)


libcni/conf_test.go
```diff
@@ -37,7 +37,7 @@ var _ = Describe("Loading configuration from disk", func() {
 			configDir, err = ioutil.TempDir("", "plugin-conf")
 			Expect(err).NotTo(HaveOccurred())
 
-			pluginConfig = []byte(`{ "name": "some-plugin", "some-key": "some-value" }`)
+			pluginConfig = []byte(`{ "name": "some-plugin", "type": "foobar", "some-key": "some-value" }`)
```
squeed: Wanna rename this to "some-network" :-)

squeed:
One silly minor test naming suggestion, otherwise fine. Feel free to merge.

PR #498, opened 2017-08-22
Extend IPAM route struct to support more route types

aaithal:
### Summary
Only unicast routes are supported today via the IPAM `route` struct. This PR extends this by supporting other route types, specifically the blocking route types.  Related to PR https://github.com/containernetworking/plugins/pull/57

### Details
The `type` field can be used to specify the following 'blocking' routes:
* unreachable
* blackhole
* prohibit

The `type` is inferred as `unicast` by default.

The following type's are not supported:
* local
* broadcast
* throw
* nat
* anycast
* multicast

Example:
```
{
  ...
  "ipam": {
    "type": "host-local",
    "subnet": "10.22.0.0/16",
    "routes": [
      { "dst": "0.0.0.0/0" },
      { "dst": "169.254.169.254/32", "type":"blackhole" }
    ]
  }
}
```

### Testing
Ran the test suite without any errors. Also tested that I can add a 'blackhole' route. 

```
$ cat /etc/cni/net.d/10-mynet.conf
{
        "cniVersion": "0.3.0",
        "name": "mynet",
        "type": "bridge",
        "bridge": "cni0",
        "isGateway": true,
        "ipMasq": true,
        "ipam": {
                "type": "host-local",
                "subnet": "10.22.0.0/16",
                "routes": [
                        { "dst": "0.0.0.0/0" },
                        { "dst": "169.254.169.254/32", "type":"blackhole" }
                ]
        }
}
$ sudo CNI_PATH=$CNI_PATH ./docker-run.sh busybox ip route show
default via 10.22.0.1 dev eth0
10.22.0.0/16 dev eth0 scope link  src 10.22.0.37
blackhole 169.254.169.254

$ ./test.sh
Building API
Building reference CLI
Building plugins
...
Running tests without coverage profile generation...
ok      github.com/containernetworking/cni/libcni       5.707s  coverage: 88.0% of statements
ok      github.com/containernetworking/cni/pkg/invoke   6.886s  coverage: 97.1% of statements
ok      github.com/containernetworking/cni/pkg/skel     0.023s  coverage: 85.7% of statements
ok      github.com/containernetworking/cni/pkg/types    0.019s  coverage: 44.3% of statements
ok      github.com/containernetworking/cni/pkg/types/current    0.012s  coverage: 26.9% of statements
ok      github.com/containernetworking/cni/pkg/types/020        0.005s  coverage: 32.4% of statements
ok      github.com/containernetworking/cni/pkg/version  0.009s  coverage: 75.7% of statements
ok      github.com/containernetworking/cni/pkg/version/testhelpers      1.348s  coverage: 76.2% of statements
ok      github.com/containernetworking/cni/plugins/test/noop    1.044s  coverage: 0.0% of statements
Checking gofmt...
Checking govet...
Checking license header...
Success
```

### DIFF
--- a/pkg/types/020/types_test.go
+++ b/pkg/types/020/types_test.go
@@ -46,6 +46,10 @@ var _ = Describe("Ensures compatibility with the 0.1.0/0.2.0 spec", func() {
 		Expect(routev6).NotTo(BeNil())
 		Expect(routegwv6).NotTo(BeNil())
 
+		blackholeIPv4, err := types.ParseCIDR("10.8.11.2/32")
+		Expect(err).NotTo(HaveOccurred())
+		Expect(blackholeIPv4).NotTo(BeNil())
+
 		// Set every field of the struct to ensure source compatibility
 		res := types020.Result{
 			CNIVersion: types020.ImplementedSpecVersion,
@@ -54,6 +58,7 @@ var _ = Describe("Ensures compatibility with the 0.1.0/0.2.0 spec", func() {
 				Gateway: net.ParseIP("1.2.3.1"),
 				Routes: []types.Route{
 					{Dst: *routev4, GW: routegwv4},
+					{Dst: *blackholeIPv4, Type: "blackhole"},
 				},
 			},
 			IP6: &types020.IPConfig{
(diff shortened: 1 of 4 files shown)


squeed:
This seems like a good idea. However, it will need a spec change.

We also need to think about downconversion. What's a good strategy to convert this to a v0.3.1 datatype? We can't just drop the "type" field, because the types added here are supposed to be unreachable. Perhaps we could set some kind of invalid gw?

I know that @dcbw wanted to consider changing the route structure for something else (IIRC, removing the `gateway` property of the IP struct).

squeed:
Also, question: Is there a real-world need for the different types of unreachable routes? These types just differ in how the application is notified that their packet was dropped.

squeed:
It also should be said that this is not sufficient for security, if that's what you're trying to do. Docker containers are started by default with `CAP_NET_RAW`, so they can send any packet they like, regardless of the routing table on the container.

So, with that in mind, what's the use case?

aaithal:
@squeed Thanks for taking a look at this.

> So, with that in mind, what's the use case?

The use-case is "security" as you mentioned. We want users of [ECS CNI Plugins](https://github.com/aaithal/amazon-ecs-cni-plugins-1/blob/9360d3a29029d3e60300d9dad0cc7b8447dcd372/README.md) to be able to add blocking routes in their containers when [provisioning containers with ENIs](https://github.com/aaithal/amazon-ecs-agent/blob/cec1ece6d619aaafe4b485cb4b74e1aafa786428/proposals/eni.md). Specifically, we want to allow customers to block access to the EC2 Instance Metadata Service, which is hosted at `169.254.169.254` for all containers as that gives containers access to the IAM Role credentials of the EC2 Instance and we have had customers wanting to block that. You're right to point out that containers are launched by default with `NET_RAW` and we need to deal with that as well. For this mode, we will allow users to drop these capability so that their routing rules are effective.

> Also, question: Is there a real-world need for the different types of unreachable routes? These types just differ in how the application is notified that their packet was dropped.

I don't think so. But, it seemed like an easy thing thing to add at relatively less cost. I'm open to suggestions on better ways of doing this.

> This seems like a good idea. However, it will need a spec change.

I can work making that change if you confirm that you (and other maintainers) are amenable to that.

> What's a good strategy to convert this to a v0.3.1 datatype? We can't just drop the "type" field, because the types added here are supposed to be unreachable. Perhaps we could set some kind of invalid gw?

Just to clarify what you're asking here, let's we support `type` in spec version `0.3.2`. Is your question how we'd make that effective in `0.3.1` result without the JSON emitting the `type` parameter?

aaithal:
@squeed closing this PR as per offline conversation. The gist was that it's better to do "security" things outside of the container.

squeed:
@aaithal Thanks nevertheless for the well-written PR. If you do decide you do want to do blackhole routing specifically, it's a good use case for a chained plugin.

mhristache:
Was there any follow up on support for blackhole routes? I am currently using an init container to add blackhole routes but would be great to replace that with CNI functionality as the container needs privileges. Thanks

PR #500, opened 2017-08-30
Use more RFC2119 style language in specification (must, should...)

matthewdupre:
I've erred on the side of doing less rather than more here, but a few things struck me while making the change.  I don't think this PR is ready yet; but I think it's far enough for discussion.

The specification does leave quite a lot of wiggle room.  I haven't tried to remove this, because it's more than just a clarification - the general theme is that there's quite a lot of flexibility in interpretation and use of the various parameters, especially around extending the core spec.  Interoperability is partially sacrificed for flexibility (and ease of implementation) - this is made clear in conventions.md, particularly.

Before going all the way down this road, I think a position on how precise the spec should be is needed.


SPEC.md
```diff
@@ -27,38 +27,37 @@ For the purposes of this proposal, we define two terms very specifically:
...
 ## General considerations
 
-The intention is for the container runtime to first create a new network namespace for the container.
-It then determines which networks this container should belong to and for each network, which plugin must be executed.
-The network configuration is in JSON format and can easily be stored in a file.
-The network configuration includes mandatory fields such as "name" and "type" as well as plugin (type) specific ones.
-The network configuration allows for fields to change values between invocations. For this purpose there is an optional field "args" which should contain the varying information.
-The container runtime sequentially sets up the networks by executing the corresponding plugin for each network.
-Upon completion of the container lifecycle, the runtime executes the plugins in reverse order (relative to the order in which they were added) to disconnect them from the networks.
+- The container runtime must create a new network namespace for the container before invoking any plugins.
```
dcbw: This doesn't really apply for hypervisors, but we can fix that later.

PR #501, opened 2017-08-30
Test all non-vendored packages

rosenhouse:


rosenhouse:
@containernetworking/cni-maintainers we no longer have to remember to add new packages to the `test.sh` script.  For example  `pkg/version/legacy_examples` included tests, but they weren't being run.

See also: https://github.com/containernetworking/plugins/pull/67

dcbw:
lgtm

PR #502, opened 2017-09-07
skel: VERSION shouldn't block on stdin

anfernee:
Right now, I have to do the following
  $ echo | CNI_COMMAND=VERSION bridge
or
  CNI_COMMAND=VERSION bridge
  ^D
to be able to get the result back.

But VERSION doesn't need to read from stdin.

### DIFF
--- a/pkg/skel/skel_test.go
+++ b/pkg/skel/skel_test.go
@@ -247,20 +247,18 @@ var _ = Describe("dispatching to the correct callback", func() {
 			Entry("path", "CNI_PATH", false),
 		)
 
-		Context("when the stdin is empty", func() {
-			BeforeEach(func() {
-				dispatch.Stdin = strings.NewReader("")
-			})
+		It("does not read from Stdin", func() {
+			r := &BadReader{}
+			dispatch.Stdin = r
 
-			It("succeeds without error", func() {
-				err := dispatch.pluginMain(cmdAdd.Func, cmdDel.Func, versionInfo)
+			err := dispatch.pluginMain(cmdAdd.Func, cmdDel.Func, versionInfo)
 
-				Expect(err).NotTo(HaveOccurred())
-				Expect(stdout).To(MatchJSON(`{
-					"cniVersion": "0.3.1",
-					"supportedVersions": ["9.8.7"]
+			Expect(err).NotTo(HaveOccurred())
+			Expect(r.ReadCount).To(Equal(0))
+			Expect(stdout).To(MatchJSON(`{
+				"cniVersion": "0.3.1",
+				"supportedVersions": ["9.8.7"]
 			}`))
-			})
 		})
 	})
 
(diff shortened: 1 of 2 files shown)


bboreham:
What's the test change?

pkg/skel/skel_test.go
```diff
@@ -249,11 +252,22 @@ var _ = Describe("dispatching to the correct callback", func() {
...
-				dispatch.Stdin = strings.NewReader("")
+				r, w := io.Pipe()
+				doneCh = make(chan struct{})
+				go func() {
+					select {
+					case <-time.After(time.Second):
+						w.CloseWithError(errors.New("timeout"))
+					case <-doneCh:
+						w.Close()
+					}
```
rosenhouse: Consider modifying the `It` block below to use [Ginkgo's build-in `done` channel feature](http://onsi.github.io/ginkgo/#asynchronous-tests) instead.  Then you can remove most of this setup in the `BeforeEach` and instead just write

```go
BeforeEach(func() {
   dispatch.Stdin, _ = io.Pipe()
})

It("succeeds without error", func(done Done) {
 // ... existing test code ...
 close(done)
}
```

anfernee: great to know that. thanks!

anfernee: btw, what if i still want to close the pipe on failure case? I don't want the test to block forever for error case.

rosenhouse: @anfernee I'm not sure I understand your question.  If your test looks like this:
```go
Context("when the stdin is empty", func() {
	BeforeEach(func() {
		dispatch.Stdin, _ = io.Pipe()
	})

	It("succeeds without error", func(done Done) {
		err := dispatch.pluginMain(cmdAdd.Func, cmdDel.Func, versionInfo)

		Expect(stdout).To(MatchJSON(`{
			"cniVersion": "0.3.1",
			"supportedVersions": ["9.8.7"]
         	}`))
		close(done)
	})
})
```
and you make no changes to the library code, then the test will fail after 1 second with a `Timed out` error.  That timeout comes from the test runner itself.

Once you fix the library code by adding your `if cmd == ...` lines, then the test passes.  At no point does the test block forever.

If you're asking about resource leaks related to the pipe itself, I wouldn't worry about it.  When the testing processes are torn down, the OS will release the pipe.  It's ok to leak resources during a short-lived test.

anfernee: Sorry, I didn't make it clear. What you said is exactly correct. What I meant is I want the negative case (without the fix) to fail the test (`timeout` thrown by test runner, but it ends the test with failure), not blocking forever. When somebody breaks it, he/she knows what is broken. However, if you feel it doesn't matter, it's fine as well. It's simpler and less code and I like simpler and less code.

rosenhouse: I absolutely think it is important to get thorough test coverage.  What "negative case" is missing?

anfernee:
@rosenhouse I modified the test case to test the input stdin is never read. so it doesn't matter if it's empty reader, bad reader, etc.

anfernee:
thanks @rosenhouse can you merge the change?

rosenhouse:
that would require an approval by at least one other of the @containernetworking/cni-maintainers

dcbw:
LGTM

PR #508, opened 2017-09-21
cnitool: generate container id from the netns path

squeed:
We shouldn't be creating networks with a blank containerid. Let's synthesize one from the netns path.


dcbw:
LGTM

squeed:
@bboreham updated the commit message. I'll merge this next business day unless you have complaints :-)

PR #510, opened 2017-09-22
spec, skel: Make the container ID required and unique.

squeed:
Previously, the spec did not require the container id to be set. However, almost every plugin relies on it being unique, including the CNI-maintained plugins. So, change the spec to require the use of container id and that plugins should treat it as a primary key.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1,12 +1,12 @@
 # Container Networking Interface Specification
 
 ## Version
-This is CNI **spec** version **0.3.1**.
+This is CNI **spec** version **0.3.1-dev**. This spec contains **unreleased** changes.
 
 Note that this is **independent from the version of the CNI library and plugins** in this repository (e.g. the versions of [releases](https://github.com/containernetworking/cni/releases)).
 
-#### Older versions
-Older versions of the spec are available as Git tags.
+#### Released versions
+Released versions of the spec are available as Git tags.
 
 | tag                                                                                  | spec permalink                                                                        | major changes                     |
 | ------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------- | --------------------------------- |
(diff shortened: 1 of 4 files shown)


SPEC.md
```diff
@@ -47,12 +47,14 @@ https://docs.google.com/a/coreos.com/document/d/1CTAL4gwqRofjxyp4tTkbgHtAwb2YCcP
 - Upon completion of the container lifecycle, the runtime must execute the plugins in reverse order (relative to the order in which they were executed to add the container) to disconnect the container from the networks.
 - The container runtime must not invoke parallel operations for the same container, but is allowed to invoke parallel operations for different containers.
 - The container runtime must order ADD and DEL operations for a container, such that ADD is always followed by a corresponding DEL. DEL may be followed by additional DELs, however, and plugins should handle multiple DELs permissively (i.e. plugin DEL should be idempotent).
+- A container is uniquely identified by a ContainerID. Plugins that store state should do so using a primary key of `(network name, container id)`. 
+- A runtime must not call ADD twice (without a corresponding DEL) for the same `(network name, container id)`. In other words, a given container ID must be added to a specific network exactly once.
```
rosenhouse: Interesting.  I can't think of a good use case for a container to have two interfaces on the same network.  But nevertheless, the spec currently allows it.  I wonder if this deserves a version bump.

dcbw: We can bundle it in with GET for v0.4.0

dcbw: Hmm, what about bond-type use-cases.  Assuming there are two physical interfaces placed into the container and then bonded together.  This could be done by a "master" plugin that delegates to two "physical device" plugins, and in that case the delegated plugin would get the same name/id tuple twice (but perhaps different config).

squeed: Given that the container ends up with a single interface with a single address scope (i.e. the bonded ifs don't have addresses), it seems that bonding would be a single CNI configuration.

SPEC.md
```diff
@@ -47,12 +47,14 @@ https://docs.google.com/a/coreos.com/document/d/1CTAL4gwqRofjxyp4tTkbgHtAwb2YCcP
 - Upon completion of the container lifecycle, the runtime must execute the plugins in reverse order (relative to the order in which they were executed to add the container) to disconnect the container from the networks.
 - The container runtime must not invoke parallel operations for the same container, but is allowed to invoke parallel operations for different containers.
 - The container runtime must order ADD and DEL operations for a container, such that ADD is always followed by a corresponding DEL. DEL may be followed by additional DELs, however, and plugins should handle multiple DELs permissively (i.e. plugin DEL should be idempotent).
+- A container is uniquely identified by a ContainerID. Plugins that store state should do so using a primary key of `(network name, container id)`.
```
matthewdupre: Please change is to must.

squeed:
There was some concern that ADD has been used idempotently, and that this is a spec change. By my reading, line 49 already implies that ADD is *not* idempotent, and this is not a spec change in that regard.

rosenhouse:
one thought we discussed at maintainers meeting: bump the spec version to `-dev` or something.  otherwise the merged version of this file would be lying when it says spec version `0.3.1`.

squeed:
@rosenhouse added some "this is a development version" language.

PR #511, opened 2017-09-25
the second cni plugin will fail with repeated network name

lht1:
I have tried the feature of Kubernetes(v1.7.3) that supports multiple network plugins recently. But I have got a problem.  The part of logs like follows:
        cni.go:294] Error adding network: container veth name provided (eth0) already exists
        cni.go:243] Error while adding to cni network: container veth name provided (eth0) already exists
    I check the relative source code and find that the CNI function AddNetworkList (from containernetworking/cni/libcni/api.go) always uses default IfName for every plugin. The piece of code following:
        for _, net := range list.Plugins {
            ......
            prevResult, err = invoke.ExecPluginWithResult(pluginPath, newConf.Bytes, c.args("ADD", rt))
            ......
        }
    I think the field rt.IfName should change with loop.

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -16,7 +16,9 @@ package libcni
 
 import (
 	"os"
+	"strconv"
 	"strings"
+	"unicode"
 
 	"github.com/containernetworking/cni/pkg/invoke"
 	"github.com/containernetworking/cni/pkg/types"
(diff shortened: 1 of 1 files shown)


lht1:
Most of the CNI plugins will use rt.IfName as network interface's name directly. When multiple plugins configured in the configuration file with '.conflist' as extension. The parameter rt.IfName equals 'some-eth0' for the first plugin. The same one for the second plugin and also for others. It is unreasonable, is it?

bboreham:
The point of calling a list of plugins is to do multiple things to the same interface - e.g. create it, add a port-mapping, add traffic shaping.

If you want to add multiple interfaces the caller needs to make independent CNI calls, with a different interface name each time.

bboreham:
Thanks for taking an interest; sorry we can't merge this.

PR #514, opened 2017-10-23
Remove vendor directory

rosenhouse:
> [Libraries should never vendor their dependencies](https://peter.bourgon.org/go-best-practices-2016/#dependency-management)

We were only vendoring packages for test support anyhow

### Remove build.sh
- not much left to build, now that we've split out plugins
- for non-AMD64 platforms, travis just runs `go build` directly

### Clean up test.sh
- no vendor --> simpler logic for enumerating packages
- use ["bash strict mode"](http://redsymbol.net/articles/unofficial-bash-strict-mode)
- remove unnecessary coveralls token, [not required for public repos](https://github.com/mattn/goveralls#github-integration)

### Simplify `.travis.yml`, update to work without vendor
- fewer env vars
- just `go get -t ./...` for test dependencies 

### Clean up Vagrantfile
- goodbye `godep restore`!
- bump Golang version 1.9.2


squeed:
Strong approve.

rosenhouse:
@squeed @matthewdupre All green, so many deletes!

PR #521, opened 2017-11-15
backfill some unit tests

rosenhouse:
bonus:  non-pointer instances of `types.Route` now correctly marshal to JSON


squeed:
Nice.

Issue #523, opened 2017-12-04
Add ability to validate a CNI configuration to libcni


squeed:
It would be nice to have a function that took a CNI configuration and ensured it was reasonably executable:

- It compiles
- All the relevant plugins are on disk and executable
- They all support the correct cniVersion

We already have most of this code, we should just expose it as a public function.

nmiculinic:
Is there some CLI to validate CNI config?

squeed:
@nmiculinic no, but that would be a great addition to `cnitool`!

PR #524, opened 2017-12-06
libcni: add ValidateNetwork and ValidateNetworkList functions

squeed:
These perform some static validation on a network configuration, ensuring it is generally executable.

Fixes: #523

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -71,6 +71,9 @@ type CNI interface {
 	AddNetwork(net *NetworkConfig, rt *RuntimeConf) (types.Result, error)
 	GetNetwork(net *NetworkConfig, rt *RuntimeConf) (types.Result, error)
 	DelNetwork(net *NetworkConfig, rt *RuntimeConf) error
+
+	ValidateNetworkList(net *NetworkConfigList) ([]string, error)
+	ValidateNetwork(net *NetworkConfig) ([]string, error)
 }
 
 type CNIConfig struct {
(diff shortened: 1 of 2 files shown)


squeed:
From the CNI meeting: missing capabilities should be treated differently - they are not a stop-the-world error. I'll update this accordingly.

dcbw:
LGTM with the exception of hte capabilities thing you mention above.

libcni/api.go
```diff
@@ -336,6 +339,85 @@ func (c *CNIConfig) DelNetwork(net *NetworkConfig, rt *RuntimeConf) error {
...
+	for _, net := range list.Plugins {
+		if err := c.validatePlugin(net.Network.Type, version); err != nil {
+			errs = append(errs, err)
+		}
+		for cap, enabled := range net.Network.Capabilities {
+			if !enabled {
+				continue
+			}
+
+			if missingCaps != nil {
```
dcbw: missingCaps will always be !nil

libcni/api.go
```diff
@@ -336,6 +339,85 @@ func (c *CNIConfig) DelNetwork(net *NetworkConfig, rt *RuntimeConf) error {
...
+				if _, ok := missingCaps[cap]; ok {
+					delete(missingCaps, cap)
+				}
+			}
+		}
+	}
+
+	// scan for missing capabilities
+	caps := []string{}
+	if len(missingCaps) != 0 {
```
dcbw: don't need this len() != 0 either, the range will just not execute if there aren't any missingCaps.

libcni/api.go
```diff
@@ -336,6 +339,78 @@ func (c *CNIConfig) DelNetwork(net *NetworkConfig, rt *RuntimeConf) error {
...
+// - all the specified plugins exist on disk
+// - every plugin supports the desired version.
+//
+// It returns a list of missing but desired capabilities and error if the configuration
+// is not executable. Runtimes should decide if they care about the missing capabilities
+// separately
+func (c *CNIConfig) ValidateNetworkList(list *NetworkConfigList, desiredCapabilities []string) ([]string, error) {
+	version := list.CNIVersion
+
+	missingCaps := map[string]interface{}{}
```
bboreham: The Go idiom here is `map[string]struct{}`

libcni/api.go
```diff
@@ -336,6 +339,78 @@ func (c *CNIConfig) DelNetwork(net *NetworkConfig, rt *RuntimeConf) error {
...
+			if !enabled {
+				continue
+			}
+			delete(missingCaps, cap)
+		}
+	}
+
+	// scan for missing capabilities
+	caps := []string{}
+	for cap, _ := range missingCaps {
```
bboreham: don't need `, _`

squeed:
Updated based on feedback. Rather than doing the set arithmetic ourselves, we just return a list of existing capabilities for the runtime to handle as it pleases.

PR #525, opened 2017-12-08
CONVENTIONS: add new `ipRanges` capability.

squeed:
Also, restructure the document slightly.

This convention is implemented in containernetworking/plugins#95

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -19,21 +19,18 @@ Establishing these conventions allows plugins to work across multiple runtimes. 
 # Current conventions
 Additional conventions can be created by creating PRs which modify this document.
 
-## Plugin specific fields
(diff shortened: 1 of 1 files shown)


bboreham:
This seems specific to the host-local plugin, so could just be in the docs for that plugin.  

Are there more general use-cases?

rosenhouse:
Also, this seems like a weird dependency inversion, where the runtime is doing the work of the IPAM plugin...?

squeed:
The motivating example is Kubernetes, which manages IP block allocation but delegates individual IP allocation to the respective plugins. This means that CNI plugins need to implement a Kubernetes API client.

From what I've seen, the IP block is often the only thing that differs between hosts in large installations, so by making it an argument that can be passed from the orchestrator, it simplifies the configuration greatly.

PR #526, opened 2017-12-12
kurma is no longer maintained

davidk01:
Kurma is a derelict project so should be removed because the link just redirects to apcera.com.


rosenhouse:
Thanks!

davidk01:
Thanks for merging.

PR #527, opened 2017-12-15
Adding the Bonding CNI plugin as 3rd party plugin

rkamudhan:
- [Bonding CNI](https://github.com/Intel-Corp/bond-cni) provides a method for aggregating multiple network interfaces into a single logical "bonded" interface.
- Linx Bonding drivers provides various flavour of bonded interface depending on the mode (bonding policies), such as round robin, active aggregation according to the 802.3 ad specification
For more information on the bonding driver. Please refer to [kernel doc](https://www.kernel.org/doc/Documentation/networking/bonding.txt)

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ To avoid duplication, we think it is prudent to define a common interface betwee
 - [Linen - a CNI plugin designed for overlay networks with Open vSwitch and fit in SDN/OpenFlow network environment](https://github.com/John-Lin/linen-cni)
 - [Vhostuser - a Dataplane network plugin - Supports OVS-DPDK & VPP](https://github.com/intel/vhost-user-net-plugin)
 - [Amazon ECS CNI Plugins - a collection of CNI Plugins to configure containers with Amazon EC2 elastic network interfaces (ENIs)](https://github.com/aws/amazon-ecs-cni-plugins)
-- [Bonding CNI - a Link aggregating plugin](https://github.com/Intel-Corp/bond-cni)
+- [Bonding CNI - a Link aggregating plugin to address failover and high availability network](https://github.com/Intel-Corp/bond-cni)
 
 The CNI team also maintains some [core plugins in a separate repository](https://github.com/containernetworking/plugins).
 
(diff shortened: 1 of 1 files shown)


rkamudhan:
Open source community project such as CNCF projects like CNI is used by the various segment of the Industries. It always cases that one segment use case or requirement doesn't make sense for other segments of the Industry. We can't say that CNCF projects support only one specific use cases. I can't argue with this statement, we end up in the rathole. 

In a technical point of view, Bonding is used for physical interfaces, it is supported in Linux driver. For example, readme file has SRIOV to provide that physical interfaces. In technical point, I don't feel there is need to comment whether it should not be a 3rd party plugin.

matthewdupre:
@rkamudhan Sorry if I wasn't clear.  By "Can we get a slightly longer description", I meant can you make the descriptive text in this change more informative - I'm not asking you to justify inclusion.

I'm not trying to say it shouldn't be merged because it doesn't apply to every CNI user - just that I'd like the list to give readers a reasonable chance of realizing if this plugin is likely to be interesting to them or not.

rkamudhan:
@matthewdupre Thanks for the clarification. I will update the readme file with more information

rkamudhan:
@matthewdupre Update the readme to understand use case around failover and high availability network. Here link for more information on the use cases - https://www.howtoforge.com/tutorial/how-to-configure-high-availability-and-network-bonding-on-linux/

matthewdupre:
I think the relationship between the physical and container interfaces is important - without SRIOV or something similarly it doesn't really make sense.

rkamudhan:
Can you clarify the statement? Bonding is used to create logical interface out of two interfaces, It can run with or without SRIOV. You run bond cni with flannel using `delegate` option in flannel to connect to the overlay network.

matthewdupre:
You could bond two veths, but it doesn't make much sense unless there are different physical interfaces backing them.

rkamudhan:
Bonding CNI is generic, you can give the different physical interface to it.

dcbw:
I'm find with the existing PR.

PR #529, opened 2018-01-24
Described Mesos as Apache project.

joerg84:


jellonek:
idk if it's right. Apache is a foundation. Apache Mesos - is a project name, but Mesos - is a brand name. Just look on http://mesos.apache.org/ site where you have: 
```
What is Mesos? A distributed systems kernel
Mesos is built using the same principles as the Linux kernel,
only at a different level of abstraction. The Mesos kernel ...
```
so even on official site this name is used as a standalone thing.

dcbw:
Seems like it is described officially as Apache Mesos in a couple places, even in the logo: http://mesos.apache.org/assets/img/mesos_logo.png

PR #534, opened 2018-02-01
Add vlan id

chenchun:
So it's possible for IP Allocator to return vlan ids of ips

### DIFF
--- a/pkg/types/current/types.go
+++ b/pkg/types/current/types.go
@@ -261,6 +261,7 @@ type IPConfig struct {
 	Interface *int
 	Address   net.IPNet
 	Gateway   net.IP
+	VlanId    uint16
 }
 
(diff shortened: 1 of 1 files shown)


hustcat:
I think we should add a scalable field for this. Such as:

```
type Result struct {
	CNIVersion string         `json:"cniVersion,omitempty"`
	Interfaces []*Interface   `json:"interfaces,omitempty"`
	IPs        []*IPConfig    `json:"ips,omitempty"`
	Routes     []*types.Route `json:"routes,omitempty"`
	DNS        types.DNS      `json:"dns,omitempty"`
        Envs       map[string]string `json:"envs,omitempty"`
}
```

squeed:
Closing for now  - feel free to reopen if you pick this back up.

PR #546, opened 2018-03-06
fixes the title in specification

mhausenblas:
Might be an historical artefact but I believe the correct full name for CNI should be Container Network Interface 

CC: @bboreham


squeed:
+2

PR #551, opened 2018-04-26
Network interface name matters

MikeSpreitzer:
This PR changes the SPEC so that state is kept per `(network name, container id, name of the interface inside the container)` tuple instead of simply per `(network name, container id)`.  In other words, this change allows one container to make multiple connections to the same network.  This came up as a desired use case in the Kubernetes network plumbing proto-WG.

I picked 0.4.0 as the new SPEC version, although 0.3.2 would be OK with me too.

I started working on updating the plugins, only to discover that they are not even consistent with the 0.3.1 spec (see https://github.com/containernetworking/plugins/issues/145).

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -47,8 +47,8 @@ https://docs.google.com/a/coreos.com/document/d/1CTAL4gwqRofjxyp4tTkbgHtAwb2YCcP
 - Upon completion of the container lifecycle, the runtime must execute the plugins in reverse order (relative to the order in which they were executed to add the container) to disconnect the container from the networks.
 - The container runtime must not invoke parallel operations for the same container, but is allowed to invoke parallel operations for different containers.
 - The container runtime must order ADD and DEL operations for a container, such that ADD is always followed by a corresponding DEL. DEL may be followed by additional DELs, however, and plugins should handle multiple DELs permissively (i.e. plugin DEL should be idempotent).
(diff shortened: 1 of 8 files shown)


squeed:
I'm not immediately opposed, but I don't understand the use case. A CNI configuration is supposed to logically represent a single "network," whatever that means. Why would a container need to be connected twice to the same network?

jellonek:
E.g. you have container connected with physical interface to network (it can be raw device, it can be SR-IOV) and you want to plan some work on physical wire, without loosing connectivity from container to resources provided by this network. Other case is when you want to control on container level link aggregation.
In general - that can be useful (or to be more precise - required) for NVFs, especially when they are using network interfaces on very low level directly.

It's strange case when you are looking on CNI from perspective of public clouds user, but from perspective of bare metal users, connection to physical networks, not only logical - it has much more sense.

squeed:
Can you replace the value inside the tuple with the phrase "interface name" (from "name of the interface inside the container")?
We may also wish to be clearer and use the values (name, CNI_CONTAINERID, CNI_IFNAME)

dcbw:
Plugins issue for this: https://github.com/containernetworking/plugins/issues/151

bboreham:
Rebased and updated at #557, in case you don't get round to it @MikeSpreitzer 
I included @squeed 's suggested change in the tuple.

bboreham:
replaced by #557

PR #554, opened 2018-05-22
skel: add support for plugin version string

squeed:
This adds an additional argument to PluginMain, the `about` field. It is emitted on stderr when CNI_COMMAND is not provided.

This is for the ease of administrators.


squeed:
Up for debate: is changing the api of `pkg/skel/PluginMain` to require an `about` string a reasonable breaking change? Or should I preserve the function signature?

PR #555, opened 2018-05-22
Adding timeout in the libcni tool

### LINKED ISSUE #550: Adding mechnism to handle timeout in the libcni
The `libcni` is a useful tool to invoke the plugin for setting up the container namespace, but today there isn't a convenient way to handle the `timeout` scenario. Ideally, if something unexpected is blocking the plugin setup process, it should fail itself after a timeout, and the process should be killed. And in order to kill the plugin process started by `libcni` [here](https://github.com/containernetworking/cni/blob/master/pkg/invoke/raw_exec.go#L42), we will need to know the pid of the process. So my proposal is to add a context in all the `libcni` api, and after the context was cancelled, libcni should kill the process and return a timeout error.

Please let me know your opinion and I can make the code change if this is acceptable to you.
Thanks,
Peng

richardpen:
Fix #550

### DIFF
--- a/pkg/invoke/raw_exec.go
+++ b/pkg/invoke/raw_exec.go
@@ -16,21 +16,66 @@ package invoke
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"os/exec"
+	"time"
 
 	"github.com/containernetworking/cni/pkg/types"
 )
+
+var ErrorPluginExecTimeout = fmt.Errorf("Waiting for plugin to return timed out")
 
 type RawExec struct {
 	Stderr io.Writer
 }
 
+func (e *RawExec) ExecPluginWithTimeout(pluginPath string, stdinData []byte, environ []string, timeout time.Duration) ([]byte, error) {
+	ctx, cancel := context.WithTimeout(context.TODO(), timeout)
+	defer cancel()
+	stdout := &bytes.Buffer{}
+	cmd, err := e.start(pluginPath, stdinData, stdout, environ)
+	if err != nil {
+		return nil, pluginErr(err, stdout.Bytes())
+	}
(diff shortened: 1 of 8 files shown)


squeed:
Interesting. Would it make more sense to pass the context all the way down to `RawExec` and handle it there? That was the suggestion in your original issue.

pkg/invoke/raw_exec.go
```diff
@@ -16,21 +16,66 @@ package invoke
...
+	cmdDone := make(chan error)
+	go func() {
+		cmdDone <- cmd.Wait()
+	}()
+
+	select {
+	case <-ctx.Done():
+		// kill the plugin process
+		if cmd.Process != nil {
+			cmd.Process.Kill()
```
squeed: you need to still receive the results from cmdDone here, or else you can get a null pointer exception.

richardpen:
@squeed The way I didn't choose to pass the context all the way down is that will require lots of the change to the interface as new methods `xxxwithContext` should be added.  Or  we can include the `context` in the `args` which I think is very similar as passing the time duration. Let me know if you have a better suggestion.

squeed:
@richardpen chatted about this quickly at the maintainer's meeting - we'd rather you just change the API. Because we do releases so slowly, we don't really worry about API stability. It makes more sense to just add the context to the full API.

This means you can get rid of ExecPluginWithTimeout.

pnovotnak:
Hey all, what can I do to help unstick this PR?

I'm running into https://github.com/kubernetes/kubernetes/issues/45419

dcbw:
I personally think https://github.com/containernetworking/cni/pull/568 is a more comprehensive approach, it just needs a couple changes.

dcbw:
We decided to merge https://github.com/containernetworking/cni/pull/568 which overlaps with this PR.

PR #557, opened 2018-05-30
Network interface name matters, updated

dcbw:
This is to fix up #551 against the latest code updates.


MikeSpreitzer:
@bboreham thanks for the update.

/LGTM
(not sure if that's formally proper or matters in this case.)

PR #558, opened 2018-06-01
libcni: make exec handling an interface for better downstream testing

dcbw:
When a downstream project just uses libcni, it takes a lot of harness
code to set up binaries on-disk that get called.  Instead we can handle
that by passing an execer interface to libcni APIs that gets called
instead, which the downstreams can use to avoid execing thing entirely
and instead check arguments/environment in code.

### DIFF
--- a/cnitool/cnitool.go
+++ b/cnitool/cnitool.go
@@ -95,9 +95,7 @@ func main() {
 	s := sha512.Sum512([]byte(netns))
 	containerID := fmt.Sprintf("cnitool-%x", s[:10])
 
-	cninet := &libcni.CNIConfig{
-		Path: filepath.SplitList(os.Getenv(EnvCNIPath)),
-	}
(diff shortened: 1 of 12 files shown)


dcbw:
@containernetworking/cni-maintainers now updated to handle delegate.go and add an example of how the exec faking would be used.  PTAL thanks!

dcbw:
@bboreham here's a real-world example of using the fake exec stuff with libcni.

FakeExec object:
https://github.com/dcbw/ocicni/blob/6f12dafc3df4b3b2eb9333ec78ad9b31ef6248ed/pkg/ocicni/ocicni_test.go#L38

FakeExec used with libcni:
https://github.com/dcbw/ocicni/blob/6f12dafc3df4b3b2eb9333ec78ad9b31ef6248ed/pkg/ocicni/ocicni_test.go#L322

PR #566, opened 2018-07-03
Add Juniper Contrail to 3rd party plugin list

dineshb-jnpr:
This change is to update the 3rd Party plugin section of README
to include and list the K8s CNI solution from Juniper Networks, INC.


dcbw:
LGTM

Issue #567, opened 2018-07-03
No timeout when  cni call  cni binary file.


liucimin:
**BUG REPORT**:

/kind bug

**What happened**:
Cni just use exec.Cmd to call the cni binary file.
In some cni plugin,such as contiv,the cni binary file may be locked and no return.
In some case, Such as in k8s,when kubelet first time call RunPodSandbox got a ctx timeout. the next time kubelet re-creates PodSandbox, it will always report this timeout.It because the kubelet ask a lock before call the cni plugin.

nikopen:
https://github.com/containernetworking/cni/pull/568 got merged, this is resolved I presume?

liucimin:
@nikopen  yes

PR #568, opened 2018-07-03
No timeout when cni call cni binary file.

liucimin:
#567
To fix this problem.
No timeout when cni call cni binary file.

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -15,6 +15,7 @@
 package libcni
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
@@ -64,16 +65,16 @@ type NetworkConfigList struct {
 }
 
 type CNI interface {
-	AddNetworkList(net *NetworkConfigList, rt *RuntimeConf) (types.Result, error)
-	GetNetworkList(net *NetworkConfigList, rt *RuntimeConf) (types.Result, error)
-	DelNetworkList(net *NetworkConfigList, rt *RuntimeConf) error
+	AddNetworkList(ctx context.Context, net *NetworkConfigList, rt *RuntimeConf) (types.Result, error)
+	GetNetworkList(ctx context.Context, net *NetworkConfigList, rt *RuntimeConf) (types.Result, error)
+	DelNetworkList(ctx context.Context, net *NetworkConfigList, rt *RuntimeConf) error
 
-	AddNetwork(net *NetworkConfig, rt *RuntimeConf) (types.Result, error)
-	GetNetwork(net *NetworkConfig, rt *RuntimeConf) (types.Result, error)
-	DelNetwork(net *NetworkConfig, rt *RuntimeConf) error
+	AddNetwork(ctx context.Context, net *NetworkConfig, rt *RuntimeConf) (types.Result, error)
+	GetNetwork(ctx context.Context, net *NetworkConfig, rt *RuntimeConf) (types.Result, error)
+	DelNetwork(ctx context.Context, net *NetworkConfig, rt *RuntimeConf) error
 
-	ValidateNetworkList(net *NetworkConfigList) ([]string, error)
-	ValidateNetwork(net *NetworkConfig) ([]string, error)
+	ValidateNetworkList(ctx context.Context, net *NetworkConfigList) ([]string, error)
+	ValidateNetwork(ctx context.Context, net *NetworkConfig) ([]string, error)
 }
 
 type CNIConfig struct {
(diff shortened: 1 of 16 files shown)


squeed:
Hi there,
There was an effort to solve this in #555, in a more generic way. As was mentioned in that PR, we would like to change the API to take a context from the runtime directly, rather than taking timeout. Hard-coding a timeout is not going to be OK.

Since that PR has stalled, you could do it instead :-).

liucimin:
@squeed 
Sorry, i did not see the old pr.
If the old pr is stalled,should i fix this by change the ExecPlugin api and not add a new api ExecPluginWithTimeout?

liucimin:
@squeed 
Hi ，Casey Callendrello
I have push the newest change in here~
Please to review the code and give me some suggestion~
Thx~

theRealWardo:
also very interested in this change as we've been suffering a lot of issues that I believe would be solved by this change. can someone please review this or revive #555 instead?

pnovotnak:
Sorry, but I'm going to chime in with a "me too" because this is a really serious bug... What can _I do_ to help push this forward?

libcni/api.go
```diff
@@ -39,6 +41,7 @@ type RuntimeConf struct {
 	ContainerID string
 	NetNS       string
 	IfName      string
+	Timeout     time.Duration
```
dcbw: Instead of having a Timeout here, just have callers of libcni pass their own Context (or nil) into the calls like AddNetworkList, DelNetworkList, etc.

libcni/api.go
```diff
@@ -72,8 +75,8 @@ type CNI interface {
 	GetNetwork(net *NetworkConfig, rt *RuntimeConf) (types.Result, error)
 	DelNetwork(net *NetworkConfig, rt *RuntimeConf) error
 
-	ValidateNetworkList(net *NetworkConfigList) ([]string, error)
-	ValidateNetwork(net *NetworkConfig) ([]string, error)
+	ValidateNetworkList(ctx context.Context, net *NetworkConfigList) ([]string, error)
```
dcbw: Per above, make AddNetwork/GetNetwork/DelNetowrk and their "List" functions do the same thing as ValidateNetwork[List]; just have callers pass a Context or nil.

libcni/api.go
```diff
@@ -172,8 +177,13 @@ func (c *CNIConfig) addOrGetNetwork(command, name, cniVersion string, net *Netwo
...
 		return nil, err
 	}
-
-	return invoke.ExecPluginWithResult(pluginPath, newConf.Bytes, c.args(command, rt), c.exec)
+	if rt.Timeout.Nanoseconds() > 0 {
+		ctx, cancel = context.WithTimeout(context.Background(), rt.Timeout)
+		defer cancel()
+	} else {
+		ctx = context.Background()
+	}
```
dcbw: If you do the above suggestion and plumb the Context down through to addOrGetNetwork, then you don't need to do any of this since the caller has already done it for you.

libcni/api.go
```diff
@@ -287,8 +299,14 @@ func (c *CNIConfig) delNetwork(name, cniVersion string, net *NetworkConfig, prev
...
 		return err
 	}
+	if rt.Timeout.Nanoseconds() > 0 {
+		ctx, cancel = context.WithTimeout(context.Background(), rt.Timeout)
+		defer cancel()
 
-	return invoke.ExecPluginWithoutResult(pluginPath, newConf.Bytes, c.args("DEL", rt), c.exec)
+	} else {
+		ctx = context.Background()
+	}
```
dcbw: Same here.

pkg/invoke/delegate.go
```diff
@@ -33,7 +34,7 @@ func delegateAddOrGet(command, delegatePlugin string, netconf []byte, exec Exec)
 		return nil, err
 	}
 
-	return ExecPluginWithResult(pluginPath, netconf, ArgsFromEnv(), exec)
+	return ExecPluginWithResult(context.TODO(), pluginPath, netconf, ArgsFromEnv(), exec)
```
dcbw: Do the same for these, have callers pass their own Context to DelegateAdd, DelegateGet, and DelegateDel()

liucimin: Ok,i will change this api interface~

jellonek:
This is very important (and needed) change, but it will break an api, so this will require a bigger bump in version during a release of lib.

dcbw:
@liucimin looks good to me, thanks!

pkg/invoke/delegate.go
dcbw: @liucimin any chance you can update the PR for this change?

dcbw:
@liucimin could you actually update delegate.go to pass the Context through, then I think everything is OK.  Thanks!

liucimin:
> @liucimin could you actually update delegate.go to pass the Context through, then I think everything is OK. Thanks!
@dcbw 
ok
i just have updated delegate.go to pass the Context through,.

dcbw:
> if the plugin delegates, will the context-cancelation propagate to the child-process? what if the IPAM plugin is what's hanging?

@rosenhouse at this point I don't think it will, we'd have to do something more complicated to pass that down.  Follow-up PR perhaps?

dcbw:
@liucimin thanks for the update, looks all good to me now.

PR #570, opened 2018-07-17
types/version: add helper to parse PrevResult

dcbw:
Plugins can use the new helpers to parse a prevResult.  When parsing
their NetConf structure, assuming it has a types.NetConf embedded:
```golang
  type MyConf struct {
    types.NetConf
    ...
  }

  conf := MyConf{}
  if err := json.Unmarshal(stdin, &conf); err != nil {
    return nil, fmt.Errorf("failed to parse network configuration: %v", err)
  }

  if err := version.ParsePrevResult(&conf.NetConf); err != nil {
    return nil, err
  }
```
and then when the plugin wishes to access the PrevResult it can transform
conf.PrevResult into any Result version it wants:
```golang
  if conf.PrevResult != nil {
    res, err := current.NewResultFromResult(conf.PrevResult)
    if err != nil {
      return nil, fmt.Errorf("could not convert result to current version: %v", err)
    }
    for _, ip := range res.IPs {
      ...
    }
  }
```

### DIFF
--- a/pkg/version/version_test.go
+++ b/pkg/version/version_test.go
@@ -1,0 +1,131 @@
+// Copyright 2018 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


pkg/version/version_test.go
```diff
@@ -0,0 +1,131 @@
+// Copyright 2016 CNI authors
```
jellonek: New file, so it probably should be 2018 ;)

dcbw: @jellonek fixed

dcbw:
Approved by bborham and jellonek so merging...

PR #572, opened 2018-07-26
skel: only print about message, not errors

dcbw:
Instead of:

```
CNI_COMMAND env variable missing
<about message here>
{
    "code": 100,
    "msg": "required env variables missing"
}
```

print:
`<about message here>`

@squeed @bboreham @jellonek @containernetworking/cni-maintainers


kevinsingapore:
and, how to solve? thank u

PR #574, opened 2018-08-03
List Knitter as 3rd party plugin

xiangpengzhao:
This PR adds Knitter to 3rd party plugin list in README. Knitter is a CNI plugin supporting multiple networking for Kubernetes.


dcbw:
Thanks!

xiangpengzhao:
Thanks Dan!

PR #576, opened 2018-08-08
Add name field in 99-loopback.conf because network name is required.

lifubang:
If there were no name field in 99-loopback.conf, there would be an error:
```
root@iZ2ze8rkx46zyx6mojyeb5Z:~/gocode/src/github.com/containernetworking/cni/scripts# ./priv-net-run.sh ifconfig
null : error executing ADD: missing network name
```


dcbw:
Thanks!

PR #578, opened 2018-08-13
when type is loopback, IP4 or IP6 is always nil

lifubang:
Signed-off-by: Lifubang <lifubang@aliyun.com>

issue: https://github.com/containerd/containerd/issues/2540

When type is loopback, and cniVersion is 0.2.0, the NewResultFromResult will return an error:
cannot convert: no valid IP addresses

I think we should not check IPs field in method convertFrom020. Let it check by user. Because all have been successful except this check.

Please check it, thanks.

### DIFF
--- a/pkg/types/current/types.go
+++ b/pkg/types/current/types.go
@@ -104,10 +104,6 @@ func convertFrom020(result types.Result) (*Result, error) {
 		}
 	}
 
-	if len(newResult.IPs) == 0 {
-		return nil, fmt.Errorf("cannot convert: no valid IP addresses")
-	}
-
 	return newResult, nil
 }
 


jellonek:
The bug is not returning an ip address in cni result at first by loopback plugin imo and not the conversion, which expects an ip address to be included in old type of result.
According to https://github.com/containernetworking/cni/blob/spec-v0.2.0/SPEC.md#ip-address-management-ipam-interface ip for both ip4 and ip6 should be set (with routes and gateway as optional fields).
So imo loopback plugin should be fixed so structure declared there: https://github.com/containernetworking/plugins/blob/master/plugins/main/loopback/loopback.go#L46 should be filled additionally with data about interface name (required by spec in versions more recent than 0.2.0) and ip addresses on the link.

lifubang:
@jellonek Do you know why not fix loopback result in https://github.com/containernetworking/plugins/blob/master/plugins/main/loopback/loopback.go#L46

dcbw:
On the face of it, I think you're right, the conversion code should not require any IP addresses in the Result structure.  I'll dig into it more since I wrote that code.

jellonek:
@dcbw any progress? :)

There will be almost identical issue with [l2-bridge plugin](https://github.com/containernetworking/plugins/pull/187) if one would set `cniVersion` to 0.2.0 in net config, so I think that this is another valid situation showing that this check should be removed.

PR #579, opened 2018-08-15
Replace GET action in spec with proposed CHECK action

matthewdupre:
This follows the CHECK spec we drafted in the maintainers call a couple of weeks ago.  

This provides a CHECK feature to make sure networking seems to be in a good state - it doesn't provide a mechanism for communicating changes in networking (e.g. a different DHCP leased IP).  That's something that will need a bit more thought and seems easier to allow later than to close the door on if we do it wrong.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -97,7 +97,7 @@ The operations that CNI plugins must support are:
     - **Extra arguments**, as defined for `ADD`.
     - **Name of the interface inside the container**, as defined for `ADD`.
   - Result:
-    - The plugin must return either a `Result` structuring matching the provided `prevResult`, or an error.
+    - The plugin must return either nothing or an error.
   - The plugin must consult the `prevResult` to determine the expected interfaces and addresses.
   - The plugin must allow for a later chained plugin to have modified networking resources, e.g. routes.
   - The plugin should return an error if an interface, address or route:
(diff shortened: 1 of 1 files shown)


matthewdupre:
@containernetworking/cni-maintainers ^^

matthewdupre:
Markups made

SPEC.md
```diff
@@ -85,27 +85,40 @@ The operations that CNI plugins must support are:
...
-  - The plugin should NOT return an error if its expected sandbox state (eg interfaces, IP addresses, routes, etc) is not found, as subsequent elements in the plugin's chain may alter sandbox state.
-  - A runtime may call `GET` at any time; but if `GET` is called for a container before an `ADD` or after a `DEL` for that container, the plugin should return error 3 to indicate the container is unknown (see [Well-known Error Codes](#well-known-error-codes) section).
-  - If the previous action for the container was `ADD` or `GET`, the runtime must add a `prevResult` field to the configuration JSON of the plugin (or all plugins in the chain), which must be the `Result` of that previous `ADD` or `GET` action in JSON format ([see below](#network-configuration-list-runtime-examples)).
+    - The plugin must return either nothing or an error.
+  - The plugin must consult the `prevResult` to determine the expected interfaces and addresses.
+  - The plugin must allow for a later chained plugin to have modified networking resources, e.g. routes.
```
rosenhouse: `must allow for` maybe could be clarified to `must not error on account of`?

rosenhouse: Also, should we clarify that the resources here are ones specifically tracked in the `prevResult`, and this does not include the "side-effect" resources that aren't tracked.

SPEC.md
```diff
@@ -85,27 +85,40 @@ The operations that CNI plugins must support are:
...
+  - The plugin should return an error if an interface, address or route:
+    - was created by the plugin, and
+    - is listed in `prevResult`, and
+    - does not exist, or is in an invalid state.
+  - The plugin should return an error if other expected resources such as the following do not exist or are in an invalid state:
```
rosenhouse: I wonder if we could give a name to these resources to more explicitly call out that they're not tracked in `prevResult`.  Maybe `untracked resources`?

SPEC.md
```diff
@@ -85,27 +85,40 @@ The operations that CNI plugins must support are:
...
+  - If there was a known previous `ADD` action for the container, the runtime MUST add a `prevResult` field to the configuration JSON of the plugin (or all plugins in a chain), which MUST be the `Result` of the immediately previous `ADD` action in JSON format ([see below](#network-configuration-list-runtime-examples)).  The runtime may wish to use libcni's support for caching `Result`s.
   - When `CNI_NETNS` and/or `prevResult` are not provided, the plugin should clean up as many resources as possible (e.g. releasing IPAM allocations) and return a successful response.
-  - If the runtime cached the `Result` of a previous `ADD` or `GET` response for a given container, it must delete that cached response on a successful `DEL` for that container.
+  - If the runtime cached the `Result` of a previous `ADD` response for a given container, it must delete that cached response on a successful `DEL` for that container.
 
-- `GET`: Get container network configuration
+- `CHECK`: Check container's networking is correct
```
bboreham: "correct" doesn't seem the right word.  Maybe "operational" or "as expected" ?

squeed:
since @matthewdupre is on vacation, I made the requested changes in this PR.

PR #580, opened 2018-08-15
Update cnitool.go

adivjoseph:
Added get support to cnitool

### DIFF
--- a/cnitool/cnitool.go
+++ b/cnitool/cnitool.go
@@ -35,6 +35,7 @@ const (
 
 	CmdAdd = "add"
 	CmdDel = "del"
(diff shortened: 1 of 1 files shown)


adivjoseph:
Add get command support

dcbw:
@adivjoseph thanks for the PR!  We're actually about to change how GET works (including its name) before the SPEC release, so this PR will need a bit of rework soon.  Once that change happens in the next couple of weeks, I'm happy to help explain what should be changed.

dcbw:
@adivjoseph we ended up adding CHECK support to cnitool.go as part of #579.

PR #583, opened 2018-08-27
Update Result.String implementation to work as declared in interface

### LINKED ISSUE #581: Inconsistent Result documentation
It looks like there are some inconsistencies in the `String()` function between the `Result` interface and struct. The interface is [documented](https://github.com/containernetworking/cni/blob/5ef0e9aa4cdb7c8d9647d81c93ae0c7139f19f71/pkg/types/types.go#L100) as returning the JSON representation of the result,  while the struct is [documented](https://github.com/containernetworking/cni/blob/47cf2da00befcca6fdd5d58ae044e399945c4b09/pkg/types/current/types.go#L213) as returning some formatted string that isn't JSON.

Not sure which direction should be fixed here, but the documentation should at least be consistent.

jellonek:
`String()` of `Result` declared in `pkg/types/types.go` should return a `Result`
representation in JSON, instead of custom formatted string, which
additionally is not in the same form as described in docstring, which also e.g. is missing `Routes` part.

This PR changes this behavior to be consistent with behavior documented in the interface.

As `String()` interface does not have any place for errors - if there
would be any - instead of JSON value output will be an empty string.

Closes #581

_edit/update:_
As discussed in comments - this PR removes funcs which are turning `Result` to string.

### DIFF
--- a/pkg/types/020/types.go
+++ b/pkg/types/020/types.go
@@ -86,20 +86,6 @@ func (r *Result) PrintTo(writer io.Writer) error {
 	return err
 }
 
-// String returns a formatted string in the form of "[IP4: $1,][ IP6: $2,] DNS: $3" where
-// $1 represents the receiver's IPv4, $2 represents the receiver's IPv6 and $3 the
-// receiver's DNS. If $1 or $2 are nil, they won't be present in the returned string.
-func (r *Result) String() string {
-	var str string
-	if r.IP4 != nil {
-		str = fmt.Sprintf("IP4:%+v, ", *r.IP4)
-	}
-	if r.IP6 != nil {
-		str += fmt.Sprintf("IP6:%+v, ", *r.IP6)
-	}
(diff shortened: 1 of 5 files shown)


rosenhouse:
Can we just delete this method?  How are people using this method today?

andrey-ko:
Probably types020.Result also should be updated for consistency: https://github.com/containernetworking/cni/blob/8fa373b4fd6889f59133dd82484a80c110bc4c10/pkg/types/020/types.go#L87

dcbw:
I vote for just removing the function entirely.

jellonek:
+1
Ok. So I'll update PR with removal of this func and update in Spec.

jellonek:
Bump. Funcs removed.
Please remember that it is a possibly breaking change, while we are not aware if anyone was using that.
Quick look on output of `git grep "\\bString"` in plugins repository doesn't show any uses of that in our code.

PR #584, opened 2018-09-13
skel,invoke,libcni: GET -> CHECK

dcbw:
Major change is that CHECK no longer returns a Result, per the spec.

@squeed @containernetworking/cni-maintainers @mcctcpip @jellonek

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -58,10 +58,11 @@ type NetworkConfig struct {
 }
 
 type NetworkConfigList struct {
-	Name       string
-	CNIVersion string
-	Plugins    []*NetworkConfig
-	Bytes      []byte
+	Name         string
+	CNIVersion   string
+	DisableCheck bool
+	Plugins      []*NetworkConfig
+	Bytes        []byte
 }
 
 type CNI interface {
@@ -287,6 +288,10 @@ func (c *CNIConfig) CheckNetworkList(ctx context.Context, list *NetworkConfigLis
 		return err
 	} else if !gtet {
 		return fmt.Errorf("configuration version %q does not support the CHECK command", list.CNIVersion)
+	}
+
+	if list.DisableCheck {
+		return nil
 	}
 
 	cachedResult, err := getCachedResult(list.Name, list.CNIVersion, rt)
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -72,6 +72,7 @@ type CNI interface {
 	AddNetwork(ctx context.Context, net *NetworkConfig, rt *RuntimeConf) (types.Result, error)
 	CheckNetwork(ctx context.Context, net *NetworkConfig, rt *RuntimeConf) error
 	DelNetwork(ctx context.Context, net *NetworkConfig, rt *RuntimeConf) error
+	GetNetworkCachedResult(net *NetworkConfig, rt *RuntimeConf) (types.Result, error)
 
 	ValidateNetworkList(ctx context.Context, net *NetworkConfigList) ([]string, error)
 	ValidateNetwork(ctx context.Context, net *NetworkConfig) ([]string, error)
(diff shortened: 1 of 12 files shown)


libcni/api.go
```diff
@@ -246,34 +218,75 @@ func getCachedResult(netName, cniVersion string, rt *RuntimeConf) (types.Result,
...
+		return err
 	} else if !gtet {
-		return nil, fmt.Errorf("configuration version %q does not support the GET command", list.CNIVersion)
+		return fmt.Errorf("configuration version %q does not support the CHECK command", list.CNIVersion)
 	}
 
 	cachedResult, err := getCachedResult(list.Name, list.CNIVersion, rt)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get network '%s' cached result: %v", list.Name, err)
+		return fmt.Errorf("failed to get network '%s' cached result: %v", list.Name, err)
```
jellonek: It's a good time to replace '%s' with %q. Same in line 257.

libcni/api.go
```diff
@@ -330,20 +343,20 @@ func (c *CNIConfig) AddNetwork(net *NetworkConfig, rt *RuntimeConf) (types.Resul
...
+		return err
 	} else if !gtet {
-		return nil, fmt.Errorf("configuration version %q does not support the GET command", net.Network.CNIVersion)
+		return fmt.Errorf("configuration version %q does not support the CHECK command", net.Network.CNIVersion)
 	}
 
 	cachedResult, err := getCachedResult(net.Network.Name, net.Network.CNIVersion, rt)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get network '%s' cached result: %v", net.Network.Name, err)
+		return fmt.Errorf("failed to get network '%s' cached result: %v", net.Network.Name, err)
```
jellonek: Same as above.

plugins/test/noop/main.go
```diff
@@ -212,5 +212,5 @@ func main() {
 	}
 
 	supportedVersions := debugGetSupportedVersions(stdinData)
-	skel.PluginMain(cmdAdd, cmdGet, cmdDel, version.PluginSupports(supportedVersions...), "CNI nnop plugin v0.7.0")
+	skel.PluginMain(cmdAdd, cmdCheck, cmdDel, version.PluginSupports(supportedVersions...), "CNI nnop plugin v0.7.0")
```
jellonek: Small typo: nnop -> noop.

dcbw:
@jellonek I think I've addressed all your comments, PTAL thanks!

Also, added a new commit with new GetNetwork[List]CachedResult() functions since we no longer have a GET command that returns it.

dcbw:
Travis error seems to be something Go + ARM specific:

```
$ go get -t ./...
package golang.org/x/net/html/charset: unrecognized import path "golang.org/x/net/html/charset" (parse https://golang.org/x/net/html/charset?go-get=1: no go-import meta tags ())
```

jellonek:
That import was not introduced in this PR so this looks like a flake probably because of response from `https://golang.org/x/net/html/charset?go-get=1`, so imo it's not even arm specific.
I would suggest to restart this test or simply ignore it's output. There is no way that this change would introduce anything so specific to go1.9 on arm.

squeed:
Looks like you need to implement `disableCheck`?

Does it make sense to always call `Check` for every `GetCached`?

dcbw:
> Looks like you need to implement `disableCheck`?

@squeed done

> Does it make sense to always call `Check` for every `GetCached`?

I thought about it and decided no.  A runtime may wish to get the cached result much more often than it would want to call check on an entire plugin chain.  I'd rather keep the operations separate for now.

squeed:
Makes sense. I think this is almost there - just one minor configuration question.

I'm having a bit of trouble grokking the integration test. Is it ensuring that CHECK always gets prevResult?

pkg/types/types.go
```diff
@@ -78,8 +78,9 @@ type IPAM struct {
 type NetConfList struct {
 	CNIVersion string `json:"cniVersion,omitempty"`
 
-	Name    string     `json:"name,omitempty"`
-	Plugins []*NetConf `json:"plugins,omitempty"`
+	Name         string     `json:"name,omitempty"`
+	DisableCheck bool       `json:"disableCheck"`
```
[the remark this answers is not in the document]
dcbw: Because then we have to do stuff like:

```
foo := true
list.DisableCheck = &foo
```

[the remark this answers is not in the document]
dcbw: But I forgot the omitempty, which I just added in the latest push.

dcbw:
Updated; @rhcambria pointed out that the caching code didn't account for the new 3-part uniqueness tuple we now have.  Fixed in a new commit.

dcbw:
Rebased to resolve conflicts.

PR #590, opened 2018-10-17
Move spec text about DEL and errors

bboreham:
This paragraph was under a heading specifically about configuration lists, but it applies to all DEL operations whether in a list or not.


s1061123:
Just a double check but this paragraph only mention golang error and CNI pluguin could allow to mention some error (resource leak, resource lock) as error message (to stderr), correct?

bboreham:
@s1061123 where do you see "golang" or "stderr" in this change ?

It says "Plugins should generally complete a `DEL` action without error", and gives some examples.

s1061123:
To me, "error" seems to be ambiguous a bit, whether "error message(of stderr)" or "error of golang"...

bboreham:
"error" in this context refers to the exit code of the process being non-zero.

PR #597, opened 2018-11-28
CONVENTIONS: add support for IP prefixes

dcbw:
While many plugins already have IPAM range info not all do. Without
a range/subnet in configuration, passing a plain IP address in
CNI_ARGS or via 'args' does not provide enough information.


s1061123:
@AlonaKaplan yeah, right. In case of IPv4/v6 dual stack case, user may want to set IPv4/v6 addresses at the same time, hence I suppose it should support to pass two or more IPs into args.

PR #602, opened 2018-12-06
fix the ParseVersion when the version is  will get fnAtoi

liucimin:
When i test the cni config.
I found that if the string is "" the split function return 1 length Slice not 0 length.
So the IF statement will not Effective.

image

and it will raise a fnAtoi because of we " major, err := strconv.Atoi(parts[0]) " atoi the parts[0].

I think will should add the statement before the split function.


dcbw:
@liucimin could you update the testcase in plugin_test.go to add "" as a version parse test?  Basically just do this:

```
		It("returns an error for malformed versions", func() {
-			badVersions := []string{"asdfasdf", "asdf.", ".asdfas", "asdf.adsf.", "0.", "..", "1.2.3.4.5"}
+			badVersions := []string{"asdfasdf", "asdf.", ".asdfas", "asdf.adsf.", "0.", "..", "1.2.3.4.5", ""}
			for _, v := range badVersions {
				_, _, _, err := version.ParseVersion(v)
			}
```

liucimin:
@dcbw ok，i have add the text code~

PR #627, opened 2019-02-22
Update deprecated link in doc

longkb:
Currently, the modified **plugins/main** was removed from master branch.
It's only available in **v0.3.0** tag. So this commit aims to update the
deprecated link to the working one.

Co-Authored-By: Nguyen Phuong An <AnNP@vn.fujitsu.com>
Signed-off-by: Kim Bao Long <longkb@vn.fujitsu.com>

### DIFF
--- a/Documentation/spec-upgrades.md
+++ b/Documentation/spec-upgrades.md
@@ -149,7 +149,7 @@ result, err := current.NewResultFromResult(ipamResult)
 ```
 
 Other examples of spec v0.3.0-compatible plugins are the
-[main plugins in this repo](https://github.com/containernetworking/cni/tree/master/plugins/main)
+[main plugins in this repo](https://github.com/containernetworking/plugins/)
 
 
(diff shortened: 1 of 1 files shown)


longkb:
@matthewdupre : Could you help me to review my PR :)
Thank you very much

bboreham:
I think the link should move to point at the plugins repo https://github.com/containernetworking/plugins/pulls.

Also part of the point of splitting the repos was that the plugins are released at a different schedule to the spec - plugins 0.5.2 were compatible with spec 0.3.0.

longkb:
> 
> 
> I think the link should move to point at the plugins repo https://github.com/containernetworking/plugins/.
> 
> Also part of the point of splitting the repos was that the plugins are released at a different schedule to the spec - plugins 0.5.2 were compatible with spec 0.3.0.

Thanks for your reviewing. I have updated my PR as your recommendation :)

PR #633, opened 2019-02-26
Remove example due to dead document link

hnwolf:
Update the document due to not available link.


dcbw:
@hnwolf could you update the link to point to the containernetworking/plugins repo on the 0.3.0 tag instead of removing the link?

dcbw:
Actually, this was already done https://github.com/containernetworking/cni/pull/627...

PR #641, opened 2019-03-13
SPEC.md: clarify meaning of 'routes'

dcbw:
Routes are expected to be relevant to the sandbox interface.


SPEC.md
```diff
@@ -720,6 +726,7 @@ All properties known to the plugin should be provided, even if not strictly requ
   ]
 ```
 
+Each `routes` entry must be relevant for the sandbox interface specified by CNI_IFNAME.
```
bboreham: The formatting of these two lines is wrong; the second is not subordinate to the first.  I suspect the best improvement would be to remove the `-` from the "is a dictionary" line.

PR #646, opened 2019-03-28
Plugin(manager) interfaces and functionality arround them

andrey-ko:
this PR adds `Plugin` and `PluginManager` interfaces and additional functionality in order to allow write plugins in component based way. 
`Plugin` interface represent abstraction of plugin or delegator. The example of plugin delegator is  cli-adapter (`CliPlugin`) that exposes Plugin interfaces and forwards calls to some executable. 
`PluginManager` is interface which allows to resolve other plugins. It have only one method `FindPlugin` that takes plugin-type as parameter and returns  `Plugin` interface in case of success. For example `CliPluginManager` will look for plugin executable on the `Path` and return cli-delegator on success.
Such decomposition very simplify writing unit tests and also dealing with non-trivial uses cases when for example you may need invoke plugins over RPC.

### DIFF
--- a/pkg/skel/adapter.go
+++ b/pkg/skel/adapter.go
@@ -1,0 +1,63 @@
+// Copyright 2014-2016 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


matthewdupre:
@andrey-ko could you please explain the motivation behind this more?  We can't review it or merge it until we understand the purpose - while the code looks properly written at a quick glance, it's better to discuss the requirements behind this kind of change before you sink time into writing it.

Things I'm interested in:
- What's the problem you're trying to solve?  Your message only says simplifies unit tests, but this currently only adds new tests and doesn't modify any existing ones, so which tests would be simplified?
- Where is this supposed to be called from?
- Does it / will it replace any existing code?  If so what (possibly outside this repo)?
- Is there some further enhancement you're looking for that this would help?

squeed:
Closing stale PR.

PR #647, opened 2019-03-28
update spec-upgrades.md make it point to a specific web page

xichengliudui:
Make it point to a specific web page


dcbw:
@xichengliudui since the text there talks about 0.3.0, can we link to a specific tag for 0.3.0 instead of master?

bboreham:
@dcbw the 0.3.0 here is _spec_ version; it would not be appropriate to link to 0.3.0 _plugins_ version.

PR #648, opened 2019-03-28
Remove invalid 404 links

xichengliudui:
I checked the other branches and did not find this file, so I think this file should have been deleted.


bboreham:
I think what happened is the `host-local` plugin moved to another repository. 
 The link should be https://github.com/containernetworking/plugins/tree/master/plugins/ipam/host-local#supported-arguments

xichengliudui:
@bboreham  Thank you for your reminding. I have modified it.

bboreham:
Thanks.  Can you squash the two commits into one please.

PR #652, opened 2019-04-09
Make link point to the correct version

SataQiu:
I found that `Library and Plugins Release v0.5.0` just linked to the latest release page.
The link address does not match the link description.
So I fixed it.


bboreham:
Thanks!  Why go with "after 0.5.1" rather than link to the specific version?
(https://github.com/containernetworking/cni/releases/tag/v0.5.0)

SataQiu:
Thanks for your feedback. @bboreham 
Your suggestion is very good. I have done it.

SataQiu:
@bboreham I think this PR can be merged. Is there anything else I need to do?

PR #661, opened 2019-04-30
libcni: add config caching

dcbw:
Cache the config JSON, CNI_ARGs, and CapabilityArgs on ADD operations,
and remove them on DEL. This allows runtimes to retrieve the cached
information to ensure that the pod is given the same config and options
on DEL as it was given on ADD.

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -69,11 +69,13 @@ type CNI interface {
 	AddNetworkList(ctx context.Context, net *NetworkConfigList, rt *RuntimeConf) (types.Result, error)
 	CheckNetworkList(ctx context.Context, net *NetworkConfigList, rt *RuntimeConf) error
 	DelNetworkList(ctx context.Context, net *NetworkConfigList, rt *RuntimeConf) error
+	GetNetworkListCachedConfig(net *NetworkConfigList, rt *RuntimeConf) ([]byte, *RuntimeConf, error)
 
 	AddNetwork(ctx context.Context, net *NetworkConfig, rt *RuntimeConf) (types.Result, error)
 	CheckNetwork(ctx context.Context, net *NetworkConfig, rt *RuntimeConf) error
 	DelNetwork(ctx context.Context, net *NetworkConfig, rt *RuntimeConf) error
 	GetNetworkCachedResult(net *NetworkConfig, rt *RuntimeConf) (types.Result, error)
+	GetNetworkCachedConfig(net *NetworkConfig, rt *RuntimeConf) ([]byte, *RuntimeConf, error)
 
 	ValidateNetworkList(ctx context.Context, net *NetworkConfigList) ([]string, error)
 	ValidateNetwork(ctx context.Context, net *NetworkConfig) ([]string, error)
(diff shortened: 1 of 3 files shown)


libcni/api.go
```diff
@@ -233,6 +242,99 @@ func (c *CNIConfig) GetNetworkCachedResult(net *NetworkConfig, rt *RuntimeConf)
...
+			return nil, nil, fmt.Errorf("failed to parse cached config CNI Args %q: %v", unmarshaled.CniArgs, err)
+		}
+	}
+	newRt.CapabilityArgs = unmarshaled.CapabilityArgs
+
+	return unmarshaled.Config, newRt, nil
+}
+
+// GetNetworkListCachedConfig returns the cached Config of the previous
+// previous AddNetworkList() operation for a network list, or an error.
```
bboreham: previous previous?

test.sh
```diff
@@ -4,11 +4,17 @@ set -euo pipefail
...
 cd "$(dirname $0)"
 
+go get -t ./...
+
 echo -n "Running tests "
 function testrun {
     bash -c "umask 0; PATH=$PATH go test $@"
 }
 if [ ! -z "${COVERALLS:-""}" ]; then
+    go get golang.org/x/tools/cmd/cover
```
bboreham: unrelated change?

libcni/api.go
```diff
@@ -233,6 +242,99 @@ func (c *CNIConfig) GetNetworkCachedResult(net *NetworkConfig, rt *RuntimeConf)
...
+			return nil, fmt.Errorf("invalid CNI_ARGS pair %q", pair)
+		}
+
+		result = append(result, [2]string{kv[0], kv[1]})
+	}
+
+	return result, nil
+}
+
+type cachedConfig struct {
```
squeed: For future compatability, this should be something of the form

```json
{
  "kind": "CniCacheV1",
  "value": {"cniVersion": "..."  }
}
```

squeed:
Hang on, should we really be caching the config json? I like caching the ephemeral parameters, but not the config. What if you want to update some portion of the configuration file?

dcbw:
> Hang on, should we really be caching the config json? I like caching the ephemeral parameters, but not the config. What if you want to update some portion of the configuration file?

@squeed caching config is the whole point. Do you really want to DEL a pod with different config than you ADD-ed it?

squeed:
> @squeed caching config is the whole point. Do you really want to DEL a pod with different config than you ADD-ed it? I struggle to think of a case where that's reasonable.

Funny, I would say: Definitely! Sometimes. Well, maybe sometimes. I can think of supporting cases for both.

If you change, say, the firewall chain, then you'd definitely want to delete using the old configuration. OTOH, if you change a database URL, then you'd want to use the new configuration. To me, caching the configuration for delete would be surprising.

dcbw:
> Funny, I would say: Definitely! Sometimes. Well, maybe sometimes. I can think of supporting cases for both.

@squeed OK I can buy changing a database URL or something, perhaps you're using a complex plugin and something there changed. But changing many other attributes would have a detrimental effect on the container. The problem is how do we know which ones?

At the very least, you shouldn't be able to change out the "name" and "type" keys between ADD + DEL. I'd argue you probably shouldn't change out the host-local ranges too, but that's pretty plugin-specific. You could argue that it should be possible to add/remove some non-core plugins (tuning, maybe firewall, etc) but honestly I'm not sure how we could determine that.

mars1024:
I think maybe we can just cache the network name which pod used in libcni/ocicni, and let cni config file and cni plugins to handle the consistency between ADD and DEL?

mars1024:
I think we can obey the rule that the new config can always DEL old pods which is set up by the old config with same name completely when we want to change cni configs，based on this，we can only cache the network name which pod used and trust the newest config with this name.

libcni/api.go
dcbw: @squeed Good point.

dcbw:
So @squeed and I talked about it for a while, and decided that yes there are use-cases for changing parts of a given network config underneath the container, but not for switching the network config out entirely. So much like @mars1024 is saying, a delete operation should probably only check network name (and perhaps plugin type) from the config its sending and make sure those match.

I think the config caching that this PR does is probably OK, since all it does is allow the runtime (dockershim, crio, etc) to retrieve that config. Then the runtime can make the decision about whether to use an existing updated configuration that it already knows about, or if that's no longer valid (eg old config deleted entirely) it could use the cached config.

Does that sound OK to everyone? eg libcni only does the storage/retrieval, while the runtime itself can make policy decisions about how to handle config changes.

dcbw:
Continued in https://github.com/containernetworking/cni/pull/678

PR #662, opened 2019-05-05
invoke : ensure custom envs of CNIArgs are prepended to process envs

mars1024:
From this commit [https://github.com/golang/go/commit/e73f4894949c4ced611881329ff8f37805152585#diff-2f929a7660c606fd0c5b268dad1a9bd4R61](https://github.com/golang/go/commit/e73f4894949c4ced611881329ff8f37805152585#diff-2f929a7660c606fd0c5b268dad1a9bd4R61) in golang, we can know that the RawExec will clean the duplicated envs and keep the last valid env of the same key.
So if we want to prepend custom args from CNIArgs rather than the existing process environments, we should append custom args in the end.

### DIFF
--- a/pkg/invoke/args.go
+++ b/pkg/invoke/args.go
@@ -15,6 +15,7 @@
 package invoke
 
 import (
+	"fmt"
 	"os"
 	"strings"
 )
@@ -67,7 +68,7 @@ func (args *Args) AsEnv() []string {
 		"CNI_IFNAME="+args.IfName,
 		"CNI_PATH="+args.Path,
 	)
-	return env
+	return dedupEnv(env)
 }
 
 // taken from rkt/networking/net_plugin.go
@@ -79,4 +80,27 @@ func stringify(pluginArgs [][2]string) string {
 	}
 
 	return strings.Join(entries, ";")
+}
+
+// dedupEnv returns a copy of env with any duplicates removed, in favor of later values.
+// Items not of the normal environment "key=value" form are preserved unchanged.
+func dedupEnv(env []string) []string {
+	out := make([]string, 0, len(env))
+	envMap := map[string]string{}
+
+	for _, kv := range env {
+		// find the first "=" in environment, if not, just keep it
+		eq := strings.Index(kv, "=")
+		if eq < 0 {
+			out = append(out, kv)
+			continue
+		}
+		envMap[kv[:eq]] = kv[eq+1:]
+	}
+
+	for k, v := range envMap {
+		out = append(out, fmt.Sprintf("%s=%s", k, v))
+	}
+
+	return out
 }
--- a/pkg/invoke/args.go
+++ b/pkg/invoke/args.go
@@ -57,16 +57,16 @@ func (args *Args) AsEnv() []string {
 		pluginArgsStr = stringify(args.PluginArgs)
 	}
 
-	// Ensure that the custom values are first, so any value present in
-	// the process environment won't override them.
-	env = append([]string{
-		"CNI_COMMAND=" + args.Command,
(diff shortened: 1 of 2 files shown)


mars1024:
another ref [https://github.com/golang/go/issues/12868](https://github.com/golang/go/issues/12868) , I think some code changes in golang Exec lead to this bug.

mars1024:
I think it's the original issue [https://github.com/containernetworking/cni/issues/410](https://github.com/containernetworking/cni/issues/410) related to this, if you have time, please take a look at this PR, thanks ~ @bboreham @rosenhouse

bboreham:
Thanks for the PR!
#410 was fixed by #461, but noted near the time to be broken again by a Go runtime change.
We should have opened another issue.
Wonder if you can change the test so it actually tests the end result?

mars1024:
> Thanks for the PR!
> #410 was fixed by #461, but noted near the time to be broken again by a Go runtime change.
> We should have opened another issue.
> Wonder if you can change the test so it actually tests the end result?

I don't think we can test the end result, because the deduplicate action happens in exec or delegate, not here, but I think there maybe another way, we can add a deduplicate action in func AsEnv to make sure there is no duplication in return.

mars1024:
I have add deduplication in envs return of CNIArgs and updated the tests. @bboreham

dcbw:
@mars1024 Instead of using an array in AsEnv, what about just decomposing the environment into a map, adding our CNI_ vars into the map, and recomposing it into the array we finally pass back?

```
envMap := make(map[string]string)
env := os.GetEnv()
pairs := strings.Split(args, ";")
for _, pair := range pairs {
	kv := strings.Split(pair, "=")
	if len(kv) != 2 || kv[0] == "" || kv[1] == "" {
		return nil, fmt.Errorf("invalid CNI_ARGS pair %q", pair)
	}
	envMap[kv[0]] = kv[1]
}

envMap["CNI_COMMAND"] = <xxxx>
envMap["CNI_CONTAINERID"] = <xxx>
```
I don't think we need to bother with case sensitivity because CNI defines the relevant enviornment variables that we are adding here as upper-case. If something bothers to look for "cni_command" then that thing is demonstrably wrong and not spec conformant. (unless something on Windows does env lookups case insensitively and grabs the first item in the list?)

mars1024:
> @mars1024 Instead of using an array in AsEnv, what about just decomposing the environment into a map, adding our CNI_ vars into the map, and recomposing it into the array we finally pass back?
> 
> ```
> envMap := make(map[string]string)
> env := os.GetEnv()
> pairs := strings.Split(args, ";")
> for _, pair := range pairs {
> 	kv := strings.Split(pair, "=")
> 	if len(kv) != 2 || kv[0] == "" || kv[1] == "" {
> 		return nil, fmt.Errorf("invalid CNI_ARGS pair %q", pair)
> 	}
> 	envMap[kv[0]] = kv[1]
> }
> 
> envMap["CNI_COMMAND"] = <xxxx>
> envMap["CNI_CONTAINERID"] = <xxx>
> 
> finalEnv := make([]string, 0, len(envMap))
> for k, v := range envMap {
>     finalEnv = append(finalEnv, fmt.Sprintf("%s=%s", k, v))
> }
> return finalEnv
> ```
> 
> I don't think we need to bother with case sensitivity because CNI defines the relevant enviornment variables that we are adding here as upper-case. If something bothers to look for "cni_command" then that thing is demonstrably wrong and not spec conformant. (unless something on Windows does env lookups case insensitively and grabs the first item in the list?)

OK, thanks for your comment, I will update the deduplication func.

mars1024:
Code updated! @bboreham @dcbw @squeed  Please go on reviewing, thanks.

PR #663, opened 2019-05-05
delegate : allow delegation funcs override CNI_COMMAND env automatically in heritance

mars1024:
Sometimes, we need a DEL exec for another CNI plugin in a main CNI plugin process, such as a ipam revert in ADD exec of macvlan plugin (ref to discussions in  [https://github.com/containernetworking/plugins/pull/301](https://github.com/containernetworking/plugins/pull/301)), in this scenario, I think delegation func should have the ability to override CNI_COMMAND env in process environments heritance automatically. This will help us to avoid setting and unsetting CNI_COMMAND env again and again.

### DIFF
--- a/pkg/invoke/args.go
+++ b/pkg/invoke/args.go
@@ -23,6 +23,8 @@ import (
 type CNIArgs interface {
 	// For use with os/exec; i.e., return nil to inherit the
 	// environment from this process
+	// For use in delegation; inherit the environment from this
+	// process and allow overrides
 	AsEnv() []string
 }
 
@@ -80,6 +82,26 @@ func stringify(pluginArgs [][2]string) string {
 	}
 
 	return strings.Join(entries, ";")
+}
+
+// DelegateArgs implements the CNIArgs interface
+// used for delegation to inherit from environments
+// and allow some overrides like CNI_COMMAND
+var _ CNIArgs = &DelegateArgs{}
+
+type DelegateArgs struct {
+	Command string
+}
+
+func (d *DelegateArgs) AsEnv() []string {
+	env := os.Environ()
+
+	// The custom values should come in the end to override the existing
+	// process environment of the same key.
+	env = append(env,
+		"CNI_COMMAND="+d.Command,
+	)
+	return dedupEnv(env)
 }
 
 // dedupEnv returns a copy of env with any duplicates removed, in favor of later values.
--- a/pkg/invoke/delegate.go
+++ b/pkg/invoke/delegate.go
@@ -16,20 +16,15 @@ package invoke
 
 import (
 	"context"
-	"fmt"
 	"os"
 	"path/filepath"
 
 	"github.com/containernetworking/cni/pkg/types"
 )
 
-func delegateCommon(expectedCommand, delegatePlugin string, exec Exec) (string, Exec, error) {
+func delegateCommon(delegatePlugin string, exec Exec) (string, Exec, error) {
 	if exec == nil {
 		exec = defaultExec
-	}
-
-	if os.Getenv("CNI_COMMAND") != expectedCommand {
-		return "", nil, fmt.Errorf("CNI_COMMAND is not " + expectedCommand)
 	}
 
 	paths := filepath.SplitList(os.Getenv("CNI_PATH"))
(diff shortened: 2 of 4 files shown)


pkg/invoke/args.go
```diff
@@ -80,3 +82,23 @@ func stringify(pluginArgs [][2]string) string {
...
+func (d *DelegateArgs) AsEnv() []string {
+	env := os.Environ()
+
+	// The custom values should come in the end to override the existing
+	// process environment of the same key.
+	env = append(env,
+		"CNI_COMMAND="+d.Command,
+	)
+	return env
+}
```
mccv1r0: Wouldn't it be simpler to just do:

```
diff --git a/pkg/invoke/delegate.go b/pkg/invoke/delegate.go
index 30b4672f11fc..785f63b501dd 100644
--- a/pkg/invoke/delegate.go
+++ b/pkg/invoke/delegate.go
@@ -44,6 +44,13 @@ func delegateCommon(expectedCommand, delegatePlugin string, exec Exec) (string,
 // DelegateAdd calls the given delegate plugin with the CNI ADD action and
 // JSON configuration
 func DelegateAdd(ctx context.Context, delegatePlugin string, netconf []byte, exec Exec) (types.Result, error) {
+       cmd := os.Getenv("CNI_COMMAND")
+       if cmd == "" {
+               return fmt.Errorf("environment variable CNI_COMMAND must be specified.")
+       }
+       // restore CNI_COMMAND to original value upon return.
+       defer os.Setenv("CNI_COMMAND", cmd)
+       os.Setenv("CNI_COMMAND", "ADD")
        pluginPath, realExec, err := delegateCommon("ADD", delegatePlugin, exec)
        if err != nil {
                return nil, err
@@ -55,6 +62,13 @@ func DelegateAdd(ctx context.Context, delegatePlugin string, netconf []byte, exe
 // DelegateCheck calls the given delegate plugin with the CNI CHECK action and
 // JSON configuration
 func DelegateCheck(ctx context.Context, delegatePlugin string, netconf []byte, exec Exec) error {
+       cmd := os.Getenv("CNI_COMMAND")
+       if cmd == "" {
+               return fmt.Errorf("environment variable CNI_COMMAND must be specified.")
+       }
+       // restore CNI_COMMAND to original value upon return.
+       defer os.Setenv("CNI_COMMAND", cmd)
+       os.Setenv("CNI_COMMAND", "CHECK")
        pluginPath, realExec, err := delegateCommon("CHECK", delegatePlugin, exec)
        if err != nil {
                return err
@@ -66,6 +80,13 @@ func DelegateCheck(ctx context.Context, delegatePlugin string, netconf []byte, e
 // DelegateDel calls the given delegate plugin with the CNI DEL action and
 // JSON configuration
 func DelegateDel(ctx context.Context, delegatePlugin string, netconf []byte, exec Exec) error {
+       cmd := os.Getenv("CNI_COMMAND")
+       if cmd == "" {
+               return fmt.Errorf("environment variable CNI_COMMAND must be specified.")
+       }
+       // restore CNI_COMMAND to original value upon return.
+       defer os.Setenv("CNI_COMMAND", cmd)
+       os.Setenv("CNI_COMMAND", "DEL")
        pluginPath, realExec, err := delegateCommon("DEL", delegatePlugin, exec)
        if err != nil {
                return err
```

dcbw: @mars1024 @mccv1r0 yeah, that's what I would have done here.

mars1024: @dcbw @mccv1r0  I think `DelegateArgs` seems more proper than `set/unset env`because ...
1. It's centralized, we don't have to do code change in **every** delegate funcs
2. It can be easily expanded, maybe in the future, we need to override another env ?
3. It has no influence on main process' env because we don't change it but just override it in heritance, this maybe less risky

But I also think the simpler way which you talk about can work, if you cloud persuade me about the three points above, I'd be happy to change the code ~ thanks ~

mars1024: hi, @dcbw @mccv1r0 , any further suggestions for promoting this? I'm very happy to change the code if needed.

dcbw:
Could you add a testcase to still test the old AsEnv() bits here?

https://coveralls.io/builds/23195194/source?filename=pkg%2Finvoke%2Fargs.go#L34

It looks like the tests moved testing from old to new and now the old bits here aren't tested.

pkg/invoke/delegate.go
```diff
@@ -44,32 +39,42 @@ func delegateCommon(expectedCommand, delegatePlugin string, exec Exec) (string,
...
 		return err
 	}
 
-	return ExecPluginWithoutResult(ctx, pluginPath, netconf, ArgsFromEnv(), realExec)
+	// DelegateDel will override the original CNI_COMMAND env from process with DEL
+	return ExecPluginWithoutResult(ctx, pluginPath, netconf, args("DEL"), realExec)
+}
+
+// return CNIArgs used by delegation
+func args(action string) *DelegateArgs {
```
squeed: Can you rename this to something a bit more descriptive? `args` is used in a lot of places :-)

mars1024: sure, maybe rename `delegateArgs` :)

mars1024:
> Could you add a testcase to still test the old AsEnv() bits here?
> 
> https://coveralls.io/builds/23195194/source?filename=pkg%2Finvoke%2Fargs.go#L34
> 
> It looks like the tests moved testing from old to new and now the old bits here aren't tested.

Sure, I think the coverage has been improved to **96.23** in the latest commit, https://coveralls.io/builds/23423454/source?filename=pkg/invoke/args.go .

pkg/invoke/delegate.go
mars1024: updated

mars1024:
Updated, @squeed @dcbw , please take a look, thanks ~ cc @mccv1r0

mars1024:
Friendly ping @squeed @dcbw , ptal, thanks ~

mars1024:
@squeed @bboreham @dcbw  Could you help to promote this PR ? If there is something I need to do, please let me know, thanks ~

squeed:
@mars1024 sorry for missing this - thanks!

PR #671, opened 2019-05-24
Correct the grammar mistake of depends

kedy-zhao:
Correct the grammar mistake of depends


bboreham:
Thanks for your interest, but it's correct the way it was.  "What unit" is singular so the verb is "depends".
https://ell.stackexchange.com/questions/88620/what-is-the-difference-between-depend-and-depends

PR #675, opened 2019-05-31
skip CNI cache if no CNIVersion is not passed

nirmoy:
If cni config file doesn't contain `cniVersion` then crio fails to remove network from the container and throws :
```
level=error msg="Error deleting network: failed to convert major version part "": strconv.Atoi: parsing "": invalid syntax"
level=error msg="Error while removing pod from CNI network "cilium": failed to convert major version part "": strconv.

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -284,13 +284,14 @@ func (c *CNIConfig) checkNetwork(ctx context.Context, name, cniVersion string, n
 
 // CheckNetworkList executes a sequence of plugins with the CHECK command
 func (c *CNIConfig) CheckNetworkList(ctx context.Context, list *NetworkConfigList, rt *RuntimeConf) error {
-	// CHECK was added in CNI spec version 0.4.0 and higher
-	if gtet, err := version.GreaterThanOrEqualTo(list.CNIVersion, "0.4.0"); err != nil {
-		return err
-	} else if !gtet {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: libcni/api.go [hunk 1]
```diff
@@ -284,13 +284,14 @@ func (c *CNIConfig) checkNetwork(ctx context.Context, name, cniVersion string, n
 
 // CheckNetworkList executes a sequence of plugins with the CHECK command
 func (c *CNIConfig) CheckNetworkList(ctx context.Context, list *NetworkConfigList, rt *RuntimeConf) error {
-	// CHECK was added in CNI spec version 0.4.0 and higher
-	if gtet, err := version.GreaterThanOrEqualTo(list.CNIVersion, "0.4.0"); err != nil {
-		return err
-	} else if !gtet {
-		return fmt.Errorf("configuration version %q does not support the CHECK command", list.CNIVersion)
+	if len(list.CNIVersion) > 0 {
```
mars1024: This means that if `list.CNIVersion` is empty, we will treat it as greater than 0.4.0 ?

nirmoy: No, it means do version checking only if  `list.CNIVersion` is not empty.

nirmoy:
`version.GreaterThanOrEqualTo` is actually checking empty version. I need to check more.

PR #678, opened 2019-06-04
libcni: add config caching [v2]

mccv1r0:
Continuation of https://github.com/containernetworking/cni/pull/661 that fixes up conflicts in the existing PR.


libcni/api.go
```diff
@@ -234,6 +243,99 @@ func (c *CNIConfig) GetNetworkCachedResult(net *NetworkConfig, rt *RuntimeConf)
...
+			return nil, nil, fmt.Errorf("failed to parse cached config CNI Args %q: %v", unmarshaled.CniArgs, err)
+		}
+	}
+	newRt.CapabilityArgs = unmarshaled.CapabilityArgs
+
+	return unmarshaled.Config, newRt, nil
+}
+
+// GetNetworkListCachedConfig returns the cached Config of the previous
+// previous AddNetworkList() operation for a network list, or an error.
```
jellonek: Please remote the second `previous`.

libcni/api.go
```diff
@@ -234,6 +243,99 @@ func (c *CNIConfig) GetNetworkCachedResult(net *NetworkConfig, rt *RuntimeConf)
...
+func (c *CNIConfig) GetNetworkListCachedConfig(list *NetworkConfigList, rt *RuntimeConf) ([]byte, *RuntimeConf, error) {
+	return getCachedConfig(list.Name, rt)
+}
+
+// GetNetworkCachedConfig returns the cached Config of the previous
+// previous AddNetwork() operation for a network, or an error.
```
jellonek: ditto

libcni/api.go
```diff
@@ -165,33 +167,40 @@ func (c *CNIConfig) ensureExec() invoke.Exec {
...
+	cacheFileTypeConfig cacheFileType = "config"
+)
+
+func getCacheFilePath(fileType cacheFileType, netName string, rt *RuntimeConf) string {
 	cacheDir := rt.CacheDir
 	if cacheDir == "" {
 		cacheDir = CacheDir
 	}
-	return filepath.Join(cacheDir, "results", fmt.Sprintf("%s-%s-%s", netName, rt.ContainerID, rt.IfName))
+	return filepath.Join(cacheDir, string(fileType), fmt.Sprintf("%s-%s-%s", netName, rt.ContainerID, rt.IfName))
```
squeed: Need to escape all these variables - there's no guarantee that they are filename-safe.

dcbw:
@mccv1r0 also Casey pointed out in the other review that maybe we should have some versioning in the cache file. See https://github.com/containernetworking/cni/pull/661#discussion_r280482266

mccv1r0:
I agree that variables should be [sanitized on plugins level](https://github.com/containernetworking/cni/issues/679#issuecomment-499803947).  For this version, I escaped the path.  Versioning added as well.

jellonek:
The name `C:\Users\appveyor\AppData\Local\Temp\1\cni_cachedir523654598\config\"apitest-some-container-id-some-eth0"` does not look to be healthy.

Btw. please remove the merge master branch commit from history and use instead `git rebase -i origin/master` on your branch.

libcni/api_test.go
```diff
@@ -122,11 +123,13 @@ func newPluginInfo(configValue, prevResult string, injectDebugFilePath bool, res
 }
 
 func resultCacheFilePath(cacheDirPath, netName string, rt *libcni.RuntimeConf) string {
-	return filepath.Join(cacheDirPath, "results", netName+"-"+rt.ContainerID+"-"+rt.IfName)
+	fName := fmt.Sprintf(netName + "-" + rt.ContainerID + "-" + rt.IfName)
```
jellonek: O_O
What is the reason for sum to be enclosed in `fmt.Sprintf`?

libcni/api_test.go
```diff
@@ -122,11 +123,13 @@ func newPluginInfo(configValue, prevResult string, injectDebugFilePath bool, res
...
+	return filepath.Join(cacheDirPath, "results", fmt.Sprintf("%q", fName))
```
jellonek: `%q` in this place does not do any escaping - it's adding enclosing quotation which leads to construct an output in form like: `C:\Users\appveyor\AppData\Local\Temp\1\cni_cachedir523654598\config\"apitest-some-container-id-some-eth0"`

Maybe usage of https://github.com/kennygrant/sanitize/blob/master/sanitize.go#L169 would be better there, on whole string produced by `filepath.Join`?

dcbw:
RE escaping see https://github.com/containernetworking/cni/issues/679#issuecomment-501324147

test.sh
```diff
@@ -4,11 +4,17 @@ set -euo pipefail
...
 cd "$(dirname $0)"
 
+go get -t ./...
+
 echo -n "Running tests "
 function testrun {
     bash -c "umask 0; PATH=$PATH go test $@"
 }
 if [ ! -z "${COVERALLS:-""}" ]; then
+    go get golang.org/x/tools/cmd/cover
```
dcbw: Oops, none of this stuff was supposed to be in this PR. Can you remove all the changes to this file @mccv1r0 ?

libcni/api.go
```diff
@@ -223,17 +233,123 @@ func getCachedResult(netName, cniVersion string, rt *RuntimeConf) (types.Result,
...
 }
 
 // GetNetworkCachedResult returns the cached Result of the previous
-// previous AddNetwork() operation for a network, or an error.
+// AddNetwork() operation for a network, or an error.
 func (c *CNIConfig) GetNetworkCachedResult(net *NetworkConfig, rt *RuntimeConf) (types.Result, error) {
 	return getCachedResult(net.Network.Name, net.Network.CNIVersion, rt)
 }
 
+func stringifyArgs(args [][2]string) string {
```
dcbw: @mccv1r0 let's drop this into a new pkg/util/utils.go along with the ContainerID escaping functions, then use it for skel.go.

libcni/api.go
```diff
@@ -223,17 +233,123 @@ func getCachedResult(netName, cniVersion string, rt *RuntimeConf) (types.Result,
...
+	entries := make([]string, len(args))
+	for i, kv := range args {
+		entries[i] = strings.Join(kv[:], "=")
+	}
+	return strings.Join(entries, ";")
+}
+
+func parseArgs(args string) ([][2]string, error) {
```
dcbw: @mccv1r0 let's drop this into a new pkg/util/utils.go along with the ContainerID escaping functions.

jellonek:
@mccv1r0 so there is still a question why you are calling sprintf on sum of strings.

mccv1r0:
> @mccv1r0 so there is still a question why you are calling sprintf on sum of strings.

As it stands now, each of the strings, ContainerID, networkName and ifName might need escaping.  

Would you prefer just ContainerID?  ifName won't need it, networkName "probably" won't need it but I don't want to chance it.

libcni/api.go
```diff
@@ -165,33 +169,41 @@ func (c *CNIConfig) ensureExec() invoke.Exec {
...
 
 func delCachedResult(netName string, rt *RuntimeConf) error {
-	fname := getResultCacheFilePath(netName, rt)
+	fname := getCacheFilePath(cacheFileTypeResult, netName, rt)
 	return os.Remove(fname)
 }
 
 func getCachedResult(netName, cniVersion string, rt *RuntimeConf) (types.Result, error) {
-	fname := getResultCacheFilePath(netName, rt)
+	fname := getCacheFilePath(cacheFileTypeResult, netName, rt)
```
dcbw: For backwards compat with older code, could we fall back to the un-base64-ed name here? Otherwise things may not work as well when a project revendors CNI and had already been using the previous code.

libcni/api.go
```diff
@@ -223,17 +235,99 @@ func getCachedResult(netName, cniVersion string, rt *RuntimeConf) (types.Result,
...
 
 // GetNetworkCachedResult returns the cached Result of the previous
-// previous AddNetwork() operation for a network, or an error.
+// AddNetwork() operation for a network, or an error.
 func (c *CNIConfig) GetNetworkCachedResult(net *NetworkConfig, rt *RuntimeConf) (types.Result, error) {
 	return getCachedResult(net.Network.Name, net.Network.CNIVersion, rt)
 }
 
+type cniCachedFile struct {
+	Kind         string       `json:"kind"`
```
dcbw: @squeed might have a comment or two here, not sure exactly what he wanted around versioning.

dcbw: But also, any reason to have these two structures separate? Why not combine the two?

libcni/api_test.go
dcbw: @mccv1r0 Yeah, either fmt.Sprintf("%s-%s-%s", ...) or just netName + "-" .... Best to be consistent I think.

libcni/api.go
mars1024: I think **Kind** here is better to be a enumerated type or a VERSION type for scalability in future.

libcni/api.go
```diff
@@ -165,33 +169,41 @@ func (c *CNIConfig) ensureExec() invoke.Exec {
...
+)
+
+func getCacheFilePath(fileType cacheFileType, netName string, rt *RuntimeConf) string {
 	cacheDir := rt.CacheDir
 	if cacheDir == "" {
 		cacheDir = CacheDir
 	}
-	return filepath.Join(cacheDir, "results", fmt.Sprintf("%s-%s-%s", netName, rt.ContainerID, rt.IfName))
+	fName := fmt.Sprintf("%s-%s-%s", netName, rt.ContainerID, rt.IfName)
+	return filepath.Join(cacheDir, string(fileType), base64.StdEncoding.EncodeToString([]byte(fName)))
```
mars1024: What if **rt.CacheDir** changes ? We will not get the **previous** cached file in a **new** path. Maybe we can have a fixed cache dir here.

libcni/api.go
```diff
@@ -223,17 +235,99 @@ func getCachedResult(netName, cniVersion string, rt *RuntimeConf) (types.Result,
...
+	}
+	cached := unmarshaled.CachedConfig
+
+	newRt := &RuntimeConf{}
+	if cached.CniArgs != "" {
+		if newRt.Args, err = utils.ParseArgs(cached.CniArgs); err != nil {
+			return nil, nil, fmt.Errorf("failed to parse cached config CNI Args %q: %v", cached.CniArgs, err)
+		}
+	}
+	newRt.CapabilityArgs = cached.CapabilityArgs
```
mars1024: I think **newRt** here should be filled with ContainerID, NetNS and so on, IfName, because we may use it to DelNetworkList directly and these fields are necessary.

libcni/api.go
dcbw: I would say that runtimes simply shouldn't do that, but it's an interesting point. I think we put CacheDir into RuntimeConfig because then it's not a required argument for NewCNIConfig() and thus CNI users don't have to change anything unless they want a custom cache dir.

libcni/api.go
mars1024: Yeah, seems a better way ~

libcni/api.go
mars1024: Yes, I also agree on that runtime should not change cache dir freely.
So maybe there is another way, we can put cache dir from runtimeConfig to CNIConfig, and add a new constructor `func NewCNIConfig(path []string, exec invoke.Exec, cacheDir string) *CNIConfig`, if we use the old constructor `NewCNIConfig(path []string, exec invoke.Exec) *CNIConfig`, the cache dir will be a default value. Once cache dir is set in CNIConfig, it won't be changed by any CNI user.

libcni/api.go
mars1024: OK, anyway, it's working for now :)

jellonek:
> > @mccv1r0 so there is still a question why you are calling sprintf on sum of strings.
> 
> As it stands now, each of the strings, ContainerID, networkName and ifName might need escaping.
> 
> Would you prefer just ContainerID? ifName won't need it, networkName "probably" won't need it but I don't want to chance it.

Ok, maybe my question was not fully clear - you are taking the sum of strings, which is a string, then you are passing that to sprintf as a format string, but without any format arguments. So the question is not about what should be inside constructed string (IMO this set of fields is fine), but why fully built string is used as a lone parameter for sprintf. As @dcwb noted - you should use sum, or sprintf.

libcni/api.go
mccv1r0: Ensuring these variables will be finename-safe will be address via https://github.com/containernetworking/cni/issues/679

libcni/api.go
```diff
@@ -193,24 +207,77 @@ func (c *CNIConfig) getResultCacheFilePath(netName string, rt *RuntimeConf) stri
 	return filepath.Join(c.getCacheDir(rt), "results", fmt.Sprintf("%s-%s-%s", netName, rt.ContainerID, rt.IfName))
 }
 
-func (c *CNIConfig) setCachedResult(result types.Result, netName string, rt *RuntimeConf) error {
+func (c *CNIConfig) setCachedResult(result types.Result, config []byte, netName string, rt *RuntimeConf) error {
+
+	cached := cachedInfo{Config: config,
```
dcbw: @mccv1r0 newline here between the { and the Config:.  Also put the closing brace on a line by itself eg:

```
type foo struct {
    Member1: value,
    Member2: value2,
}
```

libcni/api.go
```diff
@@ -175,6 +178,16 @@ func (c *CNIConfig) ensureExec() invoke.Exec {
 	return c.exec
 }
 
+type cachedInfo struct {
+	Kind           string                 `json:"kind"`
+	CNIVersion     string                 `json:"cniVersion"`
+	Config         []byte                 `json:"config"`
+	CniArgs        string                 `json:"cniArgs,omitempty"`
```
dcbw: @mccv1r0 If we're going full JSON here we might as well make CniArgs just be a `[][2]string` just like it is in RuntimeConf and then we don't need stringify/unstringify.

libcni/api.go
```diff
@@ -193,24 +207,77 @@ func (c *CNIConfig) getResultCacheFilePath(netName string, rt *RuntimeConf) stri
...
+	if len(rt.CapabilityArgs) > 0 {
+		cached.CapabilityArgs = rt.CapabilityArgs
+	}
+
 	data, err := json.Marshal(result)
 	if err != nil {
 		return err
 	}
+
 	fname := c.getResultCacheFilePath(netName, rt)
```
dcbw: @mccv1r0 Can we move this block (getResultCacheFilePath and MkdirAll() to just before the WriteFile()? And then also add a comment about the dance with RawResult/Result and the marshalling.

libcni/api.go
```diff
@@ -193,24 +207,77 @@ func (c *CNIConfig) getResultCacheFilePath(netName string, rt *RuntimeConf) stri
...
+
+	unmarshaled := cachedInfo{}
+	if err := json.Unmarshal(bytes, &unmarshaled); err != nil {
+		return nil, nil, fmt.Errorf("failed to unmarshal cached network %q config: %v", netName, err)
+	}
+	if unmarshaled.Kind != "cniCacheV1" {
+		return nil, nil, fmt.Errorf("read cached network %q config has wrong kind: %v", netName, unmarshaled.Kind)
+	}
+
+	newRt := &RuntimeConf{}
```
dcbw: Can we do the suggestion from https://github.com/containernetworking/cni/pull/678#discussion_r295343502 here? eg make `newRt` a copy of the incoming `rt` and then update it.

libcni/api.go
```diff
@@ -242,18 +309,76 @@ func (c *CNIConfig) getCachedResult(netName, cniVersion string, rt *RuntimeConf)
...
 
 // GetNetworkCachedResult returns the cached Result of the previous
-// previous AddNetwork() operation for a network, or an error.
+// AddNetwork() operation for a network, or an error.
 func (c *CNIConfig) GetNetworkCachedResult(net *NetworkConfig, rt *RuntimeConf) (types.Result, error) {
 	return c.getCachedResult(net.Network.Name, net.Network.CNIVersion, rt)
 }
 
+// GetNetworkListCachedConfig returns the cached Config of the previous
+// AddNetworkList() operation for a network list, or an error.
```
dcbw: @mccv1r0 we should update the comment to say that it copies the input RuntimeConf to the output RuntimeConf but updates the fields with the info from the cached config.

libcni/api.go
```diff
@@ -175,6 +177,16 @@ func (c *CNIConfig) ensureExec() invoke.Exec {
 	return c.exec
 }
 
+type cachedInfo struct {
+	Kind           string                 `json:"kind"`
+	CNIVersion     string                 `json:"cniVersion"`
```
dcbw: @mccv1r0 I don't think we use CNIVersion anywhere here now?

libcni/api.go
```diff
@@ -193,24 +206,78 @@ func (c *CNIConfig) getResultCacheFilePath(netName string, rt *RuntimeConf) stri
 	return filepath.Join(c.getCacheDir(rt), "results", fmt.Sprintf("%s-%s-%s", netName, rt.ContainerID, rt.IfName))
 }
 
-func (c *CNIConfig) setCachedResult(result types.Result, netName string, rt *RuntimeConf) error {
+func (c *CNIConfig) setCachedResult(result types.Result, config []byte, netName string, rt *RuntimeConf) error {
+
+	cached := cachedInfo{
+		Config:     config,
+		Kind:       "cniCacheV1",
```
jellonek: We should have that value as a const.

libcni/api.go
```diff
@@ -175,6 +177,16 @@ func (c *CNIConfig) ensureExec() invoke.Exec {
...
 }
 
+type cachedInfo struct {
+	Kind           string                 `json:"kind"`
+	CNIVersion     string                 `json:"cniVersion"`
+	Config         []byte                 `json:"config"`
+	CniArgs        [][2]string            `json:"cniArgs,omitempty"`
+	CapabilityArgs map[string]interface{} `json:"capabilityArgs,omitempty"`
+	RawResults     map[string]interface{} `json:"results,omitempty"`
+	Results        types.Result           `json:"-"`
```
squeed: Nit: this should be called Result everywhere - it's not plural.

libcni/api.go
```diff
@@ -193,24 +206,78 @@ func (c *CNIConfig) getResultCacheFilePath(netName string, rt *RuntimeConf) stri
 	return filepath.Join(c.getCacheDir(rt), "results", fmt.Sprintf("%s-%s-%s", netName, rt.ContainerID, rt.IfName))
 }
 
-func (c *CNIConfig) setCachedResult(result types.Result, netName string, rt *RuntimeConf) error {
+func (c *CNIConfig) setCachedResult(result types.Result, config []byte, netName string, rt *RuntimeConf) error {
+
+	cached := cachedInfo{
+		Config:     config,
+		Kind:       "cniCacheV1",
+		CNIVersion: "0.4.0"}
```
jellonek: I'm not sure about this one, but probably it should also be declared as constant. Is it anywhere used?

libcni/api.go
squeed: I don't think CNIVersion is useful here - we already have Kind.

libcni/api.go
```diff
@@ -193,24 +206,78 @@ func (c *CNIConfig) getResultCacheFilePath(netName string, rt *RuntimeConf) stri
...
+	if err != nil {
+		// Ignore read errors; the cached result may not exist on-disk
+		return nil, nil, nil
+	}
+
+	unmarshaled := cachedInfo{}
+	if err := json.Unmarshal(bytes, &unmarshaled); err != nil {
+		return nil, nil, fmt.Errorf("failed to unmarshal cached network %q config: %v", netName, err)
+	}
+	if unmarshaled.Kind != "cniCacheV1" {
```
jellonek: Mentioned above const should be also used there (while in tests it would be nice to still have a string value).

libcni/api.go
```diff
@@ -242,18 +309,76 @@ func (c *CNIConfig) getCachedResult(netName, cniVersion string, rt *RuntimeConf)
...
+		// Ignore read errors; the cached result may not exist on-disk
+		return nil, nil
+	}
+
+	cachedInfo := cachedInfo{}
+	if err := json.Unmarshal(fdata, &cachedInfo); err != nil {
+		return c.getLegacyCachedResult(netName, cniVersion, rt)
+	}
+
+	if cachedInfo.Kind != "cniCacheV1" {
```
jellonek: ditto.

pkg/utils/utils.go
```diff
@@ -0,0 +1,44 @@
+// Copyright 2019 CNI authors
```
squeed: Looks like this file isn't used anymore.

squeed:
It would be good to see a test case that covered reading an "old" result-only file.

libcni/api.go
mccv1r0: // GetNetworkListCachedConfig copies the input RuntimeConf to output
// RuntimeConf with fields updated with info from the cached Config. 
??

libcni/api.go
```diff
@@ -175,6 +181,15 @@ func (c *CNIConfig) ensureExec() invoke.Exec {
 	return c.exec
 }
 
+type cachedInfo struct {
+	Kind           string                 `json:"kind"`
+	Config         []byte                 `json:"config"`
+	CniArgs        [][2]string            `json:"cniArgs,omitempty"`
+	CapabilityArgs map[string]interface{} `json:"capabilityArgs,omitempty"`
+	RawResults     map[string]interface{} `json:"results,omitempty"`
```
dcbw: @mccv1r0 let's make this `RawResult` to stay consistent. There's really only one Result object here.

libcni/api.go
```diff
@@ -183,6 +198,7 @@ func (c *CNIConfig) getCacheDir(rt *RuntimeConf) string {
 	if c.cacheDir != "" {
 		return c.cacheDir
 	}
+
```
dcbw: Spurious newline.

libcni/api.go
```diff
@@ -193,24 +209,77 @@ func (c *CNIConfig) getResultCacheFilePath(netName string, rt *RuntimeConf) stri
 	return filepath.Join(c.getCacheDir(rt), "results", fmt.Sprintf("%s-%s-%s", netName, rt.ContainerID, rt.IfName))
 }
 
-func (c *CNIConfig) setCachedResult(result types.Result, netName string, rt *RuntimeConf) error {
+func (c *CNIConfig) setCachedResult(result types.Result, config []byte, netName string, rt *RuntimeConf) error {
+
+	cached := cachedInfo{
+		Config: config,
+		Kind:   CNICacheV1}
```
dcbw: @mccv1r0 while I know this passes gofmt, lets do the standard:

```
    Kind: CNICacheV1,
}
```

libcni/api_test.go
```diff
@@ -1513,5 +1584,82 @@ var _ = Describe("Invoking plugins", func() {
 			Expect(foundFirst).To(BeTrue())
 			Expect(foundSecond).To(BeTrue())
 		})
+
+		It("creates separate config cache files for multiple attachments to the same network", func() {
```
dcbw: Shouldn't there be just one test now for both cached config and result? Rather than two here.

dcbw:
> It would be good to see a test case that covered reading an "old" result-only file.

@mccv1r0 I didn't see a testcase for this yet, but may have missed it.

libcni/api.go
```diff
@@ -242,18 +311,76 @@ func (c *CNIConfig) getCachedResult(netName, cniVersion string, rt *RuntimeConf)
...
+func (c *CNIConfig) getCachedResult(netName, cniVersion string, rt *RuntimeConf) (types.Result, error) {
+	fname := c.getResultCacheFilePath(netName, rt)
+	fdata, err := ioutil.ReadFile(fname)
+	if err != nil {
+		// Ignore read errors; the cached result may not exist on-disk
+		return nil, nil
+	}
+
+	cachedInfo := cachedInfo{}
+	if err := json.Unmarshal(fdata, &cachedInfo); err != nil {
```
mars1024: I think judging condition here could be `err != nil || cachedInfo.Kind != CNICacheV1` to spare four lines of the following `if`.

libcni/api.go
```diff
@@ -242,18 +311,76 @@ func (c *CNIConfig) getCachedResult(netName, cniVersion string, rt *RuntimeConf)
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert to the config version to ensure plugins get prevResult
+	// in the same version as the config.  The cached result version
+	// should match the config version unless the config was changed
+	// while the container was running.
+	result, err = result.GetAsVersion(cniVersion)
+	if err != nil && resultCniVersion != cniVersion {
```
mars1024: I don't think the judgment here is right, assume `err != nil` and `resultCniVersion == cniVersion`, we just let it return?

libcni/api.go
```diff
@@ -193,24 +209,77 @@ func (c *CNIConfig) getResultCacheFilePath(netName string, rt *RuntimeConf) stri
 	return filepath.Join(c.getCacheDir(rt), "results", fmt.Sprintf("%s-%s-%s", netName, rt.ContainerID, rt.IfName))
 }
 
-func (c *CNIConfig) setCachedResult(result types.Result, netName string, rt *RuntimeConf) error {
+func (c *CNIConfig) setCachedResult(result types.Result, config []byte, netName string, rt *RuntimeConf) error {
+
+	cached := cachedInfo{
+		Config: config,
+		Kind:   CNICacheV1}
+	if len(rt.Args) > 0 {
```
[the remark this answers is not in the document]
jellonek: This could provide distinction between an empty slice and non defined slice (`[]` vs `nil`). That could be usable if we would annotations for `RuntimeConf` for marshaling, but we don't have them, so also IMO this (and later `CapabilityArgs`) could simply go above struct.

libcni/api.go
```diff
@@ -193,24 +209,77 @@ func (c *CNIConfig) getResultCacheFilePath(netName string, rt *RuntimeConf) stri
...
+	return ioutil.WriteFile(fname, newBytes, 0600)
 }
 
 func (c *CNIConfig) delCachedResult(netName string, rt *RuntimeConf) error {
 	fname := c.getResultCacheFilePath(netName, rt)
 	return os.Remove(fname)
 }
 
-func (c *CNIConfig) getCachedResult(netName, cniVersion string, rt *RuntimeConf) (types.Result, error) {
+func (c CNIConfig) getCachedConfig(netName string, rt *RuntimeConf) ([]byte, *RuntimeConf, error) {
```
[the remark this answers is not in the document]
squeed: Yup, this should take a pointer receiver.

libcni/api.go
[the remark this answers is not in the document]
rosenhouse: Agreed.  I don't want our API to depend on the difference between `[]` and `nil`.

libcni/api.go
mccv1r0: I kept things as they were when results were in a separate file.

libcni/api.go
mccv1r0: What is returned to caller is different for each if (or I don't understand your comment.)

mars1024: I mean that we can compress
```
	cachedInfo := cachedInfo{}
	if err := json.Unmarshal(fdata, &cachedInfo); err != nil {
		return c.getLegacyCachedResult(netName, cniVersion, rt)
	}

	if cachedInfo.Kind != CNICacheV1 {
		return c.getLegacyCachedResult(netName, cniVersion, rt)
	}
```
into 
```
	cachedInfo := cachedInfo{}
	if err := json.Unmarshal(fdata, &cachedInfo); err != nil || cachedInfo.Kind != CNICacheV1{
		return c.getLegacyCachedResult(netName, cniVersion, rt)
	}
```

mccv1r0:
> > It would be good to see a test case that covered reading an "old" result-only file.
> 
> @mccv1r0 I didn't see a testcase for this yet, but may have missed it.

See line 517

```
				Context("no cached config just result", func() {
					It("passes a prevResult to the plugin", func() {
						err := ioutil.WriteFile(cacheFile, []byte(`{
							"cniVersion": "0.3.1",
							"ips": [{"version": "4", "address": "10.1.2.3/24"}],
							"dns": {}
						}`), 0600)

						err = cniConfig.CheckNetwork(ctx, netConfig, runtimeConfig)
						debug, err := noop_debug.ReadDebug(debugFilePath)
						Expect(string(debug.CmdArgs.StdinData)).To(ContainSubstring("\"prevResult\":"))
						Expect(string(debug.CmdArgs.StdinData)).NotTo(ContainSubstring("\"config\":"))
						Expect(string(debug.CmdArgs.StdinData)).NotTo(ContainSubstring("\"kind\":"))
					})
				})
```

Notice the tests to ensure kind and config are not in the file.

libcni/api.go
dcbw: @mars1024 actually in the case where GetAsVersion() fails, it'll return "nil, err". The if here is just special-casing the returned error to provide more information in this specific case.

libcni/api_test.go
```diff
@@ -497,6 +514,25 @@ var _ = Describe("Invoking plugins", func() {
 					})
 				})
 
+				Context("no cached config just result", func() {
```
dcbw: @mccv1r0 can we change the context here to be more inline with Ginkgo style? Eg combine all the It(), Context(), etc messages together. For this test you have:

Context("when CHECK is called with a configuration version"
Context("no cached config just result"
It("passes a prevResult to the plugin

So that would be "when CHECK is called with a configuration version no cached config just result it passes a prevResult to the plugin" and that doesn't parse in English. Compare to the immediately previous:

Context("when CHECK is called with a configuration version"
Context("less than 0.4.0"
It("fails as CHECK is not supported before 0.4.0"

"when CHECK is called with a configuration version less than 0.4.0 it fails as CHECK is not supported before 0.4.0" which does parse.

dcbw:
@mccv1r0 since all our cache stuff is combined now, I suggest:

setCachedResult -> cacheAdd
getResultCacheFilePath -> getCacheFilePath
delCachedResult -> cacheDel

otherwise LGTM

mccv1r0:
changed

dcbw:
@mccv1r0 one last thing: `cachedDel` -> `cacheDel`

Issue #642, opened 2019-03-15
Review and address 'staticcheck' warnings


bboreham:
I ran `staticcheck`:

```
~/src/github.com/containernetworking/cni$ staticcheck ./...
libcni/api_test.go:855:33: should use make([]pluginInfo, 3) instead (S1019)
libcni/api_test.go:1288:15: this value of err is never used (SA4006)
libcni/conf.go:117:26: error strings should not be capitalized (ST1005)
libcni/conf.go:121:26: error strings should not be capitalized (ST1005)
pkg/invoke/args.go:32:7: receiver name should not be an underscore, omit the name if it is unused (ST1006)
pkg/types/args.go:39:20: error strings should not be capitalized (ST1005)
pkg/types/types.go:199:5: error var NotImplementedError should have name of the form ErrFoo (ST1012)
pkg/types/types.go:199:37: error strings should not be capitalized (ST1005)
plugins/test/noop/main.go:64:26: error strings should not end with punctuation or a newline (ST1005)
```

All seem plausible, though renaming `NotImplementedError` would be a breaking change.

Similar issue: #93 - this needed re-done after the repo split.

mars1024:
`staticcheck` again, 
```
libcni/api_test.go:855:33: should use make([]pluginInfo, 3) instead (S1019)
libcni/api_test.go:1288:15: this value of err is never used (SA4006)
libcni/conf.go:117:26: error strings should not be capitalized (ST1005)
libcni/conf.go:121:26: error strings should not be capitalized (ST1005)
pkg/invoke/args.go:35:7: receiver name should not be an underscore, omit the name if it is unused (ST1006)
pkg/types/args.go:39:20: error strings should not be capitalized (ST1005)
pkg/types/types.go:199:5: error var NotImplementedError should have name of the form ErrFoo (ST1012)
pkg/types/types.go:199:37: error strings should not be capitalized (ST1005)
plugins/test/noop/main.go:64:26: error strings should not end with punctuation or a newline (ST1005)

```

PR #681, opened 2019-06-19
clean up : fix staticcheck warnings

mars1024:
ref to [https://github.com/containernetworking/cni/issues/642](https://github.com/containernetworking/cni/issues/642)

try to fix some warnings generated by [staticcheck](https://staticcheck.io/)

### DIFF
--- a/pkg/invoke/args.go
+++ b/pkg/invoke/args.go
@@ -32,7 +32,7 @@ type inherited struct{}
 
 var inheritArgsFromEnv inherited
 
-func (i *inherited) AsEnv() []string {
+func (*inherited) AsEnv() []string {
 	return nil
 }
 
--- a/pkg/invoke/args.go
+++ b/pkg/invoke/args.go
@@ -32,7 +32,7 @@ type inherited struct{}
 
 var inheritArgsFromEnv inherited
 
-func (_ *inherited) AsEnv() []string {
+func (i *inherited) AsEnv() []string {
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/types/types.go
```diff
@@ -195,5 +195,5 @@ func prettyPrint(obj interface{}) error {
 	return err
 }
 
-// NotImplementedError is used to indicate that a method is not implemented for the given platform
-var NotImplementedError = errors.New("Not Implemented")
```
jellonek: That is desired but it has quite huge impact on API.

BTW. I don't see any usage of this variable anywhere in this repo and in plugins repo. Maybe instead of renaming it - we should remove that at all?

dcbw: If we don't use it anywhere in CNI, then we should remove it. I'm not sure we ever used it anywhere either?

mars1024: For now, it seems that this has not ever been used in CNI repos. I'll remove this, thanks for your advice @jellonek  @dcbw

mars1024: Updated!

pkg/invoke/args.go
```diff
@@ -32,7 +32,7 @@ type inherited struct{}
 
 var inheritArgsFromEnv inherited
 
-func (_ *inherited) AsEnv() []string {
+func (i *inherited) AsEnv() []string {
```
bboreham: the recommendation from `staticcheck` is "omit the name if it is unused"

mars1024: Updated, thanks! @bboreham

PR #682, opened 2019-06-20
libcni: add InitCNIConfigWithCacheDir() and deprecate RuntimeConfig.CacheDir

dcbw:
CacheDir is a property of the entire CNIConfig object and there is no
point to having it be different between container invocations. It was
originally put in RuntimeConfig to make re-vendoring easier so that
users wouldn't have to update calls to InitCNIConfig.

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -48,7 +48,7 @@ type RuntimeConf struct {
 	// to the plugin
 	CapabilityArgs map[string]interface{}
 
-	// A cache directory in which to library data.  Defaults to CacheDir
+	// DEPRECATED. Will be removed in a future release.
 	CacheDir string
 }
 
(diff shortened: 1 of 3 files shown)


mars1024:
ref to #683 , CNIConfig seems to be a more reasonable location to hold cache dir, thanks @dcbw

libcni/api.go
```diff
@@ -165,33 +178,33 @@ func (c *CNIConfig) ensureExec() invoke.Exec {
 	return c.exec
 }
 
-func getResultCacheFilePath(netName string, rt *RuntimeConf) string {
+func (c *CNIConfig) getResultCacheFilePath(netName string, rt *RuntimeConf) string {
 	cacheDir := rt.CacheDir
```
mars1024: maybe we should use c.cacheDir instead of rt.CacheDir directly because rt.CacheDir is deprecated

dcbw: Done

Issue #685, opened 2019-07-03
clean up and fix Errors in skel


mars1024:
clean up
- add some well-known error codes
- clean and use **Error** which is defined in types.go and described in SPEC.md for skel
- use a unified method to generate Error

TBD
- log Error to stderr

mars1024:
close via #686

PR #686, opened 2019-07-03
skel: clean up errors in skel and add some well-known error codes

mars1024:
ref to #685

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -758,4 +758,11 @@ Error codes 1-99 must not be used other than as specified here.
 - `1` - Incompatible CNI version
 - `2` - Unsupported field in network configuration. The error message must contain the key and value of the unsupported field.
 - `3` - Container unknown or does not exist. This error implies the runtime does not need to perform any container network cleanup (for example, calling the `DEL` action on the container).
+- `4` - Missing necessary environment variables, like CNI_COMMAND, CNI_CONTAINERID, etc. The error message must contain the names of missing variables.
+- `5` - I/O failed. For example, failed to read network config bytes from stdin.
+- `6` - Failed to decode content. For example, failed to unmarshal network config from bytes or failed to decode version info from string.
+- `7` - Failed to encode objects. For example, failed to marshal object to bytes.
+- `8` - Invalid network config. If some validations on network configs do not pass, this error will be raised.
+- `9` - Failed to call plugin's functions. If plugin's functions (cmdAdd, cmdCheck, cmdDel) return some non-typed error, this code should be used.
+- `10` - Unknown command. If CNI_COMMAND environment variable is not ADD, DEL, CHECK or VERSION, it is an unknown command.
 - `11` - Try again later. If the plugin detects some transient condition that should clear up, it can use this code to notify the runtime it should re-try the operation later.
--- a/SPEC.md
+++ b/SPEC.md
@@ -758,11 +758,8 @@ Error codes 1-99 must not be used other than as specified here.
 - `1` - Incompatible CNI version
 - `2` - Unsupported field in network configuration. The error message must contain the key and value of the unsupported field.
 - `3` - Container unknown or does not exist. This error implies the runtime does not need to perform any container network cleanup (for example, calling the `DEL` action on the container).
-- `4` - Missing necessary environment variables, like CNI_COMMAND, CNI_CONTAINERID, etc. The error message must contain the names of missing variables.
-- `5` - I/O failed. For example, failed to read network config bytes from stdin.
+- `4` - Invalid necessary environment variables, like CNI_COMMAND, CNI_CONTAINERID, etc. The error message must contain the names of invalid variables.
+- `5` - I/O failure. For example, failed to read network config bytes from stdin.
 - `6` - Failed to decode content. For example, failed to unmarshal network config from bytes or failed to decode version info from string.
-- `7` - Failed to encode objects. For example, failed to marshal object to bytes.
-- `8` - Invalid network config. If some validations on network configs do not pass, this error will be raised.
-- `9` - Failed to call plugin's functions. If plugin's functions (cmdAdd, cmdCheck, cmdDel) return some non-typed error, this code should be used.
-- `10` - Unknown command. If CNI_COMMAND environment variable is not ADD, DEL, CHECK or VERSION, it is an unknown command.
+- `7` - Invalid network config. If some validations on network configs do not pass, this error will be raised.
 - `11` - Try again later. If the plugin detects some transient condition that should clear up, it can use this code to notify the runtime it should re-try the operation later.
--- a/pkg/skel/skel.go
+++ b/pkg/skel/skel.go
@@ -138,7 +138,7 @@ func (t *dispatcher) getCmdArgsFromEnv() (string, *CmdArgs, *types.Error) {
 
 	if len(argsMissing) > 0 {
 		joined := strings.Join(argsMissing, ",")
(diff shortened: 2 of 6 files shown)


dcbw:
@mars1024 Note that the Spec says `Errors must be indicated by a non-zero return code and the following JSON being printed to stdout:` so we have to print errors to stdout by design. But they can also be (and probably should?) mirrored to stderr.

squeed:
Yeah, it would be useful to log the error message to stderr, but I personally don't like logging JSON to stderr.

jellonek:
:+1: for not logging JSON. Also if it's log - it would be nice to have a timestamp at least in it.

mars1024:
I've updated the aims in issue #685 , and i'm going to update this PR soon ~

mars1024:
Updated! PTAL, thanks a lot ~ cc @dcbw @squeed @jellonek

SPEC.md
```diff
@@ -758,4 +758,11 @@ Error codes 1-99 must not be used other than as specified here.
 - `1` - Incompatible CNI version
 - `2` - Unsupported field in network configuration. The error message must contain the key and value of the unsupported field.
 - `3` - Container unknown or does not exist. This error implies the runtime does not need to perform any container network cleanup (for example, calling the `DEL` action on the container).
+- `4` - Missing necessary environment variables, like CNI_COMMAND, CNI_CONTAINERID, etc. The error message must contain the names of missing variables.
+- `5` - I/O failed. For example, failed to read network config bytes from stdin.
+- `6` - Failed to decode content. For example, failed to unmarshal network config from bytes or failed to decode version info from string.
+- `7` - Failed to encode objects. For example, failed to marshal object to bytes.
+- `8` - Invalid network config. If some validations on network configs do not pass, this error will be raised.
+- `9` - Failed to call plugin's functions. If plugin's functions (cmdAdd, cmdCheck, cmdDel) return some non-typed error, this code should be used.
+- `10` - Unknown command. If CNI_COMMAND environment variable is not ADD, DEL, CHECK or VERSION, it is an unknown command.
```
squeed: nit: can you make this one number 5 (and bump everything else)? That way the two environment variable errors are together.

SPEC.md
```diff
@@ -758,4 +758,11 @@ Error codes 1-99 must not be used other than as specified here.
 - `1` - Incompatible CNI version
 - `2` - Unsupported field in network configuration. The error message must contain the key and value of the unsupported field.
 - `3` - Container unknown or does not exist. This error implies the runtime does not need to perform any container network cleanup (for example, calling the `DEL` action on the container).
+- `4` - Missing necessary environment variables, like CNI_COMMAND, CNI_CONTAINERID, etc. The error message must contain the names of missing variables.
+- `5` - I/O failed. For example, failed to read network config bytes from stdin.
+- `6` - Failed to decode content. For example, failed to unmarshal network config from bytes or failed to decode version info from string.
+- `7` - Failed to encode objects. For example, failed to marshal object to bytes.
```
squeed: This is an internal error (go + skel) specific - so we should just return 999 and not add encoding failures to the SPEC.

SPEC.md
```diff
@@ -758,4 +758,11 @@ Error codes 1-99 must not be used other than as specified here.
...
+- `8` - Invalid network config. If some validations on network configs do not pass, this error will be raised.
+- `9` - Failed to call plugin's functions. If plugin's functions (cmdAdd, cmdCheck, cmdDel) return some non-typed error, this code should be used.
```
squeed: Likewise, this is an implementation detail of skel - it should be a different error (maybe 998) and not be in SPEC.

SPEC.md
squeed: Correction: chatted with the rest of the maintainers, they'd like to combine this with 4 - missing & invalid are basically the same.

pkg/skel/skel.go
```diff
@@ -222,45 +226,37 @@ func (t *dispatcher) pluginMain(cmdAdd, cmdCheck, cmdDel func(_ *CmdArgs) error,
...
-			Code: types.ErrIncompatibleCNIVersion,
-			Msg:  "plugin version does not allow CHECK",
-		}
+		return types.NewError(types.ErrIncompatibleCNIVersion, "plugin version does not allow CHECK", "")
 	case "DEL":
 		err = t.checkVersionAndCall(cmdArgs, versionInfo, cmdDel)
 	case "VERSION":
-		err = versionInfo.Encode(t.Stdout)
+		if err := versionInfo.Encode(t.Stdout); err != nil {
+			return types.NewError(types.ErrFailedEncode, err.Error(), "")
```
dcbw: @mars1024 this should actually be an IO failure. It's not going to be a JSON marshal failure unless CNI internally does something really stupid which we shouldn't be doing.

squeed:
This is awesome! Thanks for such a great PR.

To make things easier for plugin authors, we'd like there to be just a bit fewer error codes. Because this is changing the spec, we need to live with it "forever." So don't feel disappointed if we have a lot of suggestions.

mars1024:
> This is awesome! Thanks for such a great PR.
> 
> To make things easier for plugin authors, we'd like there to be just a bit fewer error codes. Because this is changing the spec, we need to live with it "forever." So don't feel disappointed if we have a lot of suggestions.

Thanks for your appreciation, I'll update this soon ~

SPEC.md
mars1024: done.

SPEC.md
mars1024: Done, added ErrInternal(999).

SPEC.md
mars1024: Done, added ErrInternal(=999).

pkg/skel/skel.go
mars1024: @dcbw , thanks, I have changed it to ErrIOFailure.

jellonek:
This looks really good!

mars1024:
Updated! PTAL, thanks ~

SPEC.md
```diff
@@ -758,4 +758,8 @@ Error codes 1-99 must not be used other than as specified here.
 - `1` - Incompatible CNI version
 - `2` - Unsupported field in network configuration. The error message must contain the key and value of the unsupported field.
 - `3` - Container unknown or does not exist. This error implies the runtime does not need to perform any container network cleanup (for example, calling the `DEL` action on the container).
-- `11` - Try again later. If the plugin detects some transient condition that should clear up, it can use this code to notify the runtime it should re-try the operation later.
+- `4` - Invalid necessary environment variables, like CNI_COMMAND, CNI_CONTAINERID, etc. The error message must contain the names of invalid variables.
+- `5` - I/O failure. For example, failed to read network config bytes from stdin.
+- `6` - Failed to decode content. For example, failed to unmarshal network config from bytes or failed to decode version info from string.
+- `7` - Invalid network config. If some validations on network configs do not pass, this error will be raised.
+- `8` - Try again later. If the plugin detects some transient condition that should clear up, it can use this code to notify the runtime it should re-try the operation later.
```
bboreham: Can you put this back to 11 please.

jellonek:
Just a note if someone would not understand why we want there 11: http://man7.org/linux/man-pages/man3/errno.3.html and https://www-numi.fnal.gov/offline_software/srt_public_context/WebDocs/Errors/unix_system_errors.html
```c
#define EAGAIN          11      /* Try again */
```
;)

SPEC.md
mars1024: Done. Thanks @bboreham .

mars1024:
Thanks for your suggestion, I've revert index of ErrTryAgainLater to 11, @bboreham @dcbw @jellonek .

PR #692, opened 2019-07-24
libcni: find plugin in exec

mars1024:
Function `validatePlugin` in libcni should find plugin by `exec.FindInPath` instead of using `invoke.FindInPath` directly, just like `addNetwork` [https://github.com/containernetworking/cni/blob/df124b22e4f7a0b716d313884aee6474b319b4a3/libcni/api.go#L378](https://github.com/containernetworking/cni/blob/df124b22e4f7a0b716d313884aee6474b319b4a3/libcni/api.go#L378).

I think this breaks CI of [https://github.com/cri-o/ocicni/pull/51](https://github.com/cri-o/ocicni/pull/51).

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -594,7 +594,8 @@ func (c *CNIConfig) ValidateNetwork(ctx context.Context, net *NetworkConfig) ([]
 
 // validatePlugin checks that an individual plugin's configuration is sane
 func (c *CNIConfig) validatePlugin(ctx context.Context, pluginName, expectedVersion string) error {
-	pluginPath, err := invoke.FindInPath(pluginName, c.Path)
+	c.ensureExec()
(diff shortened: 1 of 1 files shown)


libcni/api.go
```diff
@@ -580,7 +580,7 @@ func (c *CNIConfig) ValidateNetworkList(ctx context.Context, list *NetworkConfig
 // It uses the same logic as ValidateNetworkList)
 // Returns a list of capabilities
 func (c *CNIConfig) ValidateNetwork(ctx context.Context, net *NetworkConfig) ([]string, error) {
-	caps := []string{}
+	var caps []string
```
dcbw: @mars1024 seems like an unrelated change?

mars1024: yeap, maybe some auto typo fix ~ I will revert it.

mars1024: Updated, thanks @dcbw

PR #693, opened 2019-08-02
Update README.md

f1-outsourcing:
I have been working on a netfilter plugin assuming that mesos would support the plugin chaining standard of 2017. Turns out that they do not! Better make sure others are notified about this, because I do not think they will change their documentation on this, as I have requested of them.


bboreham:
I don't think it's practical for us to track compliance of each version of each downstream project.
Maybe add a generic warning to check compliance of the specific thing you need?

f1-outsourcing:
Hmm, I thought it is good to push mesosphere a bit, to start maintaining this standard, they are publicly advertising that mesos 1.8 supports 0.3.0 which is thus incorrect. Wonder even what this apache foundation thinks of projects that are advertising compatibility and are technically not.

bboreham:
Feel free to re-open if you come back to this.

PR #695, opened 2019-08-02
[WIP] set process group ID when fork/exec so child processes terminate together

### LINKED ISSUE #687: libcni: terminate plugins on context timeout
As discussed in our maintainers meeting, we believe that libcni should handle plugin cleanup when the timeout passed by the runtime triggers. libcni should send plugins a TERM signal, wait a reasonable amount of time (a couple seconds), and then KILL plugins.

One complication is that the plugins must be part of the same process group so that (on linux) we can signal the process group and deliver the same signal to all plugins spawned from the main plugin. This could be tricky as Go doesn't easily allow changes between fork & exect.

Some references that might help:
https://stackoverflow.com/questions/34730941/ensure-executables-called-in-go-process-get-killed-when-process-is-killed

We could also use syscall.SysProcAttr on Linux like so:
```
cmd := exec.Command("./long-process")

cmd.SysProcAttr = &syscall.SysProcAttr{
    Setpgid: true,
    Pgid: <some unique number we come up with?>,
}

cmd.Run()
```

And then we can signal the `Pgid`. Code for Windows is in the above stackoverflow link to do essentially the same thing.

mars1024:
Provide a experimental way to fix #687 

We let sub process called by invoke.Exec has the process group id same as its parent process id, so when the parent process is killed, all the sub processes will be killed in tree recursively.

Fox example, if we use libcni in CRI to call plugin **A**, and plugin **A** call plugin **B** and **C**, and plugin **B** call another plugin **D**, then the pid/pgid maybe like,

process|pid|pgid
-|-|-
|CRI|100|120|
|A|103|100|
|B|105|103|
|C|106|103|
|D|108|105|

If CRI (pid==100) was killed, processes(A pid==103) with group id == 100 will be killed, and processes(B pid==105 C pid==106) with group id == 103 will be killed, and process D will be killed in the same way recursively.
For the same reason , if CRI try to kill process A, B&C&D will be killed recursively.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: go
+sudo: required
 dist: bionic
 
 go:
--- a/pkg/invoke/os_unix.go
+++ b/pkg/invoke/os_unix.go
@@ -16,5 +16,16 @@
 
 package invoke
 
+import (
+	"os"
+	"syscall"
+)
+
 // Valid file extensions for plugin executables.
 var ExecutableFileExtensions = []string{""}
+
+// SysProcAttribute holds optional, operating system-specific attributes.
+var SysProcAttribute = &syscall.SysProcAttr{
+	Setpgid: true,
+	Pgid:    os.Getpid(),
+}
--- a/pkg/invoke/os_windows.go
+++ b/pkg/invoke/os_windows.go
@@ -14,5 +14,12 @@
 
 package invoke
 
+import (
+	"syscall"
+)
+
 // Valid file extensions for plugin executables.
 var ExecutableFileExtensions = []string{".exe", ""}
+
+// SysProcAttribute holds optional, operating system-specific attributes.
+var SysProcAttribute = &syscall.SysProcAttr{}
--- a/pkg/invoke/raw_exec.go
+++ b/pkg/invoke/raw_exec.go
@@ -36,6 +36,10 @@ func (e *RawExec) ExecPlugin(ctx context.Context, pluginPath string, stdinData [
 	c.Stdin = bytes.NewBuffer(stdinData)
 	c.Stdout = stdout
 	c.Stderr = e.Stderr
+	// This attributes make sub process has group process id which is same as its parent process id,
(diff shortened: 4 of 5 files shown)


mars1024:
I think test has been broken by #535 , maybe we need **sudo** now, do you have some suggestions? @dcbw

dcbw:
@mars1024 can you add the sudo back to travis.yaml and see what happens?

rosenhouse:
Could you cover this change with a test?

bboreham:
Code looks fine; could I ask that you squash the commits and make the title more like "Set the process group so child processes terminate together"?  The current title is focused on the side-effect rather than the actual change.

mars1024:
> @mars1024 can you add the sudo back to travis.yaml and see what happens?

OK, I'll try ~

mars1024:
> Could you cover this change with a test?

Sure, after CI pass, I'll update some tests.

mars1024:
> Code looks fine; could I ask that you squash the commits and make the title more like "Set the process group so child processes terminate together"? The current title is focused on the side-effect rather than the actual change.

Squash done, thanks ~

squeed:
Hmm, something is wrong - I can't run the tests locally either.

mars1024:
> Hmm, something is wrong - I can't run the tests locally either.

Yes, me too, I'm trying to find more info about the broken CI.

squeed:
> > Hmm, something is wrong - I can't run the tests locally either.
> 
> Yes, me too, I'm trying to find more info about the broken CI.

I checked out your PR and ran `./test_linux.sh` and it failed - should be easier to debug.

mars1024:
I tried, but It seems not work if we set pgid to os.GetPid(), otherwise killing process is different from killing process group, we can not expect os/exec in golang to help us kill process group although process group id is same as process id.

So I will change to another implementation based on killing process group later.

bboreham:
Hi, are you likely to come back to this?

mars1024:
> Hi, are you likely to come back to this?

I'll try but need some time to catch up..

Issue #679, opened 2019-06-06
spec: clarify allowed characters in ContainerID


dcbw:
The spec doesn't say anything about the characters allowed in a ContainerID. We probably shouldn't allow filesystem path characters like "/" or ":", or a ContainerID of just "." or ".." since the ID is used by some plugins in cache file names.

jellonek:
Shouldn't that then be sanitized on plugins level, before they would use that on filesystem? That reason for trimming available character set in this case seems to be invalid to me.

Still, if you would proceed with that, I would suggest to use https://golang.org/pkg/os/#PathSeparator instead of hardcoding "/" there.

dcbw:
@bboreham suggests that we should limit the length of ContainerID too.

dcbw:
We have decided to add filesystem-safe ContainerID escaping functions to containernetworking/plugins/pkg/utils/utils.go because any sane plugin should do that regardless of restrictions on ContainerID content, plus it doesn't seem like a huge win to get all pedantic on ContainerID content.

jellonek:
Having a common/reusable way to sanitize the thing is a very good idea.

dcbw:
And this time around, we've decided to restrict allowed characters in ContainerID and enforce that restriction in libcni and skel.

mccv1r0:
IIRC also for networkName

dcbw:
> IIRC also for networkName

@mccv1r0 agreed.

PR #697, opened 2019-08-12
validate containerID and networkName

mccv1r0:
Validate containerID and networkName
Ensure they contain only letters and numbers

fixes [679](https://github.com/containernetworking/cni/issues/679)

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -1,0 +1,51 @@
+// Copyright 2019 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/containernetworking/cni/pkg/types"
+)
+
+// ValidInputString is the regexp used to validate valid characters in
+// containerID and networkName
+const ValidInputString = "^[a-zA-Z0-9_-]+$"
+
+// ValidateContainerID will validate that the supplied containerID does not contain invalid characters
+func ValidateContainerID(containerID string) *types.Error {
+
+	reg := regexp.MustCompile(ValidInputString)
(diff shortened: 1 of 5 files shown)


pkg/utils/utils.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/containernetworking/cni/pkg/types"
+)
+
+// ValidInputString is the regexp used to validate valid characters in
+// containerID and networkName
+const ValidInputString = "^[a-zA-Z0-9_-]+$"
```
jellonek: Looking on https://stackoverflow.com/a/42642992 it should be probably `[a-zA-Z0-9][a-zA-Z0-9_.-]*` as e.g. `-` can not be used as first character in name.
Was there dot intentionally excluded from this regex?

pkg/utils/utils.go
```diff
@@ -0,0 +1,51 @@
...
+
+// ValidateContainerID will validate that the supplied containerID does not contain invalid characters
```
jellonek: ... that the supplied containerID is not empty and does not contain invalid characters

pkg/utils/utils.go
```diff
@@ -0,0 +1,51 @@
...
+		return types.NewError(types.ErrDecodingFailure, fmt.Sprintf("error: invalid characters in containerID: %v", containerID), "")
+	}
+	return nil
+}
+
+// ValidateNetworkName will validate that the supplied networkName does not contain invalid characters
+func ValidateNetworkName(networkName string) *types.Error {
+
+	reg := regexp.MustCompile(ValidInputString)
+	if networkName != "" {
```
jellonek: This outer if is unnecessary. Is it only an optimization?

mccv1r0: Names seem to be optional for e.g. meta plugins in a list of chained plugins.  Our test cases do this for example (which is how I found the problem.)    I've updated to validate the list name itself, which is all we care about.

pkg/utils/utils.go
mccv1r0: Updated:

```
const CNIvalidNameChars = `[a-zA-Z0-9][a-zA-Z0-9_.-]`
var CNIreg = regexp.MustCompile(`^` + CNIvalidNameChars + `+$`)
```

I checked, this is consistent with what e.g. docker uses for container names.

mccv1r0:
> Can you extract `regexp.MustCompile` to `init()` func to call that only once during the initialization, setting a package scoped variable?

mccv1r0:
I didn't mean to close this <g>   Meant to just address the last comment.

mccv1r0:
Move to: https://github.com/containernetworking/cni/pull/698

PR #698, opened 2019-08-13
validate containerID and networkName

mccv1r0:
validate containerID and networkName
ensure they contain only letters and numbers

Fixes https://github.com/containernetworking/cni/issues/679

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/containernetworking/cni/pkg/invoke"
 	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/utils"
 	"github.com/containernetworking/cni/pkg/version"
 )
 
@@ -377,6 +378,12 @@ func (c *CNIConfig) addNetwork(ctx context.Context, name, cniVersion string, net
 	c.ensureExec()
 	pluginPath, err := c.exec.FindInPath(net.Network.Type, c.Path)
 	if err != nil {
+		return nil, err
+	}
+	if err := utils.ValidateContainerID(rt.ContainerID); err != nil {
+		return nil, err
+	}
+	if err := utils.ValidateNetworkName(name); err != nil {
 		return nil, err
 	}
 
--- a/pkg/skel/skel.go
+++ b/pkg/skel/skel.go
@@ -27,6 +27,7 @@ import (
 	"strings"
 
 	"github.com/containernetworking/cni/pkg/types"
+	"github.com/containernetworking/cni/pkg/utils"
 	"github.com/containernetworking/cni/pkg/version"
 )
 
@@ -183,6 +184,9 @@ func validateConfig(jsonBytes []byte) *types.Error {
 	if conf.Name == "" {
 		return types.NewError(types.ErrInvalidNetworkConfig, "missing network name", "")
 	}
+	if err := utils.ValidateNetworkName(conf.Name); err != nil {
+		return err
+	}
 	return nil
 }
 
@@ -199,6 +203,10 @@ func (t *dispatcher) pluginMain(cmdAdd, cmdCheck, cmdDel func(_ *CmdArgs) error,
 
 	if cmd != "VERSION" {
 		err = validateConfig(cmdArgs.StdinData)
+		if err != nil {
+			return err
+		}
+		err = utils.ValidateContainerID(cmdArgs.ContainerID)
 		if err != nil {
 			return err
 		}
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -1,0 +1,51 @@
+// Copyright 2019 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package utils
+
+import (
+	"regexp"
+
+	"github.com/containernetworking/cni/pkg/types"
+)
+
+// cniValidNameChars is the regexp used to validate valid characters in
+// containerID and networkName
+const cniValidNameChars = `[a-zA-Z0-9][a-zA-Z0-9_.\-]`
+
+var cniReg = regexp.MustCompile(`^` + cniValidNameChars + `*$`)
+
+// ValidateContainerID will validate that the supplied containerID is not empty does not contain invalid characters
+func ValidateContainerID(containerID string) *types.Error {
+
+	if containerID == "" {
+		return types.NewError(types.ErrUnknownContainer, "missing containerID", "")
+	}
+	if !cniReg.MatchString(containerID) {
+		return types.NewError(types.ErrInvalidEnvironmentVariables, "invalid characters in containerID", containerID)
+	}
+	return nil
+}
+
+// ValidateNetworkName will validate that the supplied networkName does not contain invalid characters
+func ValidateNetworkName(networkName string) *types.Error {
+
+	if networkName == "" {
+		return types.NewError(types.ErrInvalidNetworkConfig, "missing network name:", "")
+	}
+	if !cniReg.MatchString(networkName) {
+		return types.NewError(types.ErrInvalidNetworkConfig, "invalid characters found in network name", networkName)
+	}
+	return nil
+}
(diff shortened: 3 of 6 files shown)


mccv1r0:
Ugh

```
Bad response status from coveralls: 405
<html>
<head><title>405 Not Allowed</title></head>
<body bgcolor="white">
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx</center>
</body>
</html>
The command "if [ "${TARGET}" == "amd64" ]; then
  GOARCH="${TARGET}" ./test.sh;
else
  GOARCH="${TARGET}" go list ./... | xargs -n1 go build -v -o /dev/null
fi
" exited with 1.

```

pkg/utils/utils.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// ValidateNetworkName will validate that the supplied networkName does not contain invalid characters
+func ValidateNetworkName(networkName string) *types.Error {
+
+	if networkName == "" {
+		return types.NewError(types.ErrInvalidNetworkConfig, fmt.Sprintf("error: missing network name: %v", networkName), "")
+	}
+	if !CNIreg.MatchString(networkName) {
+		return types.NewError(types.ErrInvalidNetworkConfig, fmt.Sprintf("error: invalid characters found in network name: %v", networkName), "")
```
bboreham: I think this `Sprintf()` could be eliminated by putting the networkName in the third parameter to `NewError()`.

squeed:
reran the build, it passed.

pkg/utils/utils.go
```diff
@@ -0,0 +1,52 @@
...
+		return types.NewError(types.ErrDecodingFailure, fmt.Sprintf("error: invalid characters in containerID: %v", containerID), "")
+	}
+	return nil
+}
+
+// ValidateNetworkName will validate that the supplied networkName does not contain invalid characters
+func ValidateNetworkName(networkName string) *types.Error {
+
+	if networkName == "" {
+		return types.NewError(types.ErrInvalidNetworkConfig, fmt.Sprintf("error: missing network name: %v", networkName), "")
```
bboreham: `Sprintf()` is overkill here - `networkName` is known to be `""`.

pkg/utils/utils.go
```diff
@@ -0,0 +1,52 @@
...
+import (
+	"fmt"
+	"regexp"
+
+	"github.com/containernetworking/cni/pkg/types"
+)
+
+// ValidInputString is the regexp used to validate valid characters in
+// containerID and networkName
+const CNIvalidNameChars = `[a-zA-Z0-9][a-zA-Z0-9_.-]`
```
matthewdupre: Please escape (\_) the trailing hyphen.  While _ and . are not special characters in character classes, hyphens are - although I'm unsure whether that applies at the end.

pkg/utils/utils.go
```diff
@@ -0,0 +1,52 @@
...
+var CNIreg = regexp.MustCompile(`^` + CNIvalidNameChars + `+$`)
+
+// ValidateContainerID will validate that the supplied containerID is not empty does not contain invalid characters
+func ValidateContainerID(containerID string) *types.Error {
+
+	if containerID == "" {
+		return types.NewError(types.ErrUnknownContainer, "missing containerID", "")
+	}
+	if !CNIreg.MatchString(containerID) {
+		return types.NewError(types.ErrDecodingFailure, fmt.Sprintf("error: invalid characters in containerID: %v", containerID), "")
```
bboreham: I think this `Sprintf()` could be eliminated by putting containerID in the third parameter to `NewError()`.

pkg/skel/skel.go
```diff
@@ -199,6 +200,9 @@ func validateConfig(jsonBytes []byte) *types.Error {
 	if conf.Name == "" {
 		return types.NewError(types.ErrInvalidNetworkConfig, "missing network name", "")
 	}
+	if err := utils.ValidateNetworkName(conf.Name); err != nil {
+		return types.NewError(types.ErrInvalidNetworkConfig, "invalid characters found in network name", "")
```
bboreham: Why create a new error here when `ValidateNetworkName()` returns a reasonable one?

libcni/api.go
```diff
@@ -379,6 +380,12 @@ func (c *CNIConfig) addNetwork(ctx context.Context, name, cniVersion string, net
 	if err != nil {
 		return nil, err
 	}
+	if err := utils.ValidateContainerID(rt.ContainerID); err != nil {
+		return nil, fmt.Errorf("error: invalid characters in containerID: %v", rt.ContainerID)
```
bboreham: Why not use the error created by `ValidateContainerID()` ?

libcni/api.go
```diff
@@ -379,6 +380,12 @@ func (c *CNIConfig) addNetwork(ctx context.Context, name, cniVersion string, net
...
+	}
+	if err := utils.ValidateNetworkName(name); err != nil {
+		return nil, fmt.Errorf("error: invalid characters found in network name: %v", name)
```
bboreham: Why not use the error created by `ValidateNetworkName()` ?

pkg/utils/utils.go
mccv1r0: Is the escaping needed inside ``  ?

dcbw:
@mccv1r0 can you also update SPEC.md to describe the new restrictions on network name and containerID?

pkg/utils/utils.go
matthewdupre: It's not clear: the hyphen certainly sometimes has meaning inside the character class (because it's used for the a-z ranges).

pkg/utils/utils.go
```diff
@@ -0,0 +1,51 @@
...
+
+package utils
+
+import (
+	"regexp"
+
+	"github.com/containernetworking/cni/pkg/types"
+)
+
+// ValidInputString is the regexp used to validate valid characters in
```
jellonek: The name used in this docstring was changed, so please update it.

pkg/utils/utils.go
```diff
@@ -0,0 +1,51 @@
...
+// containerID and networkName
+const cniValidNameChars = `[a-zA-Z0-9][a-zA-Z0-9_.-]`
+
+var cniReg = regexp.MustCompile(`^` + cniValidNameChars + `+$`)
```
jellonek: This and above lines are introducing expectation/limitation that name should be constructed from two chars at least. Shouldn't you use there `*` in place of `+`? If `+` is correct - you should update also the spec with note about this limitation.

mars1024: Yes, "a" seems a good name for cni network name, but it can't pass regexp validation.

pkg/utils/utils.go
```diff
@@ -0,0 +1,51 @@
...
+
+// ValidateContainerID will validate that the supplied containerID is not empty does not contain invalid characters
+func ValidateContainerID(containerID string) *types.Error {
+
+	if containerID == "" {
+		return types.NewError(types.ErrUnknownContainer, "missing containerID", "")
+	}
+	if !cniReg.MatchString(containerID) {
+		return types.NewError(types.ErrDecodingFailure, "error: invalid characters in containerID", containerID)
```
mars1024: I think the Error code here should be **ErrInvalidEnvironmentVariables** but not **ErrDecodingFailure** .

mars1024: Any a small questions, why we use `error: ` as prefix of an error output?

pkg/skel/skel.go
```diff
@@ -199,6 +200,9 @@ func validateConfig(jsonBytes []byte) *types.Error {
 	if conf.Name == "" {
 		return types.NewError(types.ErrInvalidNetworkConfig, "missing network name", "")
 	}
+	if err := utils.ValidateNetworkName(conf.Name); err != nil {
```
mars1024: If we validate networkName/containerID in **addNetwork**, then it's not necessary to validate in **skel**, because  addNetwork will call skel, this is a serial process. Validation should be done only once.

pkg/utils/utils.go
mccv1r0: Ack.

pkg/utils/utils.go
mccv1r0: Good catch @jellonek.

pkg/utils/utils.go
mccv1r0: Removed.  The string pre-dated your error code changes, survived the merge

pkg/skel/skel.go
mars1024: resolved? @mccv1r0

dcbw: @mars1024 I think it's necessary to validate in both libcni and skel, because libcni is the runtime side and we want to keep errors closer to the caller and return them before we every get to execing. It's necessary in skel.go because we can't control whether the runtime itself uses libcni or not, and in fact there are alternate implementations of the CNI spec like https://github.com/containerd/go-cni

PR #700, opened 2019-08-19
fix wrong spells in cnitool.md

928234269:


bboreham:
I don't think this was a typo - either 'to' or 'of' works here.

matthewdupre:
I think I prefer to.

PR #704, opened 2019-08-22
libcni: also cache IfName

dcbw:
We should have all components of the uniqueness tuple in the cache file
so that they can be read out of the file without trying to parse the
cache file name.

@bboreham @jellonek @squeed @containernetworking/cni-maintainers

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -207,8 +207,11 @@ func (c *CNIConfig) getCacheDir(rt *RuntimeConf) string {
 	return CacheDir
 }
 
-func (c *CNIConfig) getCacheFilePath(netName string, rt *RuntimeConf) string {
-	return filepath.Join(c.getCacheDir(rt), "results", fmt.Sprintf("%s-%s-%s", netName, rt.ContainerID, rt.IfName))
+func (c *CNIConfig) getCacheFilePath(netName string, rt *RuntimeConf) (string, error) {
+	if netName == "" || rt.ContainerID == "" || rt.IfName == "" {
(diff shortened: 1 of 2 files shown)


squeed:
Hmm, then we should probably put the ContainerID as a top-level variable, too.

dcbw:
@squeed @bboreham @containernetworking/cni-maintainers updated to also cache network name and container ID

libcni/api.go
```diff
@@ -271,6 +276,9 @@ func (c *CNIConfig) getCachedConfig(netName string, rt *RuntimeConf) ([]byte, *R
 		newRt.Args = unmarshaled.CniArgs
 	}
 	newRt.CapabilityArgs = unmarshaled.CapabilityArgs
+	if unmarshaled.IfName != "" {
+		newRt.IfName = unmarshaled.IfName
+	}
```
mars1024: How about adding some checks for ContainerID and NetworkName here?

dcbw: @mars1024 I should actually remove this line because if the interface name isn't already in newRt (because it was copied from the incoming RuntimeConf) then we couldn't have found the cache file since ContainerID and IfName are part of the filename generated by getCacheFilePath()

dcbw: removed and a testcase added

squeed:
looks even better to me :+1:

dcbw:
@squeed added another commit to validate we can create a correct cache file name.

libcni/api.go
mars1024: emmmm... Good catch! lgtm except a small question, since filename includes ContainerID, IfName and NetworkName, why do we need to cache them in file content? Looks like they are not used anywhere. @dcbw

PR #705, opened 2019-08-22
libcni: fix cache file 'result' key name

dcbw:
The structure member name was changed from Results -> Result during
review but we missed changing the JSON key name.  Fix that and fall
back to the old name if required.

@containernetworking/cni-maintainers @bboreham @jellonek @squeed


bboreham:
I think it would be ok to just change the name and invalidate any cached data stored under the incorrect name.  It hasn't gone out in a release.

dcbw:
@bboreham updated; is this all we need?

PR #706, opened 2019-08-22
cnitool: fall back to default bindirs if not given via CNI_PATH

dcbw:
Otherwise you always have to use cnitool with CNI_PATH=xxxxx which is kinda annoying.


mars1024:
Maybe we can have some related documentations in README.md

PR #707, opened 2019-09-04
[WIP] libcni: add method to return all cached info for a container

dcbw:
This is a "get all the things" function, which a runtime could use on container teardown to avoid having to itself cache the uniqueness tuple when tearing down all networks for the container.

@containernetworking/cni-maintainers @squeed @mars1024


dcbw:
Closing for now, may come back to this later.

PR #711, opened 2019-09-17
Update cnitool docs

johscheuer:
Just some markdown cleanup and I added the used environment variables directly to the cnitool `README.md` and added a reference from the Documentation file.


squeed:
Very nice, thanks!

PR #712, opened 2019-09-18
add interface name validation

mars1024:
Interface name in linux is not unrestricted, so bad interface name will break CNI plugin.
This pr will add interface name validation to libcni and skel.

validation rules ref to [https://github.com/torvalds/linux/blob/master/net/core/dev.c#L1024](https://github.com/torvalds/linux/blob/master/net/core/dev.c#L1024)

### DIFF
--- a/libcni/api_test.go
+++ b/libcni/api_test.go
@@ -1012,7 +1012,7 @@ var _ = Describe("Invoking plugins", func() {
 				It("returns the error", func() {
 					_, err := cniConfig.AddNetworkList(ctx, netConfigList, runtimeConfig)
 					Expect(err).To(Equal(&types.Error{
-						Code:    4,
+						Code:    types.ErrInvalidEnvironmentVariables,
 						Msg:     "invalid characters in containerID",
 						Details: "some-%%container-id",
 					}))
@@ -1027,9 +1027,55 @@ var _ = Describe("Invoking plugins", func() {
 				It("returns the error", func() {
 					_, err := cniConfig.AddNetworkList(ctx, netConfigList, runtimeConfig)
 					Expect(err).To(Equal(&types.Error{
-						Code:    7,
+						Code:    types.ErrInvalidNetworkConfig,
 						Msg:     "invalid characters found in network name",
 						Details: "invalid-%%-name",
+					}))
+				})
+			})
+
+			Context("return errors when interface name is invalid", func() {
+				It("interface name is empty", func() {
+					runtimeConfig.IfName = ""
+
+					_, err := cniConfig.AddNetworkList(ctx, netConfigList, runtimeConfig)
+					Expect(err).To(Equal(&types.Error{
+						Code:    types.ErrInvalidEnvironmentVariables,
+						Msg:     "interface name is empty",
+						Details: "",
+					}))
+				})
+
+				It("interface name is overflow", func() {
+					runtimeConfig.IfName = "1234567890123456"
+
+					_, err := cniConfig.AddNetworkList(ctx, netConfigList, runtimeConfig)
+					Expect(err).To(Equal(&types.Error{
+						Code:    types.ErrInvalidEnvironmentVariables,
+						Msg:     "interface name is overflow",
+						Details: "interface name length should be less than 16 characters",
+					}))
+				})
+
+				It("interface name contains invalid characters /", func() {
+					runtimeConfig.IfName = "test/test"
+
+					_, err := cniConfig.AddNetworkList(ctx, netConfigList, runtimeConfig)
+					Expect(err).To(Equal(&types.Error{
+						Code:    types.ErrInvalidEnvironmentVariables,
+						Msg:     "interface name contains / or whitespace characters",
+						Details: "",
+					}))
+				})
+
+				It("interface name contains invalid characters whitespace", func() {
+					runtimeConfig.IfName = "test test"
+
+					_, err := cniConfig.AddNetworkList(ctx, netConfigList, runtimeConfig)
+					Expect(err).To(Equal(&types.Error{
+						Code:    types.ErrInvalidEnvironmentVariables,
+						Msg:     "interface name contains / or whitespace characters",
+						Details: "",
 					}))
 				})
 			})
--- a/libcni/api_test.go
+++ b/libcni/api_test.go
@@ -1046,14 +1046,36 @@ var _ = Describe("Invoking plugins", func() {
 					}))
 				})
 
-				It("interface name is overflow", func() {
+				It("interface name is too long", func() {
 					runtimeConfig.IfName = "1234567890123456"
 
 					_, err := cniConfig.AddNetworkList(ctx, netConfigList, runtimeConfig)
 					Expect(err).To(Equal(&types.Error{
 						Code:    types.ErrInvalidEnvironmentVariables,
-						Msg:     "interface name is overflow",
-						Details: "interface name length should be less than 16 characters",
+						Msg:     "interface name is too long",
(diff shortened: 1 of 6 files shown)


libcni/api_test.go
```diff
@@ -1027,13 +1027,48 @@ var _ = Describe("Invoking plugins", func() {
...
+
+					_, err := cniConfig.AddNetworkList(ctx, netConfigList, runtimeConfig)
+					Expect(err).To(Equal(&types.Error{
+						Code:    types.ErrInvalidEnvironmentVariables,
+						Msg:     "interface name is empty",
+						Details: "",
+					}))
+				})
+
+				It("returns the error when interface name is overflow", func() {
```
jellonek: is too long

pkg/skel/skel_test.go
```diff
@@ -119,12 +119,38 @@ var _ = Describe("dispatching to the correct callback", func() {
...
 			Expect(err).To(Equal(&types.Error{
-				Code:    4,
+				Code:    types.ErrInvalidEnvironmentVariables,
 				Msg:     "invalid characters in containerID",
 				Details: "some-%%container-id",
 			}))
 		})
 
+		Context("return error when interface name is invalid", func() {
+			It("returns the error when interface name is overflow", func() {
```
jellonek: ditto

pkg/skel/skel_test.go
```diff
@@ -119,12 +119,38 @@ var _ = Describe("dispatching to the correct callback", func() {
...
+			})
+
+			It("returns the error when interface name contains invalid characters", func() {
+				environment["CNI_IFNAME"] = "test/test"
+
+				err := dispatch.pluginMain(cmdAdd.Func, cmdCheck.Func, cmdDel.Func, versionInfo, "")
+				Expect(err).To(HaveOccurred())
+				Expect(err).To(Equal(&types.Error{
+					Code:    types.ErrInvalidEnvironmentVariables,
+					Msg:     "interface name contains / or whitespace characters",
```
jellonek: It would be nice to have two separate tests there for both cases, but I do not insist on that.

mars1024: updated

libcni/api_test.go
mars1024: updated

pkg/skel/skel_test.go
mars1024: updated

mars1024:
@jellonek all updated, ptal, thanks ~

pkg/utils/utils.go
```diff
@@ -49,3 +56,21 @@ func ValidateNetworkName(networkName string) *types.Error {
...
+// 1. The name must not be empty
+// 2. The name must be less than 16 characters
+// 3. The name must not contain / or any whitespace characters
+// ref to https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/utils.c?id=1f420318bda3cc62156e89e1b56d60cc744b48ad#n827
+func ValidateInterfaceName(ifName string) *types.Error {
+	if len(ifName) == 0 {
+		return types.NewError(types.ErrInvalidEnvironmentVariables, "interface name is empty", "")
+	}
+	if len(ifName) > maxInterfaceNameLength {
+		return types.NewError(types.ErrInvalidEnvironmentVariables, "interface name is overflow", fmt.Sprintf("interface name length should be less than %d characters", maxInterfaceNameLength+1))
```
jellonek: Please use there: "is too long" instead of "is overflow".

mars1024: got it

mars1024: done

mars1024: @jellonek  updated, ptal, thanks ~

pkg/utils/utils.go
```diff
@@ -49,3 +56,21 @@ func ValidateNetworkName(networkName string) *types.Error {
...
+// 3. The name must not contain / or any whitespace characters
+// ref to https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/utils.c?id=1f420318bda3cc62156e89e1b56d60cc744b48ad#n827
+func ValidateInterfaceName(ifName string) *types.Error {
+	if len(ifName) == 0 {
+		return types.NewError(types.ErrInvalidEnvironmentVariables, "interface name is empty", "")
+	}
+	if len(ifName) > maxInterfaceNameLength {
+		return types.NewError(types.ErrInvalidEnvironmentVariables, "interface name is too long", fmt.Sprintf("interface name should be less than %d characters", maxInterfaceNameLength+1))
+	}
+	if bytes.ContainsAny([]byte(ifName), `/ `) {
```
jellonek: Sorry that I didn't caught that earlier, but what about new line, tab or vertical tab (I mean - rest of whitespaces)?
Could use there https://golang.org/pkg/unicode/#White_Space ?

mars1024: @jellonek Thanks for your suggestion, updated!

pkg/utils/utils.go
```diff
@@ -49,3 +57,24 @@ func ValidateNetworkName(networkName string) *types.Error {
...
+// 3. The name must not contain / or any whitespace characters
+// ref to https://git.kernel.org/pub/scm/network/iproute2/iproute2.git/tree/lib/utils.c?id=1f420318bda3cc62156e89e1b56d60cc744b48ad#n827
+func ValidateInterfaceName(ifName string) *types.Error {
+	if len(ifName) == 0 {
+		return types.NewError(types.ErrInvalidEnvironmentVariables, "interface name is empty", "")
+	}
+	if len(ifName) > maxInterfaceNameLength {
+		return types.NewError(types.ErrInvalidEnvironmentVariables, "interface name is too long", fmt.Sprintf("interface name should be less than %d characters", maxInterfaceNameLength+1))
+	}
+	for _, r := range bytes.Runes([]byte(ifName)) {
```
squeed: I don't think we want to convert to runes here - the iproute2 code is totally not utf-aware. It's basically just doing `uint8` comparisons.

mars1024: So you think the original change ```bytes.ContainsAny([]byte(ifName), `/ `)``` is more reasonable? But @jellonek think we should check all whitespace-related characters (like \n, \t ...) I'm a little confused.

jellonek: @squeed but what is your suggestion? to list possible characters only from ascii range? or keep original proposal for only two characters

squeed: While this is technically slightly different from what `iproute2` does, I think this is probably fine.

dcbw:
For reference, from net/core/dev.c:
```
/**
 *	dev_valid_name - check if name is okay for network device
 *	@name: name string
 *
 *	Network device names need to be valid file names to
 *	to allow sysfs to work.  We also disallow any kind of
 *	whitespace.
 */
bool dev_valid_name(const char *name)
{
	if (*name == '\0')
		return false;
	if (strnlen(name, IFNAMSIZ) == IFNAMSIZ)
		return false;
	if (!strcmp(name, ".") || !strcmp(name, ".."))
		return false;

	while (*name) {
		if (*name == '/' || *name == ':' || isspace(*name))
			return false;
		name++;
	}
	return true;
}
```

squeed:
Aha, interesting. @mars1024 can you block "." and ".." too?

jellonek:
If we want to follow this kernel code - we should also block ":".

mars1024:
@dcbw @squeed @jellonek  Thanks for all your suggestions, I'll update this soon ~

mars1024:
Updated!

Issue #717, opened 2019-09-29
cni repo, README.md hyperlink cannot find the page


xichengliudui:
![image](https://user-images.githubusercontent.com/40875627/65825098-46fd5700-e2a5-11e9-9a39-6888bbf55036.png)
**Click on the hyperlink**
![image](https://user-images.githubusercontent.com/40875627/65825110-7b711300-e2a5-11e9-9eda-a7f7d3e98dca.png)

xichengliudui:
Can we create an images directory for these images? Just like logo.png, it's in the root directory

bboreham:
The underlying problem is #657

PR #718, opened 2019-09-29
update hyperlink

xichengliudui:
fixed: #717


xichengliudui:
I have deleted the link temporarily, this is not the correct repair method. May I ask who knows the correct link?

Issue #720, opened 2019-10-03
missing CNI version in config fails ValidateNetworkList()/ValidateNetwork()


dcbw:
A missing cniVersion in config JSON ends up as "". When config validation is called to match the config version with what plugins claim to support, most plugins return:

`var All = PluginSupports("0.1.0", "0.2.0", "0.3.0", "0.3.1", "0.4.0")`

from pkg/version/version.go, like Flannel does:

```
func main() {
	skel.PluginMain(cmdAdd, cmdCheck, cmdDel, version.All, bv.BuildString("flannel"))
}
```

But that fails validation because PluginSupports does not contain "", so validateNetwork() fails:

```
// validatePlugin checks that an individual plugin's configuration is sane
func (c *CNIConfig) validatePlugin(ctx context.Context, pluginName, expectedVersion string) error {
	c.ensureExec()
	pluginPath, err := c.exec.FindInPath(pluginName, c.Path)
	if err != nil {
		return err
	}

	vi, err := invoke.GetVersionInfo(ctx, pluginPath, c.exec)
	if err != nil {
		return err
	}
	for _, vers := range vi.SupportedVersions() {
===>	if vers == expectedVersion {
			return nil
		}
	}
	return fmt.Errorf("plugin %s does not support config version %q", pluginName, expectedVersion)
}
```

dcbw:
Possible solutions are:

- update validatePlugin to manually match expectedVersion=="" against 0.1.0 and 0.2.0
- make ValidateNetwork/ValidateNetworkList marshal the NetConf to bytes, call pkg/version/conf.go Decode(), and use that version rather than directly using list.CNIVersion that may be blank
- update libcni/conf.go to sub "" => "0.1.0" (but this doesn't cover NetConfs that people create themselves)
- make ValidateNetwork/ValidateNetworkList just manually do `if version == "" { version = "0.1.0" }`

dcbw:
@containernetworking/cni-maintainers @mccv1r0

squeed:
@bboreham and I like the first suggestion:

>     * update validatePlugin to manually match expectedVersion=="" against 0.1.0 and 0.2.0

PR #721, opened 2019-10-10
When CNI version isn't supplied in config, use default.

mccv1r0:
When CNI version isn't supplied in config, use default.

Fixes [720](https://github.com/containernetworking/cni/issues/720)

Signed-off-by: Michael Cambria <mccv1r0@gmail.com>

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -632,6 +632,9 @@ func (c *CNIConfig) validatePlugin(ctx context.Context, pluginName, expectedVers
 	if err != nil {
 		return err
 	}
+	if expectedVersion == "" {
+		expectedVersion = "0.1.0"
+	}
 
 	vi, err := invoke.GetVersionInfo(ctx, pluginPath, c.exec)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


libcni/api.go
```diff
@@ -632,6 +633,9 @@ func (c *CNIConfig) validatePlugin(ctx context.Context, pluginName, expectedVers
 	if err != nil {
 		return err
 	}
+	if expectedVersion == "" {
+		expectedVersion = types020.ImplementedSpecVersion
```
mars1024: According to the suggestion in [https://github.com/containernetworking/cni/issues/720#issuecomment-540046432](https://github.com/containernetworking/cni/issues/720#issuecomment-540046432), I think `expectedVersion` here should be 0.1.0 and 0.2.0 .

dcbw: @mars1024 @mccv1r0 I think we should probably hardcode this to "0.1.0" which is the closest version to "" that we have.  Effectively "" and 0.1.0 is equivalent.

PR #727, opened 2019-11-22
DOCUMENTATION: Address incomplete instructions in CONTRIBUTING.md

### LINKED ISSUE #726: Missing and incomplete instructions for running the test suite in CONTRIBUTING.md
I tried following the instructions in CONTRIBUTING.md and hit a few bumps in the road. It could be a combination of my newbie status with golang and some missing steps in the docs.In the section https://github.com/containernetworking/cni/blob/master/CONTRIBUTING.md#how-to-run-the-test-suite , I stumbled over a few things:

1. Simply invoking ./test.sh resulted in some dependency errors.  I was able to satisfy them manually. I'm still new to the golang ecosystem so I could be doing something wrong, but it seems like a little tooling to pull in the proper dependencies when invoking ./test.sh would be a nice touch. At a minimum, mentioning the need to resolve dependencies like ginkgo and gomega (the ones I was missing) is probably in order.

2. On master, the guide references cd'ing to plugins/main/loopback, then invoking 'go test'. That path does not exist. It might be a nice touch to clean this up.

These are minor issues, but were nonetheless impediments to me getting started with things.

rktidwell:
This change addresses a few missing steps and out-of-date items in the
contributor docs such as fetching proper dependencies and removing
references to paths that no longer exist.

Fixes #726

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -62,7 +62,7 @@ cd /go/src/github.com/containernetworking/cni
 ./test.sh
 
 # to focus on a particular test suite
-cd plugins/main/loopback
+cd libcni
 go test
 ```
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -17,6 +17,9 @@ Vagrant.configure(2) do |config|
     echo 'export GOPATH=/go; export PATH=/usr/local/go/bin:$GOPATH/bin:$PATH' >> /root/.bashrc
     eval `tail -n1 /root/.bashrc`
 
+    go get github.com/onsi/ginkgo/ginkgo
+    go get github.com/onsi/gomega/...
+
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -56,13 +56,15 @@ vagrant up
 vagrant ssh
 # you're now in a shell in a virtual machine
 sudo su
+go get github.com/onsi/ginkgo/ginkgo
```
mars1024: To avoid getting every time, could you help to move this two lines to `Vagrantfile`?

rktidwell: Done

bboreham:
Looks reasonable but I would like @squeed to cast an eye over the `go get` commands - are we expecting that to be necessary?

PR #731, opened 2019-12-03
Add Project Antrea in CNI list

yasensim:


bboreham:
Any particular reason why this takes the 2nd spot?

yasensim:
no real reason. Usually it is alphabetical order on the other lists. I don't mind reordering it. just let me know where to put it and I will do.

mars1024:
I think appending is better than inserting. Actually the order is only based on time.

yasensim:
@mars1024  done!

mars1024:
> @mars1024 done!

Thanks, lgtm!

PR #734, opened 2019-12-17
Compatible with '.conf' suffix

smarkm:
Signed-off-by: smark <smark@freecoop.net>

### DIFF
--- a/libcni/conf.go
+++ b/libcni/conf.go
@@ -183,7 +183,7 @@ func LoadConf(dir, name string) (*NetworkConfig, error) {
 }
 
 func LoadConfList(dir, name string) (*NetworkConfigList, error) {
-	files, err := ConfFiles(dir, []string{".conflist"})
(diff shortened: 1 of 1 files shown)


squeed:
Hmm - I don't think we want to do this. The `.conf` suffix indicates it's a single configuration file, whereas `.conflist` means it's a conflist. Is there something I'm missing?

smarkm:
> Hmm - I don't think we want to do this. The `.conf` suffix indicates it's a single configuration file, whereas `.conflist` means it's a conflist. Is there something I'm missing?

smarkm:
> Hmm - I don't think we want to do this. The `.conf` suffix indicates it's a single configuration file, whereas `.conflist` means it's a conflist. Is there something I'm missing?

No @squeed ,you are right 
Sorry It's my misunderstanding(I am trying to put configlist in .conf file) and I will close this PR

Issue #732, opened 2019-12-06
if plugin segfaults/panics, invoke.ExecPlugin... claims it "failed with no error message"


danwinship:
If a plugin exits with a non-0 exit status and prints stuff to stderr *but not stdout*, then `pkg/invoke/raw_exec.go:pluginErr()` will categorize this as "netplugin failed with no error message". This seems to be in part because it forwards the plugin's stderr to its parent's stderr, which is nice and all, but it means that the stderr might end up getting logged somewhere less useful (eg, in cri-o's logs rather than kubelet's). Plus, the "failed with no error message" part is just wrong.

(Seen in two unrelated contexts with different plugins in the same day, lol)

@dcbw @squeed

squeed:
Yeah, that makes sense. We give plugins the opportunity to output a reasonable error message via json, but that's not always possible.

squeed:
While we're at it, we should interpret the exit code if it indicates abnormal termination.

zjj2wry:
Encountered this problem. kubectl describe pod will only get netplugin failed with no error message

bboreham:
Reopened because I'm not sure how #735 works.

dcbw:
@bboreham can you describe what doesn't sound right to you about #735?

dcbw:
Yeah, I can't see where `c.Run()` in raw_exec.go will create an ExitError that includes stderr output.

PR #735, opened 2019-12-19
intercept netplugin std error

zjj2wry:
fixes https://github.com/containernetworking/cni/issues/732


bboreham:
Actually, it looks like we would only get this if we called `Output()` on the command rather than `Run()`.  @zjj2wry did you test this?

PR #736, opened 2019-12-20
pkg/utils: add utils_test

q384566678:
Add corresponding unit tests to the functions in util.go.

Signed-off-by: Zhou Hao <zhouhao@cn.fujitsu.com>

### DIFF
--- a/pkg/utils/utils_test.go
+++ b/pkg/utils/utils_test.go
@@ -1,0 +1,130 @@
+// Copyright 2019 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


pkg/utils/utils_test.go
```diff
@@ -0,0 +1,136 @@
...
+		{
+			description: "normal containerID",
+			containerID: "a51debf7e1eb",
+			err:         false,
+		},
+	}
+
+	for _, tt := range testData {
+		err := ValidateContainerID(tt.containerID)
+		if err != nil && !tt.err {
```
mars1024: Could you help to use DeepEqual to squash these lines?

mars1024:
BTW, commits need to be squashed either.

pkg/utils/utils_test.go
q384566678: They are different structure types, and it seems that it is not suitable to use DeepEqual. If they are changed to agree types and then compared, more code will be generated.

mars1024: FYI https://play.golang.org/p/DicOkjUajmv

q384566678: updated, thanks for your suggestions. PTAL

Issue #741, opened 2020-01-20
Support infiniband GUID in Capabilities / Runtime Configuration


adrianchiris:
In order to support container network connectivity for infiniband fabric a CNI may require to configure the Infiniband interface GUID before moving the network interface to the pod namespace much like is done today with the [mac Capability/ Runtime config ](https://github.com/containernetworking/cni/blob/master/CONVENTIONS.md#well-known-capabilities)for ethernet fabrics.

I would like to propose an addition of the `GUID` capability/ Runtime Configuration which will be consumed initially by [ib-sriov-cni](https://github.com/Mellanox/ib-sriov-cni).

while the `mac` address field may be reused for this purpose it will not be accurate as `mac` address field, as defined by the spec, represents an Interface hardware address.

For IPoIB, an interface's hardware address is composed of 20 octets but only the first (lower) 8 octets are configurable by the user.
According to section 9.1.1 in IPoIB [RFC](https://tools.ietf.org/html/rfc4391):

- bytes 0-3 bytes are flags and queue pair number which user cannot control
- bytes 4-19 GID which is composed as follows (can be found in IBTA Infiniband Specification):
- 8 bytes subnet prefix - set by the infiniband subnet manager (initial value is 0xfe80000000000000)
- 8 bytes GUID - **the element which is user configurable**

As a side note, a CNI handling IPoIB interfaces is expected to return the full 20 octet hardware address in `mac` field under `interfaces` when returning result of an `ADD` command

mars1024:
Actually CNI capabilities/runtime configuration is more like a specification than a specific behavior, if you want to add some kind of capabilities, CNI Manager(like ocicni, cniNetworkPlugin in kubernetes) and CNI plugins should be changed, not CNI itself.

adrianchiris:
Hi, So the goal is to define GUID in the specification (like mac today).
Once we have this well defined in the spec, we will do the needed changes in relevant projects to consume this capabilities/runtimeconfig attribute.

for now we plan to:
1. implement this in [ib-sriov-cni](https://github.com/Mellanox/ib-sriov-cni) CNI plugin
2. implement in multus to pass runtimeConfig to CNI

if needed in the future, support shall be added to CNI manager.

jellonek:
[...]

PR #742, opened 2020-01-28
Add GUID to well known Capabilities

adrianchiris:
This commit extends CONVENTION.md Well-known Capabilities
to inclue `guid`. this shall be used by CNI plugins who wish
to set Infiniband GUID address for a network interface.

This addresses issue #741

### DIFF
--- a/CONVENTIONS.md
+++ b/CONVENTIONS.md
@@ -63,6 +63,7 @@ But the runtime would fill in the mappings so the plugin itself would receive so
 | dns | Dynamically configure dns according to runtime | `dns` | Dictionary containing a list of `servers` (string entries), a list of `searches` (string entries), a list of `options` (string entries). <pre>{ <br> "searches" : [ "internal.yoyodyne.net", "corp.tyrell.net" ] <br> "servers": [ "8.8.8.8", "10.0.0.10" ] <br />} </pre> | kubernetes | CNI `win-bridge` plugin, CNI `win-overlay` plugin |
 | ips | Dynamically allocate IPs for container interface. Runtime which has the ability of address allocation can pass these to plugins.  | `ips` | A list of `IP` (string entries). <pre> [ "10.10.0.1/24", "3ffe:ffff:0:01ff::1/64" ] </pre> | none | CNI `static` plugin |
 | mac | Dynamically assign MAC. Runtime can pass this to plugins which need MAC as input. | `mac` | `MAC` (string entry). <pre> "c2:11:22:33:44:55" </pre> | none | CNI `tuning` plugin |
+| infiniband_guid | Dynamically assign Infiniband GUID to network interface. Runtime can pass this to plugins which need Infiniband GUID as input. | `infiniband_guid` | `GUID` (string entry). <pre> "c2:11:22:33:44:55:66:77" </pre> | none | CNI [`ib-sriov-cni`](https://github.com/Mellanox/ib-sriov-cni) plugin |
 
 ## "args" in network config
 `args` in [network config](https://github.com/containernetworking/cni/blob/master/SPEC.md#network-configuration) were introduced as an optional field into the `0.2.0` release of the CNI spec. The first CNI code release that it appeared in was `v0.4.0`. 


CONVENTIONS.md
```diff
@@ -63,6 +63,7 @@ But the runtime would fill in the mappings so the plugin itself would receive so
 | dns | Dynamically configure dns according to runtime | `dns` | Dictionary containing a list of `servers` (string entries), a list of `searches` (string entries), a list of `options` (string entries). <pre>{ <br> "searches" : [ "internal.yoyodyne.net", "corp.tyrell.net" ] <br> "servers": [ "8.8.8.8", "10.0.0.10" ] <br />} </pre> | kubernetes | CNI `win-bridge` plugin, CNI `win-overlay` plugin |
 | ips | Dynamically allocate IPs for container interface. Runtime which has the ability of address allocation can pass these to plugins.  | `ips` | A list of `IP` (string entries). <pre> [ "10.10.0.1/24", "3ffe:ffff:0:01ff::1/64" ] </pre> | none | CNI `static` plugin |
 | mac | Dynamically assign MAC. Runtime can pass this to plugins which need MAC as input. | `mac` | `MAC` (string entry). <pre> "c2:11:22:33:44:55" </pre> | none | CNI `tuning` plugin |
+| guid | Dynamically assign GUID to network interface. Runtime can pass this to plugins which need GUID as input. | `guid` | `GUID` (string entry). <pre> "c2:11:22:33:44:55:66:77" </pre> | none | CNI `ib-sriov-cni` plugin |
```
mars1024: two nits:
- Runtime implementations of guid should not be `none`, maybe `multus`?
- Plugin Implementations of guid is better to be a link because it's a 3rd party plugin

moshe010: -  Yes the plan is to support it it in multus I will update the PR
- Yes I will update with the link to the ib-sriov-cni  in Mellanox github

jellonek: IMO calling Multus as runtime is not a good direction. It's multiplexer which is called by "real" runtime, so IMO `none` is there more correct.
+1 for url

moshe010:
> Small nit: you have typo in commit message - s/inclue/include/

Fixed

CONVENTIONS.md
moshe010: Ok, so for now I just update the URL for the ib-sriov-cni

moshe010:
Can we merge this PR?

mars1024:
> Can we merge this PR?

I think we should block this because this is still under discussion in Network Plumbing Working Group Meeting by @dcbw .

moshe010:
@mars1024, I will join the next Network Plumbing Working Group Meeting to give all the information

dcbw:
@moshe010 good to hear, let's get some clarification on it then.

jellonek:
Any update after the meeting?

adrianchiris:
I believe all issues were clarified, @dcbw can we approve this PR ?
(i'll also work on the changes discussed for the multi-net-spec)

dcbw:
@adrianchiris @moshe010 would you be opposed to keeping the convention to use `mac` but fix the wording to account for GUID? Somewhat less perferable would be to add an `hwaddr` option that works for both MAC and GUID explicitly but that would require adding support to implementations.

adrianchiris:
The definition of interfaces.Mac as defined is SPEC.md is fine IMO as it basically states its the L2 hardware address (be that what it may for different L2 fabrics). This general statement fits multiple fabrics. 

If you feel this needs clarification then i can add some clarifications that on Ethernet fabric it represent the 6 bytes hardware address and on Infiniband fabric it represent the 20 bytes IP over Infiniband hardware address.

overloading `mac` field in runtime config for GUID will introduce some confusion (for infiniband users) as its only a part (which is user configurable) of an IP over Infiniband L2 hardware address (the GUID is the lower 8 bytes).
For correctness and clarity i prefer to have a separate attribute.

as an example: When using ip tool you cannot set mac address for an `ipoib` network interface you can set `port_guid`

adrianchiris:
@dcbw any inputs ?

dcbw:
@adrianchiris Since GUID is a fairly generic term, would there be value in going with "ib_guid" or something like that?

dcbw:
@adrianchiris would you mind also extending the tuning plugin to implement changing an IB GUID too? That way we have an implementation in the reference plugin set.

adrianchiris:
> @adrianchiris Since GUID is a fairly generic term, would there be value in going with "ib_guid" or something like that?

we can go with  `infiniband_guid` , will update the PR.

adrianchiris:
> @adrianchiris would you mind also extending the tuning plugin to implement changing an IB GUID too? That way we have an implementation in the reference plugin set.
I will need to investigate more.
for Infiniband SR-IOV we may need to set it in one go so not sure running it in a chain is the correct approach.

For infiniband non SR-IOV (ipoib child device) - need to look into this more.

dcbw:
@adrianchiris kernel's drivers/infiniband/ulp/ipoib/ipoib.c accepts a GUID via the normal 'ip link set address` mechanisms via netlink, and does this:

```
static int ipoib_set_mac(struct net_device *dev, void *addr)
{
	struct ipoib_dev_priv *priv = ipoib_priv(dev);
	struct sockaddr_storage *ss = addr;
	int ret;

	if (!(dev->priv_flags & IFF_LIVE_ADDR_CHANGE) && netif_running(dev))
		return -EBUSY;

	ret = ipoib_check_lladdr(dev, ss);
	if (ret)
		return ret;

	set_base_guid(priv, (union ib_gid *)(ss->__data + 4));

	queue_work(ipoib_workqueue, &priv->flush_light);

	return 0;
}
```

So I think you could just use the normal vishvananda/netlink stuff to set the address on an ipoib device in tuning.

adrianchiris:
@dcbw looking further into the code:
**For IPoIB interface over VF :** 

`ndo_set_mac_address` is not implemented [1] with current kernel implementation, one would need to set the VF's Port GUID on the PF with the VF's index, then unbind/bind the driver on the VF. - this would need to be performed in a single go by, for example, [ib-sriov-cni](https://github.com/Mellanox/ib-sriov-cni/). This may change in the future then, we could then set it via tuning.

**For IPoIB interface over PF:**
One would need to modify just the "GUID" part of the hardware address (lower 8 bytes) or else it would fail validation in kernel. If the device does not support live address change (e.g for the IPoIB interface which is directly associated with the PF), setting the link state to down is also needed

So for the latter we _can_ use tuning to set the GUID.
One could use [ipoib-cni](https://github.com/Mellanox/ipoib-cni)  together with tuning.

If there is a consensus, i can work on adding support for `infiniband_guid` runtime config in tuning.

[1] https://github.com/torvalds/linux/blob/5076190daded2197f62fe92cf69674488be44175/drivers/infiniband/ulp/ipoib/ipoib_main.c#L2064

Issue #733, opened 2019-12-13
Describe how to use comments in configuration files (or add support for them)


ensc:
Examples in the documentation suggest that `//` can be used for comments. E.g.  the [CNI spec](https://github.com/containernetworking/cni/blob/master/SPEC.md#example-configurations)
contains an example like

```
{
  "cniVersion": "0.4.0",
  "name": "dbnet",
  "type": "bridge",
  // type (plugin) specific
  "bridge": "cni0",
  ...
```

Unfortunately, using such a configuration with `podman` results into

```
Error: error parsing configuration list: invalid character '/' looking for beginning of value
```

Because configuration files seem to be JSON (which does not support comments at all), it should be clarified how to write comments in configuration files.  If not supported (and CNI spec is wrong), such an extension (`//`) should be explicitly required because comments are essential in configuration files.

mars1024:
Oh, I'm afraid this is a mistake, the comment lines is belong to the code segment but not the configuration JSON itself.

dcbw:
This is a good point, if the examples cannot be copied and pasted then we should probably change them.

mars1024:
@ensc Could you help to pull a request to fix this?

ensc:
@mars1024   my problem is that I want to comment parts of the configuration file (for documentation purposes or to disable some option temporarily).

jellonek:
That seams reasonable also to me. Disabling an option to have comments in *configuration* only because it's based on json seems to be invalid also for me.

PR #754, opened 2020-02-23
Fix linting issues in docs, add headers to json example, update errors into table

asears:
Relates #733

@mars1024 I took a look at this issue and here is a PR to update some of the documentation / remove the comments in the docs.  Addressed some linting issues while I was reading the docs.  

First time contributor here, let me know if there's any issues or improvements to be made with the PR.

Unfortunately JSON is well-known not to support comments, JSONC requires translation, then there is [hjson](https://github.com/hjson/hjson/issues/84).

Adding a dummy "_comment" or "_annotation" tag might work if it's truly required inside the JSON code, otherwise I moved any comments outside of the code itself to address one of the issues in this issue.

-- update - adding DCO to fix issue
Signed-off-by: asears <asears@users.noreply.github.com>


mars1024:
Thanks @asears , seems a lot of changes in SPEC, we should be slow and careful, I think there will be some feedbacks after maintainer meeting of this week :)

asears:
@mars1024 let me know if any that are questionable and we can revert the change, or if I should add additional reviewers to the pull request to speed up / improve the PR.

bboreham:
Can you explain "linting issues"?  Did you run some tool over the files?

I would find it easier to review if you break out the changes by type into separate commits, e.g. all the JSON changes removing comments, adding blank lines, converting "1,2,3" to "1,1,1", converting "-" bullets to numbered bullets, etc.

For example I cannot immediately see what changed on the line beginning "The runtime may also pass capability-based keys", and you can probably explain quite quickly in a commit message.

bboreham:
Some JSON in the spec was stated as an "example", and I agree it is confusing that you cannot just copy-paste it because it contains comments.  But elsewhere, for instance:

```
{
  "cniVersion": "0.4.0",
  "interfaces": [                                            (this key omitted by IPAM plugins)
      {
          "name": "<name>",
          "mac": "<MAC address>",                            (required if L2 addresses are meaningful)
          "sandbox": "<netns path or hypervisor identifier>" (required for container/hypervisor interfaces, empty/omitted for host interfaces)
      }
  ],
```

The text on the right isn't marked by `//` and the result after stripping those could never be copy-pasted - you need to substitute all the `<>` placeholders.  So I question the value of rewording these parts to put all the commentary at the end.

bboreham:
What is the meaning of marking some sections as `json` and some as `jsonc` ?

asears:
Thanks for the feedback Bryan @bboreham.  I can see that this document has had a lot of history, hope only to improve on it!

### Markdown Linter

This tool picked up the linting issues when I opened the document.  

https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint


### JSONC format

JSONC markdown formatting keyword was there originally, perhaps the author identified comments were possible?  If there's no comments or support for them it could probably be replaced with json.  Don't know that the hidden markdown format matters, though it is relevant to this PR / referenced issue.

Here's a post around support for jsonc added to markdown that might be relevant for other apps, though probably not for these configuration files. 
https://github.com/github/linguist/issues/4153
https://komkom.github.io/

### Invalid JSON

The original issue referenced was that there was invalid JSON / comments in the JSON, so I removed those.  There still may be some invalid JSON there due to missing quotes, etc.  I didn't scan each block.    Would an implementer go through this spec and copy each block of code to implement, or copy, substitute, remove comments to fix broken JSON?  Perhaps example JSON code should be used to implement and not the spec doc itself.  In that case, perhaps comments and pseudo-code are valid.

### Differences vs. master

Generally I simplify commits when checking in, don't know that I can explode things back out.
The line mentioned appears to have been just formatted / line feed added.

I use this to compare differences with WinMerge and line wrap turned on.  The line you mentioned only has formatting differences.  I split the last line into its own line.
```
git difftool master -- SPEC.md
```

image

bboreham:
> JSONC markdown formatting keyword was there originally
OK, but when I look at your diffs I see you added some with `json` and some with `jsonc`, so I ask you to account for your choices.

SPEC.md
```diff
@@ -768,11 +778,13 @@ The `dns` field contains a dictionary consisting of common DNS information.
...
 
-- `1` - Incompatible CNI version
-- `2` - Unsupported field in network configuration. The error message must contain the key and value of the unsupported field.
-- `3` - Container unknown or does not exist. This error implies the runtime does not need to perform any container network cleanup (for example, calling the `DEL` action on the container).
-- `4` - Invalid necessary environment variables, like CNI_COMMAND, CNI_CONTAINERID, etc. The error message must contain the names of invalid variables.
-- `5` - I/O failure. For example, failed to read network config bytes from stdin.
-- `6` - Failed to decode content. For example, failed to unmarshal network config from bytes or failed to decode version info from string.
-- `7` - Invalid network config. If some validations on network configs do not pass, this error will be raised.
-- `11` - Try again later. If the plugin detects some transient condition that should clear up, it can use this code to notify the runtime it should re-try the operation later.
+Error Code|Error Description
```
mars1024: This change does not look as expectation, it is still not a table.

SPEC.md
```diff
@@ -255,6 +256,8 @@ Plugins may define additional fields that they accept and may generate an error
 
 ### Example configurations
 
+#### Example dbnet configuration
```
bboreham:
```suggestion
#### Example bridge configuration
```

SPEC.md
```diff
@@ -274,6 +277,8 @@ Plugins may define additional fields that they accept and may generate an error
 }
 ```
 
+#### Example pci configuration
```
bboreham:
```suggestion
#### Example ovs configuration
```

SPEC.md
```diff
@@ -295,6 +300,8 @@ Plugins may define additional fields that they accept and may generate an error
 }
 ```
 
+#### Example wan configuration
```
bboreham:
```suggestion
#### Example macvlan configuration
```

SPEC.md
```diff
@@ -317,13 +324,16 @@ Network configuration lists provide a mechanism to run multiple CNI plugins for
...
-The runtime may also pass capability-based keys as a map in the top-level `runtimeConfig` key of the plugin's config JSON if a plugin advertises it supports a specific capability via the `capabilities` key of its network configuration.  The key passed in `runtimeConfig` MUST match the name of the specific capability from the `capabilities` key of the plugins network configuration. See CONVENTIONS.md for more information on capabilities and how they are sent to plugins via the `runtimeConfig` key.
+The runtime may also pass capability-based keys as a map in the top-level `runtimeConfig` key of the plugin's config JSON if a plugin advertises it supports a specific capability via the `capabilities` key of its network configuration.  The key passed in `runtimeConfig` MUST match the name of the specific capability from the `capabilities` key of the plugins network configuration.
+
+See CONVENTIONS.md for more information on capabilities and how they are sent to plugins via the `runtimeConfig` key.
```
bboreham: This sentence goes with the one before; I don't think it should be in a separate paragraph.

SPEC.md
```diff
@@ -430,16 +440,16 @@ Note that the runtime adds the `cniVersion` and `name` fields from configuration
     "interfaces": [
         {
             "name": "cni0",
-            "mac": "00:11:22:33:44:55"
+            "mac": "00:11:22:33:44:55",
```
bboreham: Trailing commas are not valid JSON.

SPEC.md
asears: Applied change, will need to verify

SPEC.md
asears: Removed trailing commas.  This is where it may make sense to move all JSON and code outside of markdown to separate files and reference the files instead.  Files can be then linted/scanned as part of this build process.

SPEC.md
asears: Looked good in VSCode Markdown preview, Github markdown renders differently.  Have updated for Github markdown.

squeed:
The changes look great! Can you do one more thing: squash your commits and add a Signed-Off-By, so the DCO check passes? Thanks.

asears:
Unfortunately the DCO commands being recommended appear to be causing signing commits other than mine and causing merge conflicts.  Do I really want to run 
```
git rebase HEAD~27 --signoff
```
If someone can provide me the commands as appropriate to remove these commits, squash and pass the DCO checks that would be great.  Otherwise I'll abandon and start again.

SPEC.md
```diff
@@ -768,11 +789,13 @@ The `dns` field contains a dictionary consisting of common DNS information.
...
-- `2` - Unsupported field in network configuration. The error message must contain the key and value of the unsupported field.
-- `3` - Container unknown or does not exist. This error implies the runtime does not need to perform any container network cleanup (for example, calling the `DEL` action on the container).
-- `4` - Invalid necessary environment variables, like CNI_COMMAND, CNI_CONTAINERID, etc. The error message must contain the names of invalid variables.
-- `5` - I/O failure. For example, failed to read network config bytes from stdin.
-- `6` - Failed to decode content. For example, failed to unmarshal network config from bytes or failed to decode version info from string.
-- `7` - Invalid network config. If some validations on network configs do not pass, this error will be raised.
-- `11` - Try again later. If the plugin detects some transient condition that should clear up, it can use this code to notify the runtime it should re-try the operation later.
+Error Code|Error Description
+---|---
+`-1`|Incompatible CNI version
```
mars1024, line 794: I think here should be `1` not `-1`

mars1024:
BTW, could you help to clean the commits list, seems having a lot of unrelated commits...

bboreham:
@asears are you likely to come back to this?

asears:
Closing the PR as DCO plugin not cooperating.  Opened #768

PR #759, opened 2020-04-01
Capture stderr from plugin

### LINKED ISSUE #732: if plugin segfaults/panics, invoke.ExecPlugin... claims it "failed with no error message"

bboreham:
Fixes #732 

Example from panic (using 'noop'):
```
netplugin failed: "panic: test panic\n\ngoroutine 1 [running]:\nmain.main()\n\tplugins/test/noop/main.go:220 +0x39\n"
```

If stderr is also blank I add the OS-level error:
```
netplugin failed with no error message: exit status 1
```

### DIFF
--- a/pkg/invoke/raw_exec.go
+++ b/pkg/invoke/raw_exec.go
@@ -31,26 +31,27 @@ type RawExec struct {
 
 func (e *RawExec) ExecPlugin(ctx context.Context, pluginPath string, stdinData []byte, environ []string) ([]byte, error) {
 	stdout := &bytes.Buffer{}
+	stderr := &bytes.Buffer{}
 	c := exec.CommandContext(ctx, pluginPath)
 	c.Env = environ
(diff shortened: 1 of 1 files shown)


dcbw:
See also https://github.com/containernetworking/cni/pull/760

bboreham:
closing in favour of #760

PR #762, opened 2020-04-03
add networkScaling in Network Configuration

zhlhahaha:
networkScaling enable user to turn on RPS/RFS
feature of virtual network card for container

This feature increase datacache hitrate by
steering kernel processing of packets to
the CPU where the application thread
consuming the packet is running.

For RPS/RFS detail, please refer to:
https://www.kernel.org/doc/Documentation/
networking/scaling.txt

Signed-off-by: Howard Zhang <howard.zhang@arm.com>

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -243,6 +243,7 @@ The network configuration is described in JSON form. The configuration may be st
 - `type` (string): Refers to the filename of the CNI plugin executable.
 - `args` (dictionary, optional): Additional arguments provided by the container runtime. For example a dictionary of labels could be passed to CNI plugins by adding them to a labels field under `args`.
 - `ipMasq` (boolean, optional): If supported by the plugin, sets up an IP masquerade on the host for this network. This is necessary if the host will act as a gateway to subnets that are not able to route to the IP assigned to the container.
+- `networkScaling` (boolean, optional): If supported by the plugin, sets up an RPS/RFS for this network.
 - `ipam` (dictionary, optional): Dictionary with IPAM specific values:
   - `type` (string): Refers to the filename of the IPAM plugin executable.
 - `dns` (dictionary, optional): Dictionary with DNS specific values:


zhlhahaha:
@lubinsz

bboreham:
Thanks for the PR.  Can you give some examples of plugins where this feature has been implemented?

Nowadays we often note optional things in the CONVENTIONS doc, because new versions of the spec need a lot more care.

lubinszARM:
Hi @bboreham 
This new optional configuration is based on the kernel networking scaling feature, please see the link as reference:
https://github.com/torvalds/linux/blob/master/Documentation/networking/scaling.rst

Currently,  more and more 50G/100G network cards are used in the production field.
On many platforms, cross-host CNI cannot reach 50G/100G network speed due to the weak single-core performance.
Also, networkscaling has been widely used in kubernetes-loadbalancer in the production field.
Please see a example as reference:
https://gist.github.com/wsgzao/18828f69147635f3e38a14690a633daf

We have tested some CNIs with our networkscalig patch internally, and the cross-host performance can reach the full speed of physical NIC.
Our initial plan is:
* [ ]  add an option about networkScaling in CNI configuration
* [x] add the networkScaling code for a typical CNI example, such as bridge, and pass internal tests 
* [ ] deliver a PR to enable networkScaling in some typical CNI examples in https://github.com/containernetworking/plugins
* [x] add the networkScaling code for a popular CNI, such as flannel, and pass internal tests and kubernetes conformance test.\
 * [ ] deliver a PR to enable networkScaling in flannel

What do you think of our plan?
Should we deliver the patch into some typcial CNI-s firstly? Then add the option in CNI?
Thanks.

mars1024:
Not sure if this field is **general** or **well-known** enough to be put here..., just as @bboreham said, this field seems more like a **capability** to be put in CONVENTIONS.

BTW, if RPS/RFS are all about some kernel parameter settings, [tuning plugin](https://github.com/containernetworking/plugins/tree/master/plugins/meta/tuning) will help you.

squeed:
CNI capability conventions are intended for configuration knobs that change on a per-container basis. Given that scaling seems like something that would always be enabled for a given host, I think it should just be a configuration parameter in individual plugins.

dcbw:
@zhlhahaha @lubinszARM why would you ever turn this feature off?

squeed:
Closing this for now; feel free to pick this back up.

PR #763, opened 2020-04-03
Retry exec commands on text file busy

saschagrunert:
This makes the plugin validation a bit more robust by retrying the
execution if the plugin is in text file busy state. This can happen if
we write the config and the plugin at once. To avoid such races we now
try up to 5 seconds for the plugin validation.

### DIFF
--- a/pkg/invoke/raw_exec.go
+++ b/pkg/invoke/raw_exec.go
@@ -21,6 +21,8 @@ import (
 	"fmt"
 	"io"
 	"os/exec"
+	"strings"
+	"time"
 
 	"github.com/containernetworking/cni/pkg/types"
 )
@@ -37,7 +39,24 @@ func (e *RawExec) ExecPlugin(ctx context.Context, pluginPath string, stdinData [
 	c.Stdin = bytes.NewBuffer(stdinData)
 	c.Stdout = stdout
 	c.Stderr = stderr
-	if err := c.Run(); err != nil {
+
+	// Retry the command on "text file busy" errors
+	for i := 0; i <= 5; i++ {
+		err := c.Run()
+
+		// Command succeeded
+		if err == nil {
+			break
+		}
+
+		// If the plugin is currently about to be written, then we wait a
+		// second and try it again
+		if strings.Contains(err.Error(), "text file busy") {
+			time.Sleep(time.Second)
+			continue
+		}
+
+		// All other errors except than the busy text file
 		return nil, e.pluginErr(err, stdout.Bytes(), stderr.Bytes())
 	}
 


libcni/api.go
```diff
@@ -636,10 +637,26 @@ func (c *CNIConfig) validatePlugin(ctx context.Context, pluginName, expectedVers
...
+		vi, err = invoke.GetVersionInfo(ctx, pluginPath, c.exec)
+
+		// Found the version
+		if err == nil {
+			break
+		}
+
+		// If the plugin is currently about to be written, then we wait a
+		// second and try it again
+		if strings.Contains(err.Error(), "text file busy") {
```
mars1024, line 651: Why this kind of error should be catched specially? If we catch it here, should we catch it in add, del, check functions?
IMO, I think this error will exist briefly, so maybe we can rely on the retry from upstreams like kubelet.

saschagrunert: The kubelet has no context about that scenario. For example:

1. Container runtime gets started without configured CNI, but is looking for configs in /etc/cni/net.d
2. The kubelet starts and schedules the first container initializing the CNI plugin. This copies the config and the CNI binary to a local path.
3. During the copy, the runtime already picks-up the config and tries to verify the CNI version, which results in text file busy.

Sure, if we rely on the right order of copy binary and config then we don't have do do anything at all.

mars1024: Actually dockershim-cni in kubelet is scanning CNI config every 5 seconds, just as [https://github.com/kubernetes/kubernetes/blob/0926c9c47677e944c2dadcd7b39587866331eeba/pkg/kubelet/dockershim/network/cni/cni.go#L229](https://github.com/kubernetes/kubernetes/blob/0926c9c47677e944c2dadcd7b39587866331eeba/pkg/kubelet/dockershim/network/cni/cni.go#L229).

So I think in your scenario, the error won't live for over 5 seconds, right?

saschagrunert: Yes, that's right. If you think that the CNI container should handle this then feel free to close this PR. :)

squeed: Well, I like this, but other maintainers seem to disagree :-)

dcbw: I don't really like it and think that its mostly a runtime issue (since the same thing can happen on ADD/DEL/etc), but I don't actually care enough and it does fix a problem.

/lgtm

mars1024: > Well, I like this, but other maintainers seem to disagree :-)

Haha, "other" means me? Actually I have no objections to this PR, but only curious about its original intention and wants to know more details. If there are enough approvers, i won't be a blocker.

mars1024: BTW, since dan and I all mentioned this error would happen not only in validation but also in ADD/DEL/etc, could you help to make this catch more general? Like catching and retrying in [https://github.com/containernetworking/cni/blob/bf84331c94b76d6fb3e21fb366edb9dac2c90d63/pkg/invoke/raw_exec.go#L28](https://github.com/containernetworking/cni/blob/bf84331c94b76d6fb3e21fb366edb9dac2c90d63/pkg/invoke/raw_exec.go#L28) ?

@saschagrunert

saschagrunert: I can follow-up on that if you don't mind. Sounds like a good addition to me. :+1:

squeed:
As was mentioned in the discussion above, can you move this logic to RawExec (good point, @mars1024). Then we can get this in.

pkg/invoke/raw_exec.go
```diff
@@ -36,7 +38,24 @@ func (e *RawExec) ExecPlugin(ctx context.Context, pluginPath string, stdinData [
...
+		// If the plugin is currently about to be written, then we wait a
+		// second and try it again
+		if strings.Contains(err.Error(), "text file busy") {
+			time.Sleep(time.Second)
+			continue
+		}
+
+		// All other errors except than the busy text file
 		return nil, pluginErr(err, stdout.Bytes())
 	}
```
saschagrunert, line 60: @dcbw I moved the logic over here. Should we make it configurable somehow?

pkg/invoke/raw_exec.go
```diff
@@ -37,7 +39,24 @@ func (e *RawExec) ExecPlugin(ctx context.Context, pluginPath string, stdinData [
...
+		// If the plugin is currently about to be written, then we wait a
+		// second and try it again
+		if strings.Contains(err.Error(), "text file busy") {
+			time.Sleep(time.Second)
+			continue
+		}
+
+		// All other errors except than the busy text file
 		return nil, e.pluginErr(err, stdout.Bytes(), stderr.Bytes())
 	}
```
saschagrunert, line 61: @dcbw I moved the logic over here, should me make it configurable somehow?

PR #767, opened 2020-05-30
bumped to golang 1.13

ducksecops:
Updated to use golang 1.13. Current version is using 1.12 which went eol on 25th Feb 2020 due to the release of 1.14. 

Signed-off-by: Daniel Sutton <daniel@ducksecops.uk>


bboreham:
This is just obnoxious.
(we should probably delete `release.sh`)

PR #783, opened 2020-08-12
types changes for 1.0.0

dcbw:
Moving to spec 1.0.0 requires more complicated conversions, so
put the converter determination logic in a separate module that
any of the types can call to convert between arbitrary versions.
This is necessary to ensure the types don't have import cycles.

This also implements downconversion, which we never claimed
*not* to support, but didn't implement.

It also verifies that the Result object unmarshalled by
each result type's NewResult() function is actually supported
by the result type. This is a potentially breaking change as
this was not previously done, and for example may now fail
attempts to read a cached PrevResult written by a previous
version.

Signed-off-by: Dan Williams <dcbw@redhat.com>

### DIFF
--- a/libcni/api_test.go
+++ b/libcni/api_test.go
@@ -30,7 +30,7 @@ import (
 	"github.com/containernetworking/cni/libcni"
 	"github.com/containernetworking/cni/pkg/skel"
 	"github.com/containernetworking/cni/pkg/types"
-	"github.com/containernetworking/cni/pkg/types/current"
+	current "github.com/containernetworking/cni/pkg/types/100"
 	noop_debug "github.com/containernetworking/cni/plugins/test/noop/debug"
 
 	. "github.com/onsi/ginkgo"
@@ -59,7 +59,7 @@ func stringInList(s string, list []string) bool {
 	return false
 }
 
-func newPluginInfo(configValue, prevResult string, injectDebugFilePath bool, result string, runtimeConfig map[string]interface{}, capabilities []string) pluginInfo {
+func newPluginInfo(cniVersion, configValue, prevResult string, injectDebugFilePath bool, result string, runtimeConfig map[string]interface{}, capabilities []string) pluginInfo {
 	debugFile, err := ioutil.TempFile("", "cni_debug")
 	Expect(err).NotTo(HaveOccurred())
 	Expect(debugFile.Close()).To(Succeed())
@@ -97,7 +97,7 @@ func newPluginInfo(configValue, prevResult string, injectDebugFilePath bool, res
 	err = json.Unmarshal([]byte(config), &newConfig)
 	Expect(err).NotTo(HaveOccurred())
 	newConfig["name"] = "some-list"
-	newConfig["cniVersion"] = current.ImplementedSpecVersion
+	newConfig["cniVersion"] = cniVersion
 
 	// Only include standard runtime config and capability args that this plugin advertises
 	newRuntimeConfig := make(map[string]interface{})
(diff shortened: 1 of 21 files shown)


dcbw:
@containernetworking/cni-maintainers no longer WIP, please review. Thanks!

pkg/types/convert/convert.go
```diff
@@ -0,0 +1,88 @@
...
+	c := findConverter(fromVersion, toVersion)
+	if c == nil {
+		return nil, fmt.Errorf("no converter for CNI result version %s to %s",
+			fromVersion, toVersion)
+	}
+	return c.convertFn(from, toVersion)
+}
+
+// Register registers a CNI Result converter. SHOULD NOT BE CALLED
+// EXCEPT FROM CNI ITSELF.
```
bboreham, line 73: Should this be enforced by an `internal` package?

pkg/types/020/types.go
```diff
@@ -51,6 +67,39 @@ func GetResult(r types.Result) (*Result, error) {
 	return result, nil
 }
 
+func copyIPConfig(from *IPConfig) *IPConfig {
+	if from == nil {
+		return nil
+	}
+	return from.Copy()
+}
```
bboreham, line 75: I'm not very clear why this function exists.  The `if nil` bit could go in IPConfig.Copy.

libcni/api_test.go
```diff
@@ -541,13 +559,33 @@ var _ = Describe("Invoking plugins", func() {
...
+						var err error
+						netConfig, err = libcni.ConfFromBytes([]byte(`{
+							"type": "noop",
+							"name": "apitest",
+							"cniVersion": "0.4.0",
+							"capabilities": {
+								"portMappings": true,
+								"somethingElse": true,
+								"noCapability": false
+							}
```
bboreham, line 571: How does this change arise?

pkg/types/current/types.go
```diff
@@ -22,27 +22,26 @@ import (
...
-	types020 "github.com/containernetworking/cni/pkg/types/020"
+	types040 "github.com/containernetworking/cni/pkg/types/040"
 	"github.com/containernetworking/cni/pkg/types/convert"
 )
 
-const ImplementedSpecVersion string = "0.4.0"
+const ImplementedSpecVersion string = "1.0.0"
 
-var SupportedVersions = []string{"0.3.0", "0.3.1", ImplementedSpecVersion}
+var SupportedVersions = []string{ImplementedSpecVersion}
```
bboreham, line 31: Is this right?  Looks like we would still support 0.3.x and 0.4.0.

bboreham (commented):
I have a few questions.

pkg/types/convert/convert.go
dcbw: @bboreham good point, I think it should be. Happen to have a doc handy, or is https://docs.google.com/document/d/1e8kOo3r51b2BWtTs_1uADIA5djfXhPT36s6eHVRIvaU/edit what I need to look at?

pkg/types/020/types.go
dcbw: @bboreham can't do that, becuase .Copy() is a struct method, and if you can't do `<nil>.Copy()` right?

libcni/api_test.go
dcbw: Because the default CNI JSON created in BeforeEach() is for 1.0.0 (current.ImplementedSpecVersion) and here we need to use a 0.4.0 JSON.

pkg/types/current/types.go
dcbw: No, because 1.0.0 doesn't have the "Version" field in the IPConfig struct and perhaps we'll make other changes before final.

pkg/types/020/types.go
bboreham: I haven't tried it, but I can't currently see why not.  In Go, `nil` can have a type, but I think all the calls to this one would be statically dispatched.

pkg/types/current/types.go
dcbw: Follow-up: Bryan was really asking why the plugins export this at all; there are only two uses in the reference plugins and those are for error-checking and can be trivially replaced.

bboreham (approved):
LGTM!

dcbw:
@squeed @jellonek @mccv1r0 @bboreham updated, I think this is final now

bboreham (approved):
Let's go ahead with this and see if anything shakes out when we link with plugins and runtimes.

PR #800, opened 2020-12-16
types: unexport SupportedVersions

dcbw:
SupportedVersions isn't really needed outside the types themselves.

@squeed @mccv1r0 @bboreham @matthewdupre

### DIFF
--- a/pkg/types/020/types.go
+++ b/pkg/types/020/types.go
@@ -27,12 +27,15 @@ import (
 
 const ImplementedSpecVersion string = "0.2.0"
 
-var SupportedVersions = []string{"", "0.1.0", ImplementedSpecVersion}
+var supportedVersions = []string{"", "0.1.0", ImplementedSpecVersion}
 
 // Register converters for all versions less than the implemented spec version
 func init() {
-	convert.Register("0.1.0", []string{ImplementedSpecVersion}, convertFrom010)
-	convert.Register(ImplementedSpecVersion, []string{"0.1.0"}, convertTo010)
+	convert.RegisterConverter("0.1.0", []string{ImplementedSpecVersion}, convertFrom010)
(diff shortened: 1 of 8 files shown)


dcbw:
Plugins change looks like this for 3 of the plugins:

```
diff --git a/plugins/ipam/host-local/backend/allocator/config.go b/plugins/ipam/host-local/backend/allocator/config.go
index c8cb2a7460521..936b09799782d 100644
--- a/plugins/ipam/host-local/backend/allocator/config.go
+++ b/plugins/ipam/host-local/backend/allocator/config.go
@@ -20,7 +20,7 @@ import (
        "net"
 
        "github.com/containernetworking/cni/pkg/types"
-       "github.com/containernetworking/cni/pkg/types/020"
+       "github.com/containernetworking/cni/pkg/version"
 )
 
 // The top-level network config - IPAM plugins are passed the full configuration
@@ -136,10 +136,8 @@ func LoadIPAMConfig(bytes []byte, envArgs string) (*IPAMConfig, string, error) {
 
        // CNI spec 0.2.0 and below supported only one v4 and v6 address
        if numV4 > 1 || numV6 > 1 {
-               for _, v := range types020.SupportedVersions {
-                       if n.CNIVersion == v {
-                               return nil, "", fmt.Errorf("CNI version %v does not support more than 1 address per family", n.CNIVersion)
-                       }
+               if ok, _ := version.GreaterThanOrEqualTo(n.CNIVersion, "0.3.0"); !ok {
+                       return nil, "", fmt.Errorf("CNI version %v does not support more than 1 address per family", n.CNIVersion)
                }
        }
```

bboreham:
I think it would be nice to send an email on the CNI mailing list outlining these sorts of changes that are required by plugin writers.

mccv1r0 (commented):
lgtm

squeed:
Agreed, it looks like the plugins don't need to use this field.

bboreham (approved):

PR #802, opened 2020-12-21
Log out the plugin name on ADD/DEL failure.

fasaxc:
Makes it easier to diagnose conf lists when one plugin fails.


mars1024:
ptal at the failed unit tests

fasaxc:
@mars1024 done, thanks.

bboreham (approved):
LGTM - we need to announce in the release notes that the type signature of these errors has changed, but arguably anyone calling this code should be ready to deal with that.

PR #809, opened 2021-01-29
cni:support for mips64le architecture.

houfangdong:
Increase support for mips64le architecture, because golang language supports cross-compilation and supports the construction of mips64le cpu architecture, which makes it more convenient for users to use cni.
@dcbw
@eyakubovich

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ env:
   - TARGET=arm64
   - TARGET=ppc64le
   - TARGET=s390x
-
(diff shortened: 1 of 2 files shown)


houfangdong:
hello，@steveeJ， @rosenhouse，@dcbw
Please review my code, thanks.

dcbw:
@houfangdong we just switched to GitHub actions, would you be able to adjust your changes for that? Thanks!

PR #813, opened 2021-02-10
types/040: add testcase for <= 0.2.0 Result requirement of one or more IPs

dcbw:
The 0.2.0 spec strongly suggests via lack of the "(optional)" tag that
a Result must contain one or two IP addresses. We already throw an
error for that case when down-converting a 0.3.0+ Result to <= 0.2.0,
but we didn't have a testcase for it.


bboreham (approved):
lgtm

PR #831, opened 2021-04-12
Allow MTU to be specified on Route

mccv1r0:
Allow MTU to be specified on Route

Signed-off-by: Michael Cambria <mcambria@redhat.com>

### DIFF
--- a/pkg/types/040/types.go
+++ b/pkg/types/040/types.go
@@ -304,3 +304,49 @@ func (c *IPConfig) UnmarshalJSON(data []byte) error {
 	c.Gateway = ipc.Gateway
 	return nil
 }
+
+type Route struct {
+	Dst net.IPNet
+	GW  net.IP
+}
+
+func (r *Route) String() string {
+	return fmt.Sprintf("%+v", *r)
+}
+
+func (r *Route) Copy() *Route {
+	if r == nil {
+		return nil
+	}
+
+	return &Route{
+		Dst: r.Dst,
+		GW:  r.GW,
+	}
+}
+
+// JSON (un)marshallable types
+type route struct {
+	Dst net.IPNet `json:"dst"`
+	GW  net.IP    `json:"gw,omitempty"`
+}
+
+func (r *Route) UnmarshalJSON(data []byte) error {
+	rt := route{}
+	if err := json.Unmarshal(data, &rt); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


bboreham (approved):

dcbw:
@mccv1r0 can you add the new field to SPEC.md? Also we'll need to move this struct and associated functions into pkg/types/100/ with MTU, and then into 020 and 040 without the MTU bits.

warrengunn48 (approved):

pkg/types/100/types.go
```diff
@@ -305,3 +305,54 @@ func (c *IPConfig) UnmarshalJSON(data []byte) error {
 	c.Gateway = ipc.Gateway
 	return nil
 }
+
+type Route struct {
+	Dst net.IPNet
+	GW  net.IP
+	MTU int
```
mars1024, line 312: Is it better to use int pointer here since MTU is optional?

pkg/types/100/types.go
```diff
@@ -305,3 +305,54 @@ func (c *IPConfig) UnmarshalJSON(data []byte) error {
...
+		return err
+	}
+
+	r.Dst = net.IPNet(rt.Dst)
+	r.GW = rt.GW
+	r.MTU = rt.MTU
+	return nil
+}
+
+func (r Route) MarshalJSON() ([]byte, error) {
```
mars1024, line 350: ditto

pkg/types/040/types.go
```diff
@@ -304,3 +304,49 @@ func (c *IPConfig) UnmarshalJSON(data []byte) error {
...
+	if err := json.Unmarshal(data, &rt); err != nil {
+		return err
+	}
+
+	r.Dst = net.IPNet(rt.Dst)
+	r.GW = rt.GW
+	return nil
+}
+
+func (r Route) MarshalJSON() ([]byte, error) {
```
mars1024, line 345: ditto

dcbw:
@mccv1r0 need to remove `Route` from pkg/types/types.go and that'll give you a cascade of compile issues. Which basically means you need to change each version's Route to it's own rather than types.Route, eg for 020:

```
type IPConfig struct {
	IP      net.IPNet
	Gateway net.IP
-	Routes  []types.Route
+	Routes  []Route
}
```

squeed:
closing this for now - @mccv1r0, feel free to re-open this if we should pick this back up.

PR #841, opened 2021-05-06
Correct the right sequence number

wangchenglong01:
correct the right sequence number


mars1024 (changes requested):
sorry, the error code is not sequential

wangchenglong01:
> sorry, the error code is not sequential

@mars1024 Thanks, I got it wrong. I will close the pr

PR #842, opened 2021-05-21
Remove Bryan Boreham as maintainer

bboreham:
I am moving on from Weaveworks, and will not be working in the container networking domain.


mars1024:
Sorry to hear this! Should we introduce a honorary maintainer list? cc @containernetworking/cni-maintainers

PR #843, opened 2021-05-26
chore: support both value type and pointer type in LoadArgs

mars1024:
During the implementation of [https://github.com/containernetworking/plugins/pull/630](https://github.com/containernetworking/plugins/pull/630), I found that field with pointer type `*net.IP` can not be parsed by `LoadArgs`, but this is working on `json.Unmarshal`.

This PR is trying to support both value type and pointer type which have `encoding.TextUnmarshaler` interface in `LoadArgs`.

### DIFF
--- a/pkg/types/args_test.go
+++ b/pkg/types/args_test.go
@@ -15,6 +15,7 @@
 package types_test
 
 import (
+	"net"
 	"reflect"
 
 	. "github.com/containernetworking/cni/pkg/types"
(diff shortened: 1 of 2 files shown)


pkg/types/args_test.go
```diff
@@ -129,4 +130,33 @@ var _ = Describe("LoadArgs", func() {
...
 	})
+
+	Context("When loading known arguments", func() {
+		It("should succeed if argument is marshallable value type", func() {
+			conf := struct {
+				IP net.IP
+				CommonArgs
+			}{}
+			err := LoadArgs("IP=10.0.0.0", &conf)
+			Expect(err).NotTo(HaveOccurred())
```
dcbw, line 141: @mars1024 should we check that conf.IP.String() == "10.0.0.0" here?

squeed:
Hey @mars1024, what is the use case for this? I'm a bit confused, given that slices in Go are already pointers.

mars1024:
> Hey @mars1024, what is the use case for this? I'm a bit confused, given that slices in Go are already pointers.

OK, `net.IP` and `*net.IP` may not be a good example, let's move to the new type [IP](https://github.com/containernetworking/plugins/blob/78702e9d8a2cdb6931fab433f1d1e6789162954f/pkg/ip/ip.go#L25) which was just introduced.

What I really want is make that `ip.IP` and `*ip.IP` can both be parsed by `LoadArgs` in [https://github.com/containernetworking/plugins/blob/af26bab50051249662642c00cd40d1aac69b2398/plugins/ipam/host-local/backend/allocator/config.go#L60](https://github.com/containernetworking/plugins/blob/af26bab50051249662642c00cd40d1aac69b2398/plugins/ipam/host-local/backend/allocator/config.go#L60), just like `json.Unmarshal` can parse in [https://github.com/containernetworking/plugins/blob/af26bab50051249662642c00cd40d1aac69b2398/plugins/ipam/host-local/backend/allocator/config.go#L64](https://github.com/containernetworking/plugins/blob/af26bab50051249662642c00cd40d1aac69b2398/plugins/ipam/host-local/backend/allocator/config.go#L64).

I think this is a small but useful improvement, user can use pointer type or value type more flexibly.

PR #847, opened 2021-06-23
types: ensure empty CNIVersion always creates/converts to 0.1.0

dcbw:


pkg/types/internal/create.go
```diff
@@ -44,6 +44,9 @@ func findCreator(version string) *creator {
 // Create creates a CNI Result using the given JSON, or an error if the creation
 // could not be performed
 func Create(version string, bytes []byte) (types.Result, error) {
+	if version == "" {
```
mars1024, line 47: As code in [https://github.com/containernetworking/cni/blob/b4e078af7a216436b2f4ddbd70434e8f3e23b550/pkg/types/020/types.go#L28-L39](https://github.com/containernetworking/cni/blob/b4e078af7a216436b2f4ddbd70434e8f3e23b550/pkg/types/020/types.go#L28-L39),
empty string has been registered to creators in L38, so no need to translate it to "0.1.0" here.

And IMO, if `"" --> "0.1.0"` is a convention in creating or converting, it's better to register *once*, but not translate it *everywhere*.

dcbw: @mars1024 good point, but the pkg/type/create Create() function takes a bare version that may not be fixed up; that's a previously existing function to create a CNI result using the specified version. I could put the "" -> 0.1.0 logic there, but I'm afraid of putting the logic too far away from where it's actually needed.

squeed:
Seems reasonable. @mars1024 , want to look again?

mars1024 (approved):
/lgtm

mars1024:
> Seems reasonable. @mars1024 , want to look again?

Also look good to me, approved already :)

PR #849, opened 2021-07-14
Define the differences between the networking stack

jayunit100:


SPEC.md
```diff
@@ -186,7 +186,12 @@ Plugins may define additional fields that they accept and may generate an error
 
 The CNI protocol is based on execution of binaries invoked by the container runtime. CNI defines the protocol between the plugin binary and the runtime.
 
-A CNI plugin is responsible for configuring a container's network interface in some manner. Plugins fall in to two broad categories:
+Note that the CNI API is interpretted in widely different ways by plugins, by design.  That is, plugins *do not operate at the same layer of the networking stack* but they do all implement some portion of a CNI API calls underlying functionality.
```
squeed, line 189:
```suggestion
A CNI plugin is responsible for configuring a container's network interface in some manner. The meaning of this is left undefined by design.  That is, plugins *may provide very different functionality*, but they all speak the CNI protocol.

Plugins fall in to three broad categories:
- foo
- bar
* "Delegated" plugins, which provide "helper" functionality such as IPAM, but are executed by plugins, rather than runtimes.
```

jayunit100:
@squeed once your back at work :)

jayunit100:
talking more in the CNI maintainers group today with @dcbw it feels like I should really make this a diagram.
ill think about how to do that today

jayunit100:
Closing this. Haven't been engaged in this group for a while but maybe someone else will take it over

PR #851, opened 2021-07-18
docs: correct the extension name of the configuration file

DrmagicE:


mars1024 (changes requested):
Actually `LoadConfList` used by `cnitool` will search files with extension `.conflist` firstly then fall through to `.conf` and `.json`, so this should be clear cut in documents.

DrmagicE:
Oh, I see... I was putting a list configuration into a  file with `.json` extension and then I got the "error parsing configuration: missing 'type'" error message. 
Now I realize the `.json` or `.conf` extension is for single plugin, and `.conflist` is for plugin chain.  I think it would be more clear if we mention this in the document.

DrmagicE:
@mars1024 Hi, I have revised the doc to add more details. Would you please take a look？

dcbw:
Thanks!

PR #864, opened 2021-10-25
skel: remove superfluous err nil check in (*dispatcher).pluginMain

tklauser:
There is no need to check err against nil, just return it.


tklauser:
@mars1024 @mccv1r0 can this be merged given both of you approved?

PR #865, opened 2021-10-31
[exec-plugins]: support plugin lists

pkit:
add ability to run configurations with plugin lists/chaning
see SPEC >= 0.3.1


dcbw:
/lgtm
thanks!

PR #869, opened 2021-11-19
introduce hybridnet to thrid-party plugins

mars1024:
link to [hybridnet](https://github.com/alibaba/hybridnet)


matthewdupre (commented):
What does this mean, from the perspective of someone looking to compare CNI plugins?  I think it might be clearer for people if they knew that containers could be attached to either the underlay or the overlay.

mars1024:
> What does this mean, from the perspective of someone looking to compare CNI plugins? I think it might be clearer for people if they knew that containers could be attached to either the underlay or the overlay.

Yes, from the perspective of usage, one container can only choose to be attached to underlay or overlay, but `hybrid-lay` here means overlay pods and underlay pods will have complete bidirectional network connectivity on node level in one cluster, even among multi clusters. This seems a creative feature compared to other 3rd party plugins.

matthewdupre:
Could you add some of that to the comment please @mars1024 (admittedly it's a challenge to balance with brevity)?  People should be able to get an idea for whether they'd be interested just from reading the list.

mars1024:
@matthewdupre Updated the comment, more words but more comprehensible :)

mars1024:
friendly ping @matthewdupre @dcbw , any comments on this?

mars1024:
> @mars1024 could we reword slightly to:
> 
> `a CNI plugin designed for hybrid clouds which provides both overlay and underlay networking for containers in one or more clusters. Overlay and underlay containers can run on the same node and have cluster-wide bidirectional network connectivity.`

@dcbw  Of course and I've already reworded this. Thanks!

PR #873, opened 2022-01-03
Spec: Container runtime shall tear down namespaces

### LINKED ISSUE #872: Container namespace cleanup / teardown is not documented in SPEC
The current spec at https://github.com/andreaskaris/cni/blob/master/SPEC.md#lifecycle--ordering does not explicitly state that the container runtime is responsible for the teardown of network namespaces (it does state though that the runtime must create network namespaces):
~~~
Lifecycle & Ordering

    The container runtime must create a new network namespace for the container before invoking any plugins.
    The container runtime must not invoke parallel operations for the same container, but is allowed to invoke parallel operations for different containers. This includes across multiple attachments.
    Plugins must handle being executed concurrently across different containers. If necessary, they must implement locking on shared resources (e.g. IPAM databases).
    The container runtime must ensure that add is eventually followed by a corresponding delete. The only exception is in the event of catastrophic failure, such as node loss. A delete must still be executed even if the add fails.
    delete may be followed by additional deletes.
    The network configuration should not change between add and delete.
    The network configuration should not change between attachments.
~~~

It would be good to add that the runtime is responsible for network namespace teardown in the above paragraph.

andreaskaris:
Explicitly state that the container runtime, not the plugin, shall tear
down network namespaces.

Fixes #872

Signed-off-by: Andreas Karis <ak.karis@gmail.com>

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -337,6 +337,7 @@ The operation of a network configuration on a container is called an _attachment
 - _delete_ may be followed by additional _deletes_.
 - The network configuration should not change between _add_ and _delete_.
 - The network configuration should not change between _attachments_.
+- The container runtime is responsible for cleanup of the container's network namespace.
 
 ### Attachment Parameters
 While a network configuration should not change between _attachments_, there are certain parameters supplied by the container runtime that are per-attachment. They are:


mccv1r0 (approved):

mars1024 (commented):
IMO, the items of `Lifecycle & Ordering` are more relative to some constraints of container runtime (lifecycle) and network configuration (ordering), not really a list of whole lifecycle.

andreaskaris:
@mars1024  Is there a better place to explicitly state that the CNI is not responsible for container namespace creation nor container namespace tear down? I was looking for it over the documentation and spec; maybe I missed it, but I couldn't find it. IMO, it should be stated somewhere. 

Background: I ran into some issue lately where some namespaces were not torn down, I wanted to link to a piece of documentation that states that the CNI plugin is not responsible for that.

squeed:
lgtm, thanks!

PR #880, opened 2022-02-22
Fix incorrect pointer inputs to `json.Unmarshal`

howardjohn:
See https://github.com/howardjohn/go-unmarshal-double-pointer for more
info on why this is not safe and how this is detected.


matthewdupre (approved):
Thanks!

PR #890, opened 2022-03-31
fix 714-plugin add netns validation reinforcement

Poor12:
fix-714: https://github.com/containernetworking/plugins/issues/714

### DIFF
--- a/pkg/ns/ns_windows.go
+++ b/pkg/ns/ns_windows.go
@@ -1,0 +1,21 @@
+// Copyright 2022 CNI authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ns
+
+import "github.com/containernetworking/cni/pkg/types"
+
+func CheckNetNS(nsPath string) (bool, *types.Error) {
+	return false, nil
+}
--- a/pkg/skel/skel.go
+++ b/pkg/skel/skel.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/containernetworking/cni/pkg/ns"
 	"github.com/containernetworking/cni/pkg/types"
 	"github.com/containernetworking/cni/pkg/utils"
 	"github.com/containernetworking/cni/pkg/version"
@@ -34,12 +35,13 @@ import (
 // CmdArgs captures all the arguments passed in to the plugin
 // via both env vars and stdin
 type CmdArgs struct {
-	ContainerID string
-	Netns       string
-	IfName      string
-	Args        string
-	Path        string
-	StdinData   []byte
+	ContainerID   string
+	Netns         string
+	IfName        string
+	Args          string
+	Path          string
+	NetnsOverride string
+	StdinData     []byte
 }
 
 type dispatcher struct {
@@ -55,7 +57,7 @@ type dispatcher struct {
 type reqForCmdEntry map[string]bool
 
 func (t *dispatcher) getCmdArgsFromEnv() (string, *CmdArgs, *types.Error) {
-	var cmd, contID, netns, ifName, args, path string
+	var cmd, contID, netns, ifName, args, path, netnsOverride string
 
 	vars := []struct {
 		name      string
@@ -116,6 +118,15 @@ func (t *dispatcher) getCmdArgsFromEnv() (string, *CmdArgs, *types.Error) {
 				"DEL":   true,
 			},
 		},
+		{
+			"CNI_NETNS_OVERRIDE",
+			&netnsOverride,
+			reqForCmdEntry{
+				"ADD":   false,
+				"CHECK": false,
+				"DEL":   false,
+			},
+		},
 	}
 
 	argsMissing := make([]string, 0)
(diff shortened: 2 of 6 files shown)


Poor12:
Is something like this? Need your review. :) @squeed

pkg/skel/skel.go
```diff
@@ -190,6 +203,29 @@ func validateConfig(jsonBytes []byte) *types.Error {
...
+func getCurrentThreadNetNSPath() string {
+	// /proc/self/ns/net returns the namespace of the main thread, not
+	// of whatever thread this goroutine is running on.  Make sure we
+	// use the thread's net namespace since the thread is switching around
+	return fmt.Sprintf("/proc/%d/task/%d/ns/net", os.Getpid(), unix.Gettid())
+}
+
+func checkNetNS(nsPath string) (bool, *types.Error) {
+	ns, err := netns.GetFromPath(nsPath)
+	if err != nil {
```
squeed, line 215: We need to support CNI DEL for empty nsPath as well as already-deleted nsPath. So, paradoxically, I think we can return `true` here if `err != nil`.

pkg/skel/skel.go
```diff
@@ -241,6 +277,17 @@ func (t *dispatcher) pluginMain(cmdAdd, cmdCheck, cmdDel func(_ *CmdArgs) error,
 		return types.NewError(types.ErrIncompatibleCNIVersion, "plugin version does not allow CHECK", "")
 	case "DEL":
 		err = t.checkVersionAndCall(cmdArgs, versionInfo, cmdDel)
+		if err != nil {
+			return err
+		}
+		if cmdArgs.NetnsOverride != "TRUE" {
```
dcbw, line 283: @Poor12 could we make this a case-insensitive comparison for "true", and also add a check for "1"?

pkg/skel/skel.go
```diff
@@ -241,6 +277,17 @@ func (t *dispatcher) pluginMain(cmdAdd, cmdCheck, cmdDel func(_ *CmdArgs) error,
 		return types.NewError(types.ErrIncompatibleCNIVersion, "plugin version does not allow CHECK", "")
 	case "DEL":
 		err = t.checkVersionAndCall(cmdArgs, versionInfo, cmdDel)
+		if err != nil {
```
squeed, line 280: Should we check this for ADD as well? I don't see why not...

pkg/skel/skel.go
Poor12: Do you mean we just pass the check and let plugins deal with empty plugins as well as already-deleted nsPath？

pkg/skel/skel.go
Poor12: make it.

pkg/skel/skel.go
Poor12: Add check for ADD.

squeed:
@Poor12 surprise: this doesn't build on Windows :). We'll need to stub out the function with build flags.

Poor12:
> @Poor12 surprise: this doesn't build on Windows :). We'll need to stub out the function with build flags.

I am not familiar with this. Could you do me a favor？Thanks a lot.

dcbw:
> I am not familiar with this. Could you do me a favor？Thanks a lot.

```
 # github.com/vishvananda/netns
Error: C:\Users\runneradmin\go\pkg\mod\github.com\vishvananda\netns@v0.0.0-20210104183010-2eb08e3e575f\netns.go:28:13: undefined: unix.Stat_t
Error: C:\Users\runneradmin\go\pkg\mod\github.com\vishvananda\netns@v0.0.0-20210104183010-2eb08e3e575f\netns.go:29:12: undefined: unix.Fstat
Error: C:\Users\runneradmin\go\pkg\mod\github.com\vishvananda\netns@v0.0.0-20210104183010-2eb08e3e575f\netns.go:32:12: undefined: unix.Fstat
Error: C:\Users\runneradmin\go\pkg\mod\github.com\vishvananda\netns@v0.0.0-20210104183010-2eb08e3e575f\netns.go:43:8: undefined: unix.Stat_t
Error: C:\Users\runneradmin\go\pkg\mod\github.com\vishvananda\netns@v0.0.0-20210104183010-2eb08e3e575f\netns.go:44:12: undefined: unix.Fstat
Error: C:\Users\runneradmin\go\pkg\mod\github.com\vishvananda\netns@v0.0.0-20210104183010-2eb08e3e575f\netns.go:56:8: undefined: unix.Stat_t
Error: C:\Users\runneradmin\go\pkg\mod\github.com\vishvananda\netns@v0.0.0-20210104183010-2eb08e3e575f\netns.go:57:12: undefined: unix.Fstat
Error: C:\Users\runneradmin\go\pkg\mod\github.com\vishvananda\netns@v0.0.0-20210104183010-2eb08e3e575f\netns.go:71:12: undefined: unix.Close
```

That just means that because the netlink library and Unix stuff doesn't work on Windows, you need to move that code to a different file. Here's what you do.

1. Make a new file called netns_linux.go and put getCurrentNS(), getCurrentThreadNetNSPath(), and checkNetNS() into that.
2. Make a new file called netns_windows.go and stub out the checkNetNS() function there like:
```
func checkNetNS(nsPath string) (bool, *types.Error) {
    return false, nil
}
```

Files with _<platform> at the end are automatically only built on the platform in question. It's magic :) That way Linux builds will use the full implementation, and the Windows builds will use the stub implementation that just returns OK. @MikeZappa87 says that Windows doesn't have a good way (that we know of) to get the root compartment, so let's just ignore this on Windows for now.

Poor12:
Thanks for informing this. I have not seen _windows files before. And Happy Labor Day :) @dcbw

pkg/ns/ns_windows.go
```diff
@@ -0,0 +1,19 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ns
+
+func CheckNetNS(nsPath string) (bool, *types.Error) {
```
dcbw, line 17: @Poor12 you need to import `"github.com/containernetworking/cni/pkg/types"` here to get types.Error.

Poor12: Done

Poor12:
Excuse me, several months ago I started a issue about https://github.com/containernetworking/plugins/issues/714 and have proposed a tentative solution but got very few replies lately.  Actually our productive environment has had this problem for a while and we really hope to fix this problem in the community in order to avoid the problem of upstream and downstream inconsistencies caused by our private modification.

But during several months waiting for a response from the community, I have several confusion about this problem.
1. Do we still need to fix this problem？Or we don't think it's a problem？
2. Is there a more reasonable solution for this problem？

cc @squeed @dcbw @mars1024 @matthewdupre @mccv1r0 @MikeZappa87 @jellonek
Really sorry if disturbing you.

PR #894, opened 2022-05-03
cnitool: address golint error

fujitatomoya:
exported const EnvCNIPath should have comment (or a comment on this block) or be unexported

Signed-off-by: Tomoya Fujita <Tomoya.Fujita@sony.com>

### DIFF
--- a/cnitool/cnitool.go
+++ b/cnitool/cnitool.go
@@ -26,6 +26,7 @@ import (
 	"github.com/containernetworking/cni/libcni"
 )
 
+// Protocol parameters are passed to the plugins via OS environment variables.
 const (
 	EnvCNIPath        = "CNI_PATH"
 	EnvNetDir         = "NETCONFPATH"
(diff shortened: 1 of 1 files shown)


cnitool/cnitool.go
```diff
@@ -59,7 +60,6 @@ func parseArgs(args string) ([][2]string, error) {
 func main() {
 	if len(os.Args) < 4 {
 		usage()
-		return
```
fujitatomoya, line 62: @dcbw thanks for quick response! this will never be called since `usage` calls `os.Exit(1)`? do you want me to remove this change cz it does not align the commit description? just checking on what you mean, thanks!

fujitatomoya: @dcbw friendly ping.

dcbw: Nah, that's fine.

fujitatomoya:
@dcbw thanks for iterating!

Issue #895, opened 2022-05-11
converting plugin result to wrong config version results in regression 


mikebrow:
https://github.com/weaveworks/weave/issues/3936
https://github.com/containerd/containerd/issues/6921

@fuweid drafted a prototype fix:
https://github.com/fuweid/containerd/commit/8c5607ca4edae3bfd21a15f7d3f9db2d17fee0f0

@squeed @MikeZappa87

fuweid:
@squeed For sure. I will file the pr! Thanks!

mikebrow:
nod.. puts us in a bind will need both cni lib and weave to fix.. weave fix so we can work fine from containerd 1.6.0-1.6.4 and here in cni so 1.6.5 can work with the version of weave that does not reply with config version in the result..

squeed:
@mikebrow Shouldn't this work around Weave without them needing to fix their buggy behavior?

squeed:
oh nevermind, I just parsed what you said. I'll see if I can bug anyone at weave.

mikebrow:
yes it will.. and we'll put it in our r.next service stream right away after test and cni release tag and contained/go-cni brings in the new cni and containerd/containerd brings in the new go-cni and cherry pick to service for 1.6.5 ... 
Sigh :-)

mikebrow:
@mrunalp heads up another spin on cni lib coming to resolve weave issue with containerd and cri-o cheers..

hswong3i:
Doesn't https://github.com/weaveworks/weave/pull/3939 solve this issue?

dcbw:
This is valid. But https://github.com/fuweid/containerd/commit/8c5607ca4edae3bfd21a15f7d3f9db2d17fee0f0 is not quite right.

fuweid:
> can the behavior be to back-fill the result version to the config iff the result doesn't have a version specified?

Yes.

> While it's helpful to have libcni do this, it encourages plugins to be lazy and not conform to the spec's requirement of setting the Result CNIVersion correctly.



The spec said that the cniVersion of result is the same version supplied on input https://github.com/containernetworking/cni/blob/main/SPEC.md#L480.
Based on that, I applied to request version to override the result, just in case that the plugin returns the wrong version. 

Maybe out of topic. Is it possible to remove the version in result? The request always said what version it wants...

PR #896, opened 2022-05-11
invoke: if Result CNIVersion is empty use netconf CNIVersion

dcbw:
The CNI Spec requires plugins to return a CNIVersion in the Result
that is the same as the CNIVersion of the incoming CNI config.

Empty CNIVersions are specified to map to 0.1.0 (since the first
CNI spec didn't have versioning) and libcni handles that automatically.

Unfortunately some versions of Weave don't do that and depend on
a libcni <= 0.8 quirk that used the netconf version if the result
version was empty. This is technically a libcni regression, though
the plugin itself is violating the specification.

Thus for backwards compatibility assume that if the Result
CNIVersion is empty, the netconf CNIVersion should be used as
the Result version.

Fixes: https://github.com/containernetworking/cni/issues/895

### DIFF
--- a/pkg/invoke/exec.go
+++ b/pkg/invoke/exec.go
@@ -16,6 +16,7 @@ package invoke
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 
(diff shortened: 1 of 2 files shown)


dcbw:
This needs testcases still...

fuweid (approved):
LGTM(nb)

Thanks!

fuweid:
ping @squeed @MikeZappa87 ~

MikeZappa87:
@dcbw I won’t be able to join the meeting today. What is the work remaining for this? Lmk if you need help, I’m sure someone on the containerd might be able to help to get this to completion :-)

squeed:
lgtm, feel free to merge on green.

MikeZappa87 (approved):

PR #897, opened 2022-05-16
Allow MTU to be specified on Route

muff1nman:
This is a pick-up of the initial effort to add mtu as done here: https://github.com/containernetworking/cni/pull/831


muff1nman:
cc @mccv1r0

mccv1r0:
Needs rebase

muff1nman:
Rebased, no modifications made.

pkg/types/100/types_test.go
```diff
@@ -321,4 +326,54 @@ var _ = Describe("Current types operations", func() {
     "address": "10.1.2.3/24"
 }`))
 	})
+
+	Context("when setting custom routes with mtu 0", func() {
```
mccv1r0, line 330: We should have a test for when mtu is actually set as well.  e.g. set to e.g. 1400 (adjust variable names to taste.)

```
+
+       Context("when setting custom routes with mtu 1400", func() {
+               notzero := 1400
+
+               exampleMTU := current.Route{
+                       Dst: net.IPNet{
+                               IP:   net.ParseIP("1.2.3.0"),
+                               Mask: net.CIDRMask(24, 32),
+                       },
+                       GW:  net.ParseIP("1.2.3.1"),
+                       MTU: &notzero,
+               }
+
+               It("marshals to JSON equivalent to an omitted mtu ", func() {
+                       jsonBytes, err := json.Marshal(&exampleMTU)
+                       Expect(err).NotTo(HaveOccurred())
+                       Expect(jsonBytes).To(MatchJSON(`{ "dst": "1.2.3.0/24", "gw": "1.2.3.1", "mtu": 1400 }`))
+               })
+       })

```

muff1nman: There were other tests that had covered this path (`testResult`), but I've added it to make the tests in this function complete.

mccv1r0:
Have you tested your changes with `cnitool` or something similar using currently shipping plugins?

I checked out this PR and tried a quick test (emphasis on quick) pointing plugins to use your cni repo changes.  I get compile errors when trying to build existing plugs that expect:  ```Routes  []types.Route```  not  ```Routes  []Route```

e.g. 

```
$ ./build_linux.sh 
Building plugins 
  bandwidth
# github.com/containernetworking/plugins/pkg/ip
pkg/ip/utils_linux.go:79:44: undefined: types.Route
```

```
# github.com/containernetworking/plugins/plugins/meta/sbr
plugins/meta/sbr/main.go:208:57: undefined: types.Route
``` 

```
# github.com/containernetworking/plugins/plugins/main/bridge
plugins/main/bridge/bridge.go:176:7: undefined: types.Route
```

pkg/types/040/types.go
```diff
@@ -304,3 +310,49 @@ func (c *IPConfig) UnmarshalJSON(data []byte) error {
 	c.Gateway = ipc.Gateway
 	return nil
 }
+
+type Route struct {
```
dcbw, line 314: @muff1nman Let's acutally just use the 020 implementation since they are the same. 040 types import the 020 so it's fine to do that.

squeed:
Unfortunately this change breaks Semver promises (as @mccv1r0 correctly observes) -- not of the spec, but of the go library version. Namely, this removes existing types / fields. We would have to tag this as v2.0.0 in Git, which we're not going to do.

So, the best bet is probably to make MTU a `*uint` on the Route type. I can't think of any other way to solve this.

pkg/types/040/types.go
dcbw: Nevermind... do what Casey suggests with the `MTU *uint`

PR #903, opened 2022-06-24
Add definition for attachment

MikeZappa87:
I noticed we used attachment a few times in the spec and wanted to add it terminology. At a minimum if this is not added we should correct three to four.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -64,11 +64,12 @@ Released versions of the spec are available as Git tags.
 
 This document proposes a generic plugin-based networking solution for application containers on Linux, the _Container Networking Interface_, or _CNI_.
 
-For the purposes of this proposal, we define three terms very specifically:
+For the purposes of this proposal, we define five terms very specifically:
 - _container_ is a network isolation domain, though the actual isolation technology is not defined by the specification. This could be a [network namespace][namespaces] or a virtual machine, for example.
 - _network_ refers to a group of endpoints that are uniquely addressable that can communicate amongst each other. This could be either an individual container (as specified above), a machine, or some other network device (e.g. a router). Containers can be conceptually _added to_ or _removed from_ one or more networks.
 - _runtime_ is the program responsible for executing CNI plugins.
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -64,11 +64,12 @@ Released versions of the spec are available as Git tags.
 
 This document proposes a generic plugin-based networking solution for application containers on Linux, the _Container Networking Interface_, or _CNI_.
 
-For the purposes of this proposal, we define three terms very specifically:
+For the purposes of this proposal, we define five terms very specifically:
 - _container_ is a network isolation domain, though the actual isolation technology is not defined by the specification. This could be a [network namespace][namespaces] or a virtual machine, for example.
 - _network_ refers to a group of endpoints that are uniquely addressable that can communicate amongst each other. This could be either an individual container (as specified above), a machine, or some other network device (e.g. a router). Containers can be conceptually _added to_ or _removed from_ one or more networks.
 - _runtime_ is the program responsible for executing CNI plugins.
 - _plugin_ is a program that applies a specified network configuration.
+- _attachment is the medium that connects the container to network.
```
MikeZappa87, line 72: Need an extra _ on the end

dcbw, line 72: @MikeZappa87 How about: "_attachment_ is the operation of applying a network configuration to a container, which results in an endpoint."

and then another one:

"_endpoint_ is a uniquely addressable, container-scoped connection to a single network."

squeed:
We also have this wording:

> The operation of a network configuration on a container is called an _attachment_. An attachment may be uniquely identified by the `(CNI_CONTAINERID, CNI_IFNAME)` tuple.

Maybe this should be hoisted up to the definitions?

henry118:
Yup I like the idea of reusing the existing words. 

How about this?

"_attachment_ is the operation of applying a network configuration to a container. It is uniquely identified by the `(CNI_CONTAINERID, CNI_IFNAME)` tuple."

Issue #877, opened 2022-01-20
Questions on usage of Check cmd in cni spec >= v0.4.0


cgchinmay:
Couple of questions
1. How to use the disableCheck flag, what is the correct datatype
In [master](https://github.com/containernetworking/cni/blob/master/SPEC.md#configuration-format)
Here its boolean
```
disableCheck (boolean): Either true or false. If disableCheck is true, runtimes must not call CHECK for this network configuration list. 
```

In [spec-v0.4.0 spec](https://github.com/containernetworking/cni/blob/spec-v0.4.0/SPEC.md#network-configuration-lists)
Here its string
```
disableCheck (string): Either true or false. If disableCheck is true, runtimes must not call CHECK for this network configuration list. 
```
I tried following, but nodes wont reach ready state 
```
"disableCheck": "true" 
```
So I modified it to boolean, and that worked
```
"disableCheck": true
```

2. Is CHECK command implementation optional ?  It appears to be since you have a way to disable it but hasn't been explicitly called out.

MikeZappa87:
@cgchinmay right now containerd or cri-o do not implement the CHECK command. I am working on the side to get it implemented into containerd.

gojoy:
> @cgchinmay right now containerd or cri-o do not implement the CHECK command. I am working on the side to get it implemented into containerd.

when containerd can implement CHECK command, is there an issue or PR for it
thanks @MikeZappa87

dcbw:
> I tried following, but nodes wont reach ready state
> 
> ```
> "disableCheck": "true" 
> ```

That formatting is not correct JSON; `disableCheck` is a boolean value so it must be either `true` or `false` without quotes.

PR #915, opened 2022-10-17
libcni: handle string-type disableCheck values

dcbw:
CNI spec 0.4x (and possibly earlier) mistakenly specified disableCheck as a string type, while 1.0 cleaned that up to be a real boolean.

We should be nice and interpret the string value.

Fixes: https://github.com/containernetworking/cni/issues/877


mars1024 (approved):
/lgtm

mccv1r0 (approved):

PR #926, opened 2022-11-14
cnitool: deduplicate environment variable definitions

UiP9AV6Y:
create a reusable package for environment variable to be used by `cnitool` and the other library code.

Signed-off-by: Gordon Bleux <33967640+UiP9AV6Y@users.noreply.github.com>


pkg/env/env.go
```diff
@@ -0,0 +1,95 @@
...
+	VarCNIContainerId   = "CNI_CONTAINERID"
+	VarCNIIfname        = "CNI_IFNAME"
+	VarCNINetNs         = "CNI_NETNS"
+	VarCNINetNsOverride = "CNI_NETNS_OVERRIDE"
+	VarCNIPath          = "CNI_PATH"
+	VarNetDir           = "NETCONFPATH"
+
+	DefaultCapabilityArgs = ""
+	DefaultCNIArgs        = ""
+	DefaultCNIIfname      = "eth0"
```
squeed, line 36: This should probably be the empty string

pkg/env/env.go
```diff
@@ -0,0 +1,95 @@
...
+	DefaultNetDir         = "/etc/cni/net.d"
```
squeed, line 37: This should be the empty string too.

pkg/env/env.go
```diff
@@ -0,0 +1,95 @@
...
+	if val == "" {
+		return fallbackValue
+	}
+
+	return val
+}
+
+// GetNetDir gets the network configuration
+// directory setting from the environment
+func GetNetDir() string {
```
squeed, line 63: Should this be in a central package? It's only used by cnitool, right?

squeed:
This looks basically reasonable, but any of the environment variables that are `cnitool`-specific should probably not be in `pkg`. That's just because we don't want to make any public library commitments in that regard.

pkg/skel/skel.go
```diff
@@ -65,77 +66,54 @@ func (t *dispatcher) getCmdArgsFromEnv() (string, *CmdArgs, *types.Error) {
...
 		{
-			"CNI_COMMAND",
+			env.VarCNICommand,
 			&cmd,
-			reqForCmdEntry{
-				"ADD":   true,
-				"CHECK": true,
-				"DEL":   true,
-			},
+			newReqForCmdEntry(true, true, true),
```
squeed, line 71: nit: I prefer the literal here, if you don't mind.

UiP9AV6Y:
i have moved all `cnitool` specific default values back into
the cmd package to prevent having too many *empty defaults*
(the default has to be defined by the consumer when using 
`GetVal()`)

furthermore i have moved the CNI Args parsing logic into
the library API as this is something covered by the
[SPEC](https://github.com/containernetworking/cni/blob/e629d106b4e02030d0f7794758b0f0438b6a78dd/SPEC.md?plain=1#L209)

PR #942, opened 2023-01-09
cnitool: fix default CNI directory on Windows and add extra checks.

aznashwan:


aznashwan:
Note that only validation logic which may introduce extra constraints is [this new `filepath.IsAbs()` check](https://github.com/containernetworking/cni/pull/942/files#diff-f53c197bf509416a6a733371e65d3ed7e66ca6e136b8eaeedad8646192f2d0c7R68-R70), but I can't think of a usecase where one might want to use a relative path instead of an absolute path.

aznashwan:
> Note that the only new validation logic which may introduce extra constraints is [this new filepath.IsAbs() check](https://github.com/containernetworking/cni/pull/942/files#diff-f53c197bf509416a6a733371e65d3ed7e66ca6e136b8eaeedad8646192f2d0c7R68-R70)

FWIW I have switched to using `filepath.Abs()` to bring it in line with the rest of the code.

I've also compared the pre and post-patch behavior on Linux:

```bash
### Pre-patch:
# Relative paths work:
$ NETCONFPATH=../../../../etc/cni/net.d/ ./cnitool check nonexistent-net /var/run/netns/testing 
no net configuration with name "nonexistent-net" in ../../../../etc/cni/net.d/  # correct 

# Setting a nonexistent directory will only claim there are no net configurations:
$ NETCONFPATH=/nonexistent/path ./cnitool check nonexistent-net /var/run/netns/testing 
no net configurations found in /nonexistent/path

# Setting a nonexistent relative directory will also only claim there are no net configurations:
$ NETCONFPATH=./nonexistent/relative/path ./cnitool check nonexistent-net /var/run/netns/testing 
no net configurations found in ./nonexistent/relative/path

### Post-patch:
# Relative paths still work:
$ NETCONFPATH=../../../../etc/cni/net.d/ ./cnitool check nonexistent-net /var/run/netns/testing 
no net configuration with name "nonexistent-net" in /etc/cni/net.d  # correct

# Setting a nonexistent directory will error out with a clear message:
$ NETCONFPATH=/nonexistent/path ./cnitool check nonexistent-net /var/run/netns/testing 
the provided CNI config path ($NETCONFPATH) does not exist: "/nonexistent/path"

# Setting a nonexistent relative directory will also error out with a clear message:
$ NETCONFPATH=./nonexistent/relative/path ./cnitool check nonexistent-net /var/run/netns/testing 
the provided CNI config path ($NETCONFPATH) does not exist: "/home/nashu/cni/cnitool/nonexistent/relative/path"
```

cnitool/constants_windows.go
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+const (
+	// NOTE: this is the most reasonable default as most CNI setups on Windows
+	// will have been made using the helper script in the containerd repo.
+	// https://github.com/containerd/containerd/blob/main/script/setup/install-cni-windows
+	DefaultNetDir = "C:\\Program Files\\containerd\\cni\\conf"
```
dcbw, line 21: @aznashwan this seems containerd specific, but CNI is used by more than just containerd. If each CRI implementation has its own default directory, should this be a list of default dirs that get tried in sequence?

MikeZappa87:
@aznashwan is this still required?

aznashwan:
@MikeZappa87 
> is this still required?

While I have't worked with Windows CNIs in almost 2 whole years, this patch still looks relevant for anyone wanting to run `cnitool` on Windows (not sure how relevant that is nowadays however).

I've just rebased the PR FWIW.

cnitool/constants_windows.go
aznashwan: Sorry for the huge delay, I must've somehow missed this notification when it happened.

> seems containerd specific

Correct.

> CNI is used by more than just containerd

Correct, but generally not really Windows (at least when I was involved in CNI porting efforts on Windows a whole year ago).

Back then, the vast majority of k8s/docker setups on Windows directly installed containerd from their build scripts in the main `containerd/containerd` repo, so it seemed like the most reasonable choice...

I'm not sure whether there is a "standardised way" to install k8s on Windows nowadays, but I'm willing to bet it uses containerd with this default.

Plus at the end of the day this path is configurable via an env var, so the `C:\Program Files\containerd\cni\conf` default isn't preventing anyone from using `cnitool`, and it's certainly a hell of a lot saner of a default on Windows than `/etc/cni/net.d` 🤣

PR #957, opened 2023-01-30
Add an iptables rule to drop packets that conntrack considers invalid.

mccv1r0:
Add an iptables rule to drop packets that conntrack considers invalid.

One rule for the life of CNI vs one for ipMasq, portmap etc.

This is an alternative to adding an unique container IP specific rules every cniADD for ipMasq, portmap and then removing when cniDel is called.

Fixes [plugins 817]( https://github.com/containernetworking/plugins/pull/817)

Signed-off-by: Michael Cambria <mcambria@redhat.com>


mccv1r0:
The existing conntrack entry should stick around for 2*MSL to address any in flight packets from needing such a rule.

PR #961, opened 2023-02-07
Docs: remove obsolete information about building release artifacts

### LINKED ISSUE #960: Not able to find the ./build.sh
The script release.sh is giving me the following error when it runs:

```
+ apk --no-cache add bash tar
fetch https://dl-cdn.alpinelinux.org/alpine/v3.13/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.13/community/x86_64/APKINDEX.tar.gz
(1/6) Installing ncurses-terminfo-base (6.2_p20210109-r1)
(2/6) Installing ncurses-libs (6.2_p20210109-r1)
(3/6) Installing readline (8.1.0-r0)
(4/6) Installing bash (5.1.16-r0)
Executing bash-5.1.16-r0.post-install
(5/6) Installing libacl (2.2.53-r0)
(6/6) Installing tar (1.34-r0)
Executing busybox-1.32.1-r2.trigger
OK: 9 MiB in 21 packages
+ cd /opt/src
+ umask 0022
+ rm -f 'bin/*'
+ CGO_ENABLED=0 GOARCH=amd64 ./build.sh -a --ldflags '-extldflags "-static"'
/bin/sh: ./build.sh: not found
```

Also, the [golang:1.14-alpine](https://github.com/containernetworking/cni/blob/main/scripts/release.sh#L17) used is too old. Can we upgrade the image here as well.
The [plugins](https://github.com/containernetworking/plugins/blob/main/build_linux.sh) dir contains the build_linux script.
Am I missing something here.

squeed:
Now that plugins live in a different repository, we don't need these old files. We don't actually build any release binaries.

Signed-off-by: Casey Callendrello <c1@caseyc.net>

Fixes #960


henry118 (approved):
LGTM

PR #963, opened 2023-02-17
Porting over getCachedNetworkInfo routine from cri-o

henry118:
Port over `getCachedNetworkInfo` routine from cri-o/ocicni to CNI library.

`getCachedNetworkInfo` routine is currently implemented in [cri-o/ocicni](https://github.com/cri-o/ocicni/blob/master/pkg/ocicni/ocicni.go#L576-L642).

I think it would be useful to have this routine in CNI library and shared with other consumers. One use case is the network GC feature that I am working on for containerd, this routine can be used to provide a list of all network attachments known to CNI during GC.

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -26,6 +26,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/containernetworking/cni/pkg/invoke"
@@ -78,6 +79,16 @@ type NetworkConfigList struct {
 	Bytes        []byte
 }
 
+type NetworkAttachment struct {
+	ContainerID    string
+	Network        string
+	IfName         string
+	Config         []byte
+	NetNS          string
+	CniArgs        [][2]string
+	CapabilityArgs map[string]interface{}
+}
+
 type CNI interface {
 	AddNetworkList(ctx context.Context, net *NetworkConfigList, rt *RuntimeConf) (types.Result, error)
 	CheckNetworkList(ctx context.Context, net *NetworkConfigList, rt *RuntimeConf) error
@@ -93,6 +104,8 @@ type CNI interface {
 
 	ValidateNetworkList(ctx context.Context, net *NetworkConfigList) ([]string, error)
 	ValidateNetwork(ctx context.Context, net *NetworkConfig) ([]string, error)
+
+	GetCachedAttachments(containerID string) ([]*NetworkAttachment, error)
 }
 
 type CNIConfig struct {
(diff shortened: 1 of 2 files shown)


henry118:
@dcbw PTAL if this makes sense. thanks!

libcni/api.go
```diff
@@ -391,6 +406,65 @@ func (c *CNIConfig) GetNetworkCachedConfig(net *NetworkConfig, rt *RuntimeConf)
...
+	entries, err := os.ReadDir(dirPath)
+	if err != nil {
+		return nil, err
+	}
+
+	fileNames := make([]string, 0, len(entries))
+	for _, e := range entries {
+		fileNames = append(fileNames, e.Name())
+	}
+	sort.Strings(fileNames)
```
mccv1r0, line 422: Why sort?

libcni/api.go
```diff
@@ -196,6 +209,7 @@ type cachedInfo struct {
 	Config         []byte                 `json:"config"`
 	IfName         string                 `json:"ifName"`
 	NetworkName    string                 `json:"networkName"`
+	NetNS          string                 `json:"netns,omitempty"`
```
s1061123, line 212: This PR introduces `NetNS` in cachedInfo as new field. I'm a bit curious that this code works with previous version CNI cache, that does not have `NetNS` field. Could you please double check that?

henry118: Good catch. I've created a small program to validate the json marshaling for this change [here](https://go.dev/play/p/z71ZJfB29a_1). It seems fine.

libcni/api.go
henry118: Good question. I didn't author this code but ported from [ocicni](https://github.com/cri-o/ocicni/blob/master/pkg/ocicni/ocicni.go#L592). And I had same question when reading the code.

I guess the intention was making it easier for the caller to process the results in logical groups, because the cache file names are in the format of "network-container-ifname" as defined [here](https://github.com/containernetworking/cni/blob/6626820a52c9d460651c805d81f6307cf2c4b865/libcni/api.go#L223).

libcni/api.go
```diff
@@ -391,6 +406,65 @@ func (c *CNIConfig) GetNetworkCachedConfig(net *NetworkConfig, rt *RuntimeConf)
 	return c.getCachedConfig(net.Network.Name, rt)
 }
 
+// GetCachedAttachments returns a list of network attachments from the cache.
+// The returned list will be filtered by the containerID if the value is not empty.
+func (c *CNIConfig) GetCachedAttachments(containerID string) ([]*NetworkAttachment, error) {
```
MikeZappa87, line 411: Does it make sense to have a collection of NetworkAttachments or some type that has a parent with multiple attachment type children? Some of the properties won't change. The netns property could be in the parent type and then the properties that are attachment specific in the children?

henry118: Yes it does make sense in higher level packages like `go-cni` and `ocicni`, each of them have similar constructs like `Namespace` and `PodNetwork`, both concern at the level of container/pod/ns, and have container id, netns, and list of attachments as properties. `libcni` works at lower level and mostly concerns about individual attachments. IMO in the current model it's better to keep the higher level constructs in those packages.

J90307 (approved):

PR #981, opened 2023-03-20
SPEC: add wording for GC

squeed:
This is an initial draft of the wording for the GC verb.

TODO:
- [x] Need to declare that concurrent operations are not allowed (e.g. no ADD and GC at the same time)
- [x] Describe the execution flow (akin to deletion)
- [ ] Final proofreading

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -17,20 +17,23 @@
       - [`DEL`: Remove container from network, or un-apply modifications](#del-remove-container-from-network-or-un-apply-modifications)
       - [`CHECK`: Check container's networking is as expected](#check-check-containers-networking-is-as-expected)
       - [`VERSION`: probe plugin version support](#version-probe-plugin-version-support)
+      - [`GC`: Clean up any stale resources](#gc-clean-up-any-stale-resources)
   - [Section 3: Execution of Network Configurations](#section-3-execution-of-network-configurations)
     - [Lifecycle \& Ordering](#lifecycle--ordering)
     - [Attachment Parameters](#attachment-parameters)
     - [Adding an attachment](#adding-an-attachment)
     - [Deleting an attachment](#deleting-an-attachment)
     - [Checking an attachment](#checking-an-attachment)
-    - [Deriving execution configuration from plugin configuration](#deriving-execution-configuration-from-plugin-configuration)
+    - [Garbage-collecting a network](#garbage-collecting-a-network)
+    - [Deriving request configuration from plugin configuration](#deriving-request-configuration-from-plugin-configuration)
       - [Deriving `runtimeConfig`](#deriving-runtimeconfig)
   - [Section 4: Plugin Delegation](#section-4-plugin-delegation)
     - [Delegated Plugin protocol](#delegated-plugin-protocol)
     - [Delegated plugin execution procedure](#delegated-plugin-execution-procedure)
   - [Section 5: Result Types](#section-5-result-types)
-    - [Success](#success)
+    - [ADD Success](#add-success)
       - [Delegated plugins (IPAM)](#delegated-plugins-ipam)
+    - [VERSION Success](#version-success)
     - [Error](#error)
     - [Version](#version-1)
   - [Appendix: Examples](#appendix-examples)
@@ -202,7 +205,7 @@ The runtime must execute the plugin in the runtime's networking domain. (For mos
 
 Protocol parameters are passed to the plugins via OS environment variables.
 
-- `CNI_COMMAND`: indicates the desired operation; `ADD`, `DEL`, `CHECK`, or `VERSION`.
+- `CNI_COMMAND`: indicates the desired operation; `ADD`, `DEL`, `CHECK`, `GC`, or `VERSION`.
 - `CNI_CONTAINERID`: Container ID. A unique plaintext identifier for a container, allocated by the runtime. Must not be empty.  Must start with an alphanumeric character, optionally followed by any combination of one or more alphanumeric characters, underscore (), dot (.) or hyphen (-).
 - `CNI_NETNS`: A reference to the container's "isolation domain". If using network namespaces, then a path to the network namespace (e.g. `/run/netns/[nsname]`)
 - `CNI_IFNAME`: Name of the interface to create inside the container; if the plugin is unable to use this interface name it must return an error.
@@ -214,7 +217,7 @@ A plugin must exit with a return code of 0 on success, and non-zero on failure. 
 
 ### CNI operations
 
-CNI defines 4 operations: `ADD`, `DEL`, `CHECK`, and `VERSION`. These are passed to the plugin via the `CNI_COMMAND` environment variable.
+CNI defines 5 operations: `ADD`, `DEL`, `CHECK`, `GC`, and `VERSION`. These are passed to the plugin via the `CNI_COMMAND` environment variable.
 
 #### `ADD`: Add container to network, or apply modifications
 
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -321,6 +322,33 @@ A json-serialized object, with the following key:
 Required environment parameters:
 - `CNI_COMMAND`
 
+#### `GC`: Clean up any stale resources
+
+The GC comand provides a way for runtimes to specify the expected set of attachments to a network.
+The network plugin may then remove any resources related to attachments that do not exist in this set.
```
henry118, line 328: What about attachments existing in this set but not found by plugins? Do we need to define the behavior? e.g. error out vs. silently ignored?

SPEC.md
```diff
@@ -321,6 +322,33 @@ A json-serialized object, with the following key:
...
+
+Resources may, for example, include:
+- IPAM reservations
+- Firewall rules
+
+A plugin SHOULD remove as many stale resources as possible. For example, a plugin should remove any IPAM reservations associated with attachments not in the provided list. The plugin MAY assume that the isolation domain (e.g. network namespace) has been deleted, and thus any resources (e.g. network interfaces) therein have been removed.
```
henry118, line 334: What should be the return value of GC? Can the runtime assume GC always succeeds? Otherwise how does runtime know which attachments are removed and which are not?

SPEC.md
```diff
@@ -321,6 +322,33 @@ A json-serialized object, with the following key:
...
+
+Plugins MUST, additionally, forward any GC calls to delegated plugins they are configured to use (see section 4).
+
+The runtime MUST NOT use GC as a substitute for DEL. Plugins may be unable to clean up some resources from GC that they would have been able to clean up from DEL.
+
+** Input:**
+
+The runtime must provide a JSON-serialized plugin configuration object (defined below) on standard in. It contains an additional key;
+
+- `cni.dev/attachments` (array of objects): The list of **still valid** attachments to this network:
```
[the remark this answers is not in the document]
squeed: Good point; attachments are implied by the spec but it would be good to make them clear.

SPEC.md
squeed: Hmm, interesting idea. We could certainly consider returning those as "potentially failed attachments". However, I think CHECK already covers this use case reasonably well, and I don't want to push too much of a burden on individual plugins.

SPEC.md
squeed: Good question. What sort of information would be useful for the runtime? For example, we could have the plugin return whether or not any resources were cleaned up. If we wanted to indicate potentially-misconfigured attachments, we could return them here (though I'm not yet convinced this is useful).

As for returning error, the semantics are the same as with DEL: don't return an error if resources are missing, however, if an error condition prevents successful GC, expose that fact to the runtime.

henry118: Yeah I am not sure how much the runtime can do in cases of GC failures, apart from logging an error message. But I do think it would be useful for the runtimes to be aware of GC failures. So if someone were to analyze resource leaks, the log message could be a good indicator. 

So we might want to define "what is a GC failure" for CNI plugin. e.g.
* Plugin's GC procedure cannot be started; or
* An attachment failed to be deleted; or
* The resources of an attachment are missing (this case is probably safe to ignore)

When CNI finishes invoking GC on all "leaked" attachments, it should return an error to runtime if "any" of the attachment fails GC.

SPEC.md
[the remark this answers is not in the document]
squeed: Funny that you mentioned this -- https://github.com/containernetworking/cni/pull/903 has been lying around :-)

SPEC.md
squeed: OK, we worked on wording for this, PTAL.

henry118: Thanks.

henry118 (commented):
LGTM in general. Just some nits, PTAL

SPEC.md
```diff
@@ -321,6 +323,63 @@ A json-serialized object, with the following key:
...
+    - `containerID` (string): the value of CNI_CONTAINERID as provided during the CNI ADD operation
+    - `ifname` (string): the value of CNI_IFNAME as provided during the CNI ADD operation
+
+Required environment parameters:
+- `CNI_COMMAND`
+- `CNI_PATH`#### `GC`: Clean up any stale resources
+
+The GC comand provides a way for runtimes to specify the expected set of attachments to a network.
```
tamilmani1989, line 353: how periodically GC command should be executed by runtimes?

SPEC.md
```diff
@@ -321,6 +323,63 @@ A json-serialized object, with the following key:
...
+The network plugin may then remove any resources related to attachments that do not exist in this set.
```
tamilmani1989, line 354: can runtime issue ADD/DEL calls before GC call returns?

SPEC.md
```diff
@@ -321,6 +323,63 @@ A json-serialized object, with the following key:
...
+
+Resources may, for example, include:
+- IPAM reservations
+- Firewall rules
+
+A plugin SHOULD remove as many stale resources as possible. For example, a plugin should remove any IPAM reservations associated with attachments not in the provided list. The plugin MAY assume that the isolation domain (e.g. network namespace) has been deleted, and thus any resources (e.g. network interfaces) therein have been removed.
+
+Plugins should generally complete a `GC` action without error. If an error is encountered, a plugin should continue; removing as many resources as possible, and report the errors back to the runtime.
```
tamilmani1989, line 362: does runtime have any timeout for GC call to return from plugin or uses same timeout for other calls?

SPEC.md
squeed: No! we'll call this out in the spec.

SPEC.md
squeed: A good question, but timeouts are not part of the CNI spec. The runtime can cancel if it likes.

MikeZappa87:
@henry118 we made some updates

SPEC.md
squeed: That is runtime-dependent; it may execute it 0 times if it likes.

SPEC.md
```diff
@@ -332,6 +366,7 @@ The operation of a network configuration on a container is called an _attachment
 
 - The container runtime must create a new network namespace for the container before invoking any plugins.
 - The container runtime must not invoke parallel operations for the same container, but is allowed to invoke parallel operations for different containers. This includes across multiple attachments.
+  - **Exception**: The runtime must exclusively execute either _gc_ or _add_ and _delete_. The runtime must ensure that no _add_ or _delete_ operations are in progress before executing _gc_, and must wait for _gc_ to complete before issuing new _add_ or _delete_ commands.
```
MikeZappa87, line 369: The verbs are lower case in this block, is that intentional?

SPEC.md
```diff
@@ -321,6 +323,38 @@ A json-serialized object, with the following key:
...
+
+The runtime MUST NOT use GC as a substitute for DEL. Plugins may be unable to clean up some resources from GC that they would have been able to clean up from DEL.
+
+**Input:**
+
+The runtime must provide a JSON-serialized plugin configuration object (defined below) on standard in. It contains an additional key;
+
+- `cni.dev/attachments` (array of objects): The list of **still valid** attachments to this network:
+    - `containerID` (string): the value of CNI_CONTAINERID as provided during the CNI ADD operation
+    - `ifname` (string): the value of CNI_IFNAME as provided during the CNI ADD operation
```
MikeZappa87, line 349: Is the CNI_IFNAME important for GC? I associate that interface inside of the network namespace vs something in the root network namespace.

SPEC.md
```diff
@@ -321,6 +323,38 @@ A json-serialized object, with the following key:
...
+
+Plugins MUST, additionally, forward any GC calls to delegated plugins they are configured to use (see section 4).
+
+The runtime MUST NOT use GC as a substitute for DEL. Plugins may be unable to clean up some resources from GC that they would have been able to clean up from DEL.
+
+**Input:**
+
+The runtime must provide a JSON-serialized plugin configuration object (defined below) on standard in. It contains an additional key;
+
+- `cni.dev/attachments` (array of objects): The list of **still valid** attachments to this network:
```
MikeZappa87, line 347: Should we provide a json block?

SPEC.md
```diff
@@ -389,6 +424,21 @@ For every plugin defined in the `plugins` key of the network configuration,
...
+- There are no attachment parameters (e.g. capability args, CNI_ARGS, etc.)
+- `prevResult` is not supplied
+
+For every plugin defined in the `plugins` key of the network configuration,
+1. Look up the executable specified in the `type` field. If this does not exist, then this is an error.
+2. Derive request configuration from the plugin configuration.
+3. Execute the plugin binary, with `CNI_COMMAND=GC`. Supply the derived configuration via standard in.
+4. If the plugin returns an error, **continue** with execution, returning all errors to the caller.
+
+If all plugins return success, return success to the caller.
```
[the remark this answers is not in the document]
squeed: This has been taken care of.

SPEC.md
squeed: I'm matching the style in this section.

SPEC.md
squeed: This is done.

dcbw:
LGTM

SPEC.md
squeed: Yeah, that matches style in this section.

PR #987, opened 2023-04-20
Define Makefile with golangci-lint

mmorel-35:
As govet and gofmt are now handled by golangci-lint, there is no need to apply them here.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,1 @@
+include mk/lint.mk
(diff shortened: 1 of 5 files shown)


squeed:
While you're at it, do you want to also add a lint.sh for people to run manually? Or just add a Makefile? (can be a separate PR)

Makefile
```diff
@@ -0,0 +1 @@
+include mk/lint.mk
```
squeed, line 1: one minor bikeshed: I like to be able to do `make lint` - could you add that as a make target :-). Thanks!

henry118 (approved):
LGTM

mmorel-35:
@squeed ,
Anything to add here ?

squeed (approved):

Issue #927, opened 2022-11-14
Proposal: feature discovery / version "negotiation" flag-file


squeed:
## The problem
We at the CNI project have long held administrators wholly responsible for understanding and creating their CNI configurations. In other words, there is expected to be a single party who totally manages the contents of the CNI configuration file.

In today's world, this is rarely the case. CNI configurations are often generated by third-party networking providers, and are expected to function in a wide variety of environments and software versions.

A CNI configuration manager might want to know
- What is the highest supported CNI version?
- Are CNI drop-ins supported?

## Specification?
CNI _as a specification_ does not make any statement about configuration management and discovery. Thus whatever is proposed is, currently, outside of the body of the specification and falls in to the murky world of `libcni` convention and inertia. Perhaps that shouldn't be the case. As part of CNI 2.0, we may wish to codify this behavior as well (some kind of discovery is clearly in the cards).

That said, better is the enemy of done, and I'd like to solve this without getting *too far* in the weeds.

## Proposed solution:

`libcni` writes a file in its configuration directory indicating which versions and, possibly, features are supported. Something like 

```jsonc
{
    "apiVersion": "v1",
    "kind": "CNIRuntimeInfo",
    "supportedVersions":["1.5", "1.0"],
    "supportedFeatures": {
        "dropins": false
    }
}
```

The absence of such a file would imply libcni v1.1.

squeed:
This came to head as part of the discussion with STATUS (#859). Specifically, plugins need a way to know if they can expect CNI v1.1 behavior (STATUS), or if they should use pre-v1.1 behavior and write their configuration only when ready.

So far, the proposed solutions are:

1. v1.1 runtimes read from `/etc/cni/net.d/cniv1.1/` as well as `/etc/cni/net.d`
    **Pros**: solves backwards-compat; could flag other features like dropin directories
    **Cons**: complicates runtimes (crio, containerd) that use fsnotify to watch for changes
2. Plugins write an invalid v1.0 but valid v1.1 config file
    **Pros**: Doesn't introduce any new files
    **Cons**: Total hack, relies on specific behavior from runtimes to not ignore invalid files
3. Switch to an entirely new directory
    **Pros**: Easy way to enable new features... once
    **Cons**: Need to update the entire world
4. Add a new filename extension (e.g. `.conflist.1`) to signify a v1.1 plugin (in not-ready state)
    **Pros**: Easy way to roll out this feature, backwards compatible
[...]

PR #1003, opened 2023-06-19
spec: STATUS wording

squeed:
This adds a new verb, STATUS, which indicates whether or not a plugin is ready to accept ADD requests. Runtimes may choose to use the STATUS mechanism to determine if a network is ready.

TODO:
- [ ] Add STATUS subsection in Execution flow (Section 3). This needs to be restructured for GC as well.


squeed:
hey @danwinship, might you have some time to look at this?

SPEC.md
```diff
@@ -310,6 +311,30 @@ Optional environment parameters:
...
+
+#### `STATUS`: Check plugin status
+`STATUS` is a way for a runtime to determine the readiness of a network plugin.
+
+A plugin must exit with a zero (success) return code if the plugin is ready to service ADD requests. If the plugin is not able to service ADD requests, it must exit with a non-zero return code and output an error on standard out (see below).
+
+The following error codes are defined in the context of `STATUS`:
+
+- 50: The plugin is not available (i.e. cannot service `ADD` requests)
+- 51: The plugin is not available, and existing containers in the network may have limited connectivity.
```
squeed, line 323: After discussing this forever in the maintainers' meeting, we're not sure if we should actually include this.

SPEC.md
```diff
@@ -310,6 +311,30 @@ Optional environment parameters:
...
+
+Plugin considerations:
+- Status is purely informational. A plugin MUST NOT rely on `STATUS` being called.
```
danwinship, line 326: That kills the install/upgrade use case; if you don't know for sure that the runtime is going to call `STATUS`, then you can't reliably use `STATUS` to signal plugin unavailability, so you have to fall back to the existing mechanism of not installing your CNI config until after the plugin is fully available.

SPEC.md
```diff
@@ -310,6 +311,30 @@ Optional environment parameters:
...
+- Plugins should always expect other CNI operations (like `ADD`, `DEL`, etc) even if `STATUS` returns an error.`STATUS` does not prevent other runtime requests.
+- If a plugin relies on a delegated plugin (e.g. IPAM) to service `ADD` requests, it must also execute a `STATUS` request to that plugin. If the delegated plugin return an error result, the executing plugin should return an error result.
```
danwinship, line 328:
```suggestion
- If a plugin relies on a delegated plugin (e.g. IPAM) to service `ADD` requests, it must also execute a `STATUS` request to that plugin when it receives a `STATUS` request for itself. If the delegated plugin return an error result, the executing plugin should return an error result.
```

SPEC.md
squeed: There are two issues here -- determining runtime CNI version and readiness racing.

CNI doesn't currently have a notion of, say, node readiness -- that is a Kubernetes concept. This PR provides a signal in the case where the runtime and plugins both care about readiness. However, there could always be runtimes (e.g. Podman) that don't have the notion of readiness, or even a node at all. The spec needs to support that use case.

I also want to strongly discourage STATUS being used as a pseudo-INIT.

That said, plugins that are operating in the broader kubernetes context do need a way to determine which behavior the runtime expects: status via file presence or status via STATUS. My proposal for that was a feature discovery mechanism (#927), and this is a good motivation to finish that proposal as well.

aojea: > if you don't know for sure that the runtime is going to call STATUS, 

we need to provide the mechanisms to solve the problem,  and then take this changes to the container runtimes, as Dan says this does not work if the runtimes does not use it, but the plan is to get this into containerd and crio ...

SPEC.md
aojea: I was suggesting to start small without specific error codes and iterate on feedback and use cases, agree with Dan that is hard to know in advance which code is better suit for what scenario or more important, what to do in each case on the upstream implementations

aojea:
related KEP https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3836-kube-proxy-improved-ingress-connectivity-reliability

SPEC.md
```diff
@@ -310,6 +311,30 @@ Optional environment parameters:
 
 All parameters, with the exception of `CNI_PATH`, must be the same as the corresponding `ADD` for this container.
 
+
+#### `STATUS`: Check plugin status
+`STATUS` is a way for a runtime to determine the readiness of a network plugin.
```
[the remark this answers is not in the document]
dcbw: I like this idea; it's less confusing.

SPEC.md
squeed: Follow-up: we couldn't find a great solution to this problem; plan is to release a small utility library that allows CNI-generating libraries to determine CNI version from the runtime version. This information is released via the kubernetes Node object, for example.

SPEC.md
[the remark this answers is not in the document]
squeed: @s1061123 suggests READY

squeed:
As discussed in the meeting, this is ready to merge.

Issue #1004, opened 2023-06-27
Add MTU as a Route property


aojea:
I think this comment from @joestringer is spot on  https://github.com/cilium/cilium/issues/23711#issuecomment-1557554013

> One possible solution is to teach kubevirt about the route (+MTU) when configuring the devices. By the looks, the [CNI spec doesn't encode this today in its routes](https://github.com/containernetworking/cni/blob/main/pkg/types/types.go#L131), but maybe it should encode MTU in the routes there.

https://github.com/containernetworking/cni/blob/e255525e4da22c886b3562a5fed82ec2112031a5/pkg/types/types.go#L131-L134

For the cases of CNIs that does not use the traditional way of changing the network interface and use routeMTU to handle the discrepancies between interfaces, it will be useful if the plugins will be able to set the mtu directly from the CNI configuration.

aojea:
previous work from @mccv1r0 https://github.com/containernetworking/cni/pull/831

PR #1005, opened 2023-07-04
Add MTU field to route type definition

Joffref:
Fixes: #1004


pkg/types/020/types_test.go
```diff
@@ -79,7 +79,8 @@ func testResult(resultCNIVersion, jsonCNIVersion string) (*types020.Result, stri
         "routes": [
             {
                 "dst": "15.5.6.0/24",
-                "gw": "15.5.6.8"
+                "gw": "15.5.6.8",
+				"mtu": 1024
```
aojea, line 83: indentation looks wrong

pkg/version/legacy_examples/examples.go
```diff
@@ -135,7 +135,8 @@ var ExpectedResult = &types020.Result{
 					IP:   net.ParseIP("0.0.0.0"),
 					Mask: net.CIDRMask(0, 32),
 				},
-				GW: net.ParseIP("10.1.0.1"),
+				GW:  net.ParseIP("10.1.0.1"),
+				MTU: 0,
```
aojea, line 139: does this make sense? to pass 0?

aojea:
have you tested this? it does not look like https://github.com/containernetworking/cni/pull/831

It also required to update Spec.md

/cc @squeed @dcbw

pkg/types/020/types_test.go
```diff
@@ -53,14 +53,14 @@ func testResult(resultCNIVersion, jsonCNIVersion string) (*types020.Result, stri
 			IP:      *ipv4,
 			Gateway: net.ParseIP("1.2.3.1"),
 			Routes: []types.Route{
-				{Dst: *routev4, GW: routegwv4},
+				{Dst: *routev4, GW: routegwv4, MTU: 1024},
```
squeed, line 56: We should probably not add any new fields to v0.2.0 and v0.4.0 -- what do you think, @dcbw ?

squeed:
This looks basically good. We also need to describe this in SPEC.md. Shouldn't be more than a line or two needed.

pkg/types/020/types_test.go
squeed: @Joffref yes please.
(Sorry, I somehow missed your question!)

Joffref: Change made!

pkg/types/020/types_test.go
```diff
@@ -17,13 +17,13 @@ package types020_test
 import (
 	"encoding/json"
 	"fmt"
+	types020 "github.com/containernetworking/cni/pkg/types/020"
```
aojea, line 20: nit this change same unrelated

Joffref:
Added in #1041

Issue #941, opened 2023-01-02
Proposal: version negotiation / cniVersions array


squeed:
In #927 I proposed writing a "flag-file" for runtimes to indicate which versions of the CNI protocol they support.

The reasoning there is the same here - namely, CNI assumes a network crafted by a single administrator, but this is not often the case. Networking vendors may not know which version(s) of CNI they should support, and thus stay at the lowest-common-denominator.

This proposes an alternate way to solve this problem: the cniVersions field. It would allow for the CNI runtime and plugins to select the highest-possible version without any explicit configuration on the part of the administrator.

Example:
```jsonc
{
  "cniVersion": "0.3.1",
  "cniVersions": ["0.3.1", "0.4", "1.1", "1.2"],
  "plugins": {
    // ...
  }
}
```

The spec would be something like

### Version negotiation

If the `cniVersion` field is present, and `cniVersions` is absent, the runtime MUST use the version specified there.

If the `cniVersions` field is present, the runtime MUST use the highest version that it supports.

If both fields are present, then all versions specified should be considered while selecting the protocol version.

When the CNI configuration is transformed to an execution configuration, the `cniVersions` field MUST NOT be set. The `cniVersion` field MUST be set to the version selected above.

PR #1010, opened 2023-07-19
spec, libcni: add cniVersions field in CNI configurations

squeed:
As outlined in #941, it would be useful to CNI vendors to specify that they support multiple versions. This PR provides the spec and libcni implementation of that feature.

Specifically, it allows writing config lists like

```jsonc
{
    "cniVersion": "1.0.0",
    "cniVersions": ["1.0.0", "1.1.0"],
    "plugins": [
        // ...
    ]
}
```
for which libcni would select version 1.1.0 when calling the plugins.


squeed:
In retrospect, I think we do want this. It will make auto-generating CNI configuration file versions easier.

If a cluster is in the midst of changing container runtimes (e.g updating CRI-O), then it may not be possible for an administrator to easily determine the CNI version in use. Allowing for multiple versions simplifies the rollout strategy.

SPEC.md
```diff
@@ -185,6 +188,15 @@ Plugins may define additional fields that they accept and may generate an error
 }
 ```
 
+### Version considerations
+
+CNI runtimes and plugins may support multiple versions of the protocol. To indicate that a particular configuration supports
```
dcbw, line 193: @squeed maybe "CNI runtimes, plugins, and configurations may each support multiple versions of the CNI specification independently. Plugins indicate supported CNI specification versions through the reply to the `VERSION` command, while network configurations may indicate multiple supported CNI specification versions through the `cniVersions` field.

Runtimes are thus able to query both plugins and configurations and select a commonly supported CNI specification version for the CNI request. The runtime MUST select the highest supplied version that it supports. If both `cniVersion` and `cniVersions` are supplied, the runtime MUST select the highest of the two.

libcni/conf.go
```diff
@@ -86,6 +89,45 @@ func ConfListFromBytes(bytes []byte) (*NetworkConfigList, error) {
...
+		if cniVersion != "" {
+			if v, err := semver.NewVersion(cniVersion); err != nil {
+				return nil, fmt.Errorf("error parsing configuration list: invalid cniVersion %s: %w", cniVersion, err)
+			} else if !v.GreaterThan(currentVersion) {
+				// ignore any versions higher than the current implemented spec version
+				vs = append(vs, v)
+			}
+		}
+		sort.Sort(semver.Collection(vs))
+		cniVersion = vs[len(vs)-1].String()
```
dcbw, line 128: @squeed What if `vs` is empty?

dcbw (approved):
LGTM

PR #1011, opened 2023-07-20
pkg/types: explicitly import all versions

squeed:
There was a bug in which some versions didn't register their creation function, thus leaving some types un-decodeable. Fix that by explicitly importing all known versions.


squeed:
I'm not sure how tests didn't catch this. I suspect because we did other things to load these packages, and loading them was (sigh) side-effecting.

dcbw:
LGTM

MikeZappa87 (approved):

Issue #974, opened 2023-03-06
GC Verb for CNI 1.1


MikeZappa87:
We have come to the conclusion that GC should be implemented in the CNI. To do this we need to do the following.

1. Come up with use cases
2. Propose a spec change
3. Implement the change in libcni

aojea:
Use case 1:

I have a chained configuration and one of the plugins at the last of the chain fails, however, the first plugins executed succeeded but don't have any way to revert the change

Use case 2:
Bugs exists, for whatever reason my CNI plugin leaked IPs or my IPAM plugin did, I can detect the leakage and open a bug to fix it, but in the meantime I would like my system to be able to come back to a working state without having to fully restart it

haircommander:
I have interest in this as CRI-O currently keeps track of the pods on a node reboot to be able to call CNI del on all of them. Could this verb be used to mark a node as having rebooted and tell the plugin to invalidate all of the resources there?

(this is all me inferencing based on the name, so apologies if it's not correct)

squeed:
@haircommander that is precisely the intended usage. Although, some care is needed since k8s still needs to do its reconciliation loop and issue a StopPodSandbox for all the now-stale sandboxes.

So, I anticipate this being some kind of periodic reconciliation loop rather than a oneshot on-startup call.

PR #1022, opened 2023-09-15
implement GC keyword for CNI spec 1.1

henry118:
This PR implements the `GC` keyword in `libcni` for CNI spec 1.1

Fixes: #974

### DIFF
--- a/libcni/api_test.go
+++ b/libcni/api_test.go
@@ -1513,16 +1513,71 @@ var _ = Describe("Invoking plugins", func() {
 				_, err := cniConfig.AddNetworkList(ctx, netConfigList, runtimeConfig)
 				Expect(err).NotTo(HaveOccurred())
 
+				By("Issuing a GC with valid networks")
+				gcargs := &libcni.GCArgs{
+					ValidAttachments: []libcni.GCAttachment{{
+						ContainerID: runtimeConfig.ContainerID,
+						IfName:      runtimeConfig.IfName,
+					}},
+				}
+				err = cniConfig.GCNetworkList(ctx, netConfigList, gcargs)
+				Expect(err).NotTo(HaveOccurred())
+
 				By("Issuing a GC with no valid networks")
-				err = cniConfig.GCNetworkList(ctx, netConfigList, libcni.GCArgs{
-					ValidAttachments: []libcni.GCAttachment{},
-				})
+				gcargs.ValidAttachments = nil
+				err = cniConfig.GCNetworkList(ctx, netConfigList, gcargs)
 				Expect(err).NotTo(HaveOccurred())
 
 				commands, err := noop_debug.ReadCommandLog(plugins[0].commandFilePath)
 				Expect(err).NotTo(HaveOccurred())
-				Expect(commands).To(Equal([]string{"ADD", "DEL", "GC"}))
+				Expect(commands).To(HaveLen(4))
 
+				validations := []struct {
+					name string
+					fn   func(entry noop_debug.CmdLogEntry)
+				}{
+					{
+						name: "ADD",
+						fn: func(entry noop_debug.CmdLogEntry) {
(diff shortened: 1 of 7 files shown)


pkg/skel/skel.go
```diff
@@ -319,8 +364,8 @@ func PluginMainWithError(cmdAdd, cmdCheck, cmdDel func(_ *CmdArgs) error, versio
 // as JSON to stdout and call os.Exit(1).
 //
 // To have more control over error handling, use PluginMainWithError() instead.
-func PluginMain(cmdAdd, cmdCheck, cmdDel func(_ *CmdArgs) error, versionInfo version.PluginInfo, about string) {
-	if e := PluginMainWithError(cmdAdd, cmdCheck, cmdDel, versionInfo, about); e != nil {
+func PluginMain(cmdAdd, cmdCheck, cmdDel, cmdGC func(_ *CmdArgs) error, versionInfo version.PluginInfo, about string) {
```
squeed, line 367: Rather than adding more functions here, let's just deprecate the existing functions and leave their signatures as-is.

plugins/test/noop/debug/debug.go
```diff
@@ -70,3 +72,25 @@ func (debug *Debug) WriteDebug(debugFilePath string) error {
 
 	return nil
 }
+
+// WriteCommandLog appends the executed cni command to the record file
+func WriteCommandLog(path string, cmd string) error {
+	fp, err := os.OpenFile(path, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0o655)
```
s1061123, line 78: Why permission is `655`, `rw-r-xr-x`? I'm just curious why 'x' is required to group/others?

libcni/api_test.go
```diff
@@ -1499,6 +1507,32 @@ var _ = Describe("Invoking plugins", func() {
 				Expect(err).To(MatchError("[plugin noop does not support config version \"broken\" plugin noop does not support config version \"broken\" plugin noop does not support config version \"broken\"]"))
 			})
 		})
+		Describe("GCNetworkList", func() {
```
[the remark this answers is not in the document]
squeed: aha, perfect.

pkg/skel/skel.go
```diff
@@ -256,15 +268,15 @@ func (t *dispatcher) pluginMain(cmdAdd, cmdCheck, cmdDel func(_ *CmdArgs) error,
 			if err != nil {
 				return types.NewError(types.ErrDecodingFailure, err.Error(), "")
 			} else if gtet {
-				if err := t.checkVersionAndCall(cmdArgs, versionInfo, cmdCheck); err != nil {
+				if err := t.checkVersionAndCall(cmdArgs, versionInfo, funcs.Check); err != nil {
```
s1061123, line 271: This requires `func.Check` must not be nil (otherwise nil pointer reference happen). Should we add validate all func.{Add,Del,Check,...} in initiliazation (or support 'nil' to mention unsupported)

squeed (changes requested):
Chatted with @dcbw for a bit, and we both agreed we want to see just one more thing: can you somehow expand the debug plugin so we can determine the containerID and ifname that is being passed to the DEL?

After that we can merge this. Thanks!

s1061123:
I'm wondering that plugin's GC might be causes several race conditions: 

For example:
1. plugin A's GC is invoked by container runtime and plugin A starts the GC call
2. During pluing A's GC call, (unfortunately) container runtime invokes plugin A's GC again (oh god...)

In addition, we also have another example:
1. plugin A's GC is invoked by container runtime and plugin A starts the GC call
2. During pluing A's GC call, (unfortunately) container runtime invokes plugin A's Add

To simplify plugin code, both should be blocked (i.e. prevent to invoke), I suppose and it should be implemented common code, in libcni.

What do you think about that?

squeed (approved):

PR #1031, opened 2023-10-23
use replace directive for github.com/containernetworking/cni

mmorel-35:


squeed (approved):

squeed:
good catch

PR #1007, opened 2023-07-10
update NetConf.DNS's json tag to omitempty

cyclinder:
Fixes https://github.com/containernetworking/cni/issues/1006

PR #1035, opened 2023-10-26
Change DNS to pointer in NetConf/Results

s1061123:
This PR changes DNS in NetConf/Results to pointer. This PR contains #1007 and DNS Result change commit.

Plugin repo needs change: https://github.com/containernetworking/plugins/pull/964

### DIFF
--- a/pkg/types/100/types.go
+++ b/pkg/types/100/types.go
@@ -92,7 +92,7 @@ type Result struct {
 	Interfaces []*Interface   `json:"interfaces,omitempty"`
 	IPs        []*IPConfig    `json:"ips,omitempty"`
 	Routes     []*types.Route `json:"routes,omitempty"`
-	DNS        types.DNS      `json:"dns,omitempty"`
+	DNS        *types.DNS     `json:"dns,omitempty"`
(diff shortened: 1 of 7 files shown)


pkg/types/100/types.go
```diff
@@ -92,7 +92,7 @@ type Result struct {
 	Interfaces []*Interface   `json:"interfaces,omitempty"`
 	IPs        []*IPConfig    `json:"ips,omitempty"`
 	Routes     []*types.Route `json:"routes,omitempty"`
-	DNS        types.DNS      `json:"dns,omitempty"`
+	DNS        *types.DNS     `json:"dns,omitempty"`
```
squeed, line 95: Unfortunately, we can't change the API here, since it would break upgrades :-(.

s1061123: OK, I will change the way to keep the structure type in another PRs. thank you for comment!

s1061123: @squeed create another PR for alternative: https://github.com/containernetworking/cni/pull/1039 PTAL?

PR #1037, opened 2023-10-30
fix workflow warnings

mmorel-35:
This is executing checkout first.


squeed (approved):

squeed (commented):
Wait, do we want to do this? I've seen a lot of projects use old versions in go.mod.

mmorel-35:
I removed the go-version-file so only the ordering of action is left.

PR #1039, opened 2023-10-31
Add Marshal function in Result/NetConf to omit empty value

s1061123:
This change fix to avoid empty DNS/IPAM field in NetConf/Result if DNS is empty.

### DIFF
--- a/pkg/types/100/types.go
+++ b/pkg/types/100/types.go
@@ -95,6 +95,29 @@ type Result struct {
 	DNS        types.DNS      `json:"dns,omitempty"`
 }
 
+// Note: DNS should be omit if DNS is empty but default Marshal function
+// will output empty structure hence need to write a Marshal function
+func (r *Result) MarshalJSON() ([]byte, error) {
+	// use type alias to escape recursion for json.Marshal() to MarshalJSON()
+	type fixObjType = Result
+
+	bytes, err := json.Marshal(fixObjType(*r)) //nolint:all
+	if err != nil {
+		return nil, err
+	}
+
+	fixupObj := make(map[string]interface{})
+	if err := json.Unmarshal(bytes, &fixupObj); err != nil {
+		return nil, err
+	}
+
+	if r.DNS.IsEmpty() {
+		delete(fixupObj, "dns")
+	}
+
+	return json.Marshal(fixupObj)
+}
+
 // convertFrom100 does nothing except set the version; the types are the same
 func convertFrom100(from types.Result, toVersion string) (types.Result, error) {
 	fromResult := from.(*Result)
--- a/pkg/types/types.go
+++ b/pkg/types/types.go
@@ -64,14 +64,42 @@ type NetConf struct {
 	Type         string          `json:"type,omitempty"`
 	Capabilities map[string]bool `json:"capabilities,omitempty"`
 	IPAM         IPAM            `json:"ipam,omitempty"`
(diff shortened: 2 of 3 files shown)


s1061123:
This will oversede #1035 and #1007

pkg/types/types.go
```diff
@@ -64,16 +65,86 @@ type NetConf struct {
...
 	Capabilities map[string]bool `json:"capabilities,omitempty"`
 	IPAM         IPAM            `json:"ipam,omitempty"`
-	DNS          DNS             `json:"dns"`
+	DNS          DNS             `json:"dns,omitempty"`
 
 	RawPrevResult map[string]interface{} `json:"prevResult,omitempty"`
 	PrevResult    Result                 `json:"-"`
 }
 
+// Note: DNS should be omit if DNS is empty empty but default Marshal function
```
cyclinder, line 74: Is `empty empty` a nit?

pkg/types/100/types.go
```diff
@@ -95,6 +96,61 @@ type Result struct {
 	DNS        types.DNS      `json:"dns,omitempty"`
 }
 
+// Note: DNS should be omit if DNS is empty empty but default Marshal function
```
cyclinder, line 99: Is `empty empty` a nit?

pkg/types/types.go
s1061123: fixed.

pkg/types/100/types.go
s1061123: fixed.

pkg/types/100/types.go
```diff
@@ -95,6 +96,61 @@ type Result struct {
 	DNS        types.DNS      `json:"dns,omitempty"`
 }
 
+// Note: DNS should be omit if DNS is empty but default Marshal function
+// will output empty structure hence need to write a Marshal function
+func (r *Result) MarshalJSON() ([]byte, error) {
+	var buf bytes.Buffer
```
dcbw, line 102: @s1061123 I think a simpler way to do this would be to marshal to JSON, then unmarshal to a `map[string]interface{}`, check if the original DNS is empty and delete it from the map, then re-marshal the map[string]interface{} to a []byte.

That way we don't have to handle each field individually and do a bunch of string manipulation.

```
    bytes, err := json.Marshal(r)
    if err != nil {
        return nil, err
    }

    fixupObj := make(map[string]interface{})
    if err := json.Unmarshal(bytes, &fixupObj); err != nil {
        return err
    }

    if r.DNS.IsEmpty() {
        delete(interimObj, "dns")
    }

    return json.Marshal(interimObj)
```

s1061123: This makes it simple, but this calls json.Unmarshal()/Marshal(). Hence it is a tradeoff between simple code v.s. performance, I guess. But we could choose simple code in such case, because it does not so 'super good' performance. OK. I will change that.

s1061123: Oh, we cannot use 'json.Marshal' in MarshalJSON() because it makes recursion (i.e. json.Marshal(result) -> it calls MarshalJSON() because golang found 'MarshalJSON()' for result -> MarshalJSON() calls json.Marshal(result) again)...

s1061123: I could resolved by type alias. Changed.

Adawntoremember:
b62753aa2bfa365c1ceaff6f25774a8047c896b5

PR #1041, opened 2023-11-20
Add route attributes - MTU, AdvMSS, Priority

alopintsev:
Extend route attributes to make route selection available based on route priority and enable per-route MTU settings.
It's will be helpfull in multi-CNI environment, and networks, that support jumbo-frame.

Fixes: #1004


alopintsev:
> @alopintsev looks reasonable. However, we will need to update the specification as well, as these are defined types. Thanks!

I've made changes in specification, would you mind to check it?

squeed (approved):

squeed (approved):
Looks good, thanks!

squeed:
FYI @aojea / @Joffref, who had also looked at this.

aojea:
LGTM

Issue #928, opened 2022-11-14
Proposal: drop-in directories


squeed:
As mentioned in #927, the the reality is that network configuration is not managed top-down by an administrator in real-world clusters. Rather, it is maintained by one or more third parties, all of which may need to react according to the existing configuration and/or modify it.

This makes composability awkward, to say the least.

An _extremely common_ use-case is wanting to insert a chained plugin in to an existing configuration. That existing configuration is managed by yet another third (fourth?) party. Currently, this can only be done by reading, editing, and re-writing the configuration file. This is fragile, since the file is nominally "owned" by one network provider, but edited by another.

This proposal is for CNI dropin directories, which `libcni` would read from to scan for additional chained plugins.

### Example

Imagine there is a configuration file, `10-awesome.conflist` with the following contents:

```jsonc
{
    "cniVersion": "1.0",
    "name": "my-awesome-network",
    "plugins": [
        { "type": "awesome" },
        { "type": "chained-1" }
    ]
}
```

Then, libcni would look for a folder named `my-awesome-network`. It would parse any `.conf` files contained in them. Imagine it finds `11-chained.conf` with the following contents:

```jsonc
{
    "type": "chained-2"
}
```

This would cause `libcni` to insert "chained-2" to the *end* of list of chained plugins when parsing `10-awesome.conflist`.

bleggett:
> Currently, this can only be done by reading, editing, and re-writing the configuration file. This is fragile, since the file is nominally "owned" by one network provider, but edited by another.

Yes, and if there are multiple node agents installing plugins that are mutually unaware of each other, they essentially are contending over the same node-level config file, which can and does lead to [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) bugs which render the entire node unusable.

It seems simplest and safest to effectively mimic Unix init system design here, which encountered and solved the same problem years ago using the mechanism proposed by @squeed

bleggett:
Took a stab at a quick spec update to add this in #1052 (looking ahead to 1.2+) - mostly to clarify what a spec-based approach would look like.

Issue #878, opened 2022-01-24
podman, kubernetes and more fight over the CNI config directory


fasaxc:
I'm not sure what the solution is but as more and more apps start to use CNI, I'm seeing them fight over the /etc/cni/net.d directory.  For example, RH CoreOS pre-installs podman, which includes its own CNI in /etc/cni/net.d.  

If you then install kubernetes on the same system, you'd typically want to install a kubernetes focussed CNI (such as Calico or Flannel) but:
* In the window before the k8s CNI is installed, podman's CNI will get used for early-scheduled pods (ugh!)
* After the k8s CNI is installed then podman will try to use it, which is generally unexpected and likely broken.

CRIO also installs its own CNI by default.

In general the problems this causes are strange and confusing for an end user who didn't even know they had podman installed.  They just see broken networking with some pods getting IPs from a different IP range.

I think it'd be good if the CNI spec could address this somehow.  In general, I think it's more common to want a different CNI for each app that's installed rather than to share.

I suppose one option would be to add something to the netconf that says "this config should only be used by kubernetes/podman/whatever".  Then the bootstrap CNIs installed by those apps could include such a stanza.

mccv1r0:
> I think it'd be good if the CNI spec could address this somehow.

The user of CNI could easily specify where their configuration files are located at by setting the NETCONFPATH environment variable.  The binaries themselves can co-exist by each setting CNI_PATH to where they placed their binaries, either the reference binaries from here or their own.

haircommander:
CRI-O and podman support filtering for a network name. I don't think CRI-O will use podman's network configs, and vice versa. We have a [similar issue](https://github.com/cri-o/cri-o/issues/2885) we're looking to fix in 1.24 about not installing a network config by default in the CRI-O installation so other networking providers can be the only one installed (beginning at runtime)

haircommander:
I was wrong, CRI-O does support filtering, but doesn't do so OOTB. If you know the [network config name](https://github.com/containernetworking/cni/blob/master/libcni/api.go#L73) ahead of time you can update your cri-o config to filter for that by setting cni_default_network to the network name

danwinship:
> In general, I think it's more common to want a different CNI for each app that's installed rather than to share.

Yes, exactly this. The CNI spec imagines a world where people install a random container runtime and a random CNI plugin onto a host, and the user wants the runtime to autodetect the plugin and use it. But in my experience, that's not the way that CNI gets used most of the time.

[...]

PR #1052, opened 2023-12-13
RFC - Support safe subdirectory-based plugin conf loading

bleggett:
Designed to address https://github.com/containernetworking/cni/issues/928 (and related issues such as https://github.com/containernetworking/cni/issues/878), largely building off the ideas in #928.

The basic problem is that since plugin config is stored in a centralized, non-locked config file on the node, if multiple node agents install or remove CNI plugins in the same chain, they are both mutating a shared node resource with no locking, which unavoidably leads to unfixable TOCTOU errors.

This resolves that by moving the plugin configs out of the main shared node config, and into subdirectories, Unix init-system style.

~One thing that is TBD is whether we rev the network config object spec for this or not.~

- ~#928 describes a mechanism where the subdir-based loading is dynamic and not explicit in the top-level config.~

- ~This approach revisions the network config object spec to remove the old `plugins` array, more clearly indicating that subdirectories will be used instead, and forcing a hard version bump so it is not ambiguous as to what runtimes support this.~

This adds a new config flag `loadPluginsFromFolder` - if present, for a given named network `bar`, plugin configuration objects will be loaded from `<path-to-bar-network-config-file>/bar/xxx.conf`

If this flag is not present, inlined plugin configs will be handled as they were previously. If this flag is `true` and there are inlined plugin config objects in the primary network config, the `libcni` config loading will combine everything into one plugin list.

Current things to note:

- Impl is open for debate but here to spark debate.
- ~This _will_ rev the config spec from 1.0.0~
- ~Given that we still keep support for pre-1.0.0 spec formats in `libcni`, and this proposal will add yet another post-1.0.0 spec format, and we should support both 1.0.0 and post-1.0.0 formats in `libcni` for transitional purposes, that leaves us with `libcni`'s API surface supporting 3 different versions of the spec, 2 of which would be non-current.~
- I have marked all pre-1.0.0 conf loading functions as `deprecated`.
- Related to that, we have a lot of typenames that no longer accurately represent what they are, in this PR I have renamed them while keeping backwards-compatible aliases (marked as deprecated) around for the time being.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -109,14 +109,16 @@ A network configuration consists of a JSON object with the following keys:
 
 - `cniVersion` (string): [Semantic Version 2.0](https://semver.org) of CNI specification to which this configuration list and all the individual configurations conform. Currently "1.1.0"
 - `cniVersions` (string list): List of all CNI versions which this configuration supports. See [version selection](#version-selection) below.
-- `name` (string): Network name. This should be unique across all network configurations on a host (or other administrative domain).  Must start with an alphanumeric character, optionally followed by any combination of one or more alphanumeric characters, underscore, dot (.) or hyphen (-).
+- `name` (string): Network name. This should be unique across all network configurations on a host (or other administrative domain).  Must start with an alphanumeric character, optionally followed by any combination of one or more alphanumeric characters, underscore, dot (.) or hyphen (-). Must not contain characters disallowed in file paths.
 - `disableCheck` (boolean): Either `true` or `false`.  If `disableCheck` is `true`, runtimes must not call `CHECK` for this network configuration list.  This allows an administrator to prevent `CHECK`ing where a combination of plugins is known to return spurious errors.
 - `disableGC` (boolean): Either `true` or `false`.  If `disableGC` is `true`, runtimes must not call `GC` for this network configuration list.  This allows an administrator to prevent `GC`ing when it is known that garbage collection may have undesired effects (e.g. shared configuration between multiple runtimes).
-- `plugins` (list): A list of CNI plugins and their configuration, which is a list of plugin configuration objects.
+- `loadOnlyInlinedPlugins` (boolean): Either `true` or `false`. If `false` (default), indicates [plugin configuration objects](#plugin-configuration-objects) can be aggregated from multiple sources. Any valid plugin configuration objects aggregated from other sources must be appended to the final list of `plugins` for that network name. If set to `true`, indicates that valid plugin configuration objects aggregated from sources other than the main network configuration will be ignored. If `plugins` is not present in the network configuration, `loadOnlyInlinedPlugins` cannot be set to `true`.
+- `plugins` (list): A list of inlined [plugin configuration objects](#plugin-configuration-objects). If this key is populated with inlined plugin objects, and `loadOnlyInlinedPlugins` is true, the final set of plugins for a network must consist of all the plugin objects in this list, merged with all the plugins loaded from the sibling folder with the same name as the network.
 
 #### Plugin configuration objects:
-Plugin configuration objects may contain additional fields than the ones defined here.
-The runtime MUST pass through these fields, unchanged, to the plugin, as defined in section 3.
+Runtimes may aggregate plugin configuration objects from multiple sources, and must unambiguously associate each loaded plugin configuration object with a single, valid network configuration. All aggregated plugin configuration objects must be validated, and each plugin with a valid configuration object must be invoked.
+
+Plugin configuration objects may contain additional fields beyond the ones defined here. The runtime MUST pass through these fields, unchanged, to the invoked plugin, as defined in section 3.
 
 **Required keys:**
 - `type` (string): Matches the name of the CNI plugin binary on disk. Must not contain characters disallowed in file paths for the system (e.g. / or \\).
@@ -147,6 +149,7 @@ Plugins that consume any of these configuration keys should respect their intend
 Plugins may define additional fields that they accept and may generate an error if called with unknown fields. Runtimes must preserve unknown fields in plugin configuration objects when transforming for execution.
 
 #### Example configuration
+The following is an example JSON representation of a network configuration `dbnet` with three plugin configurations (`bridge`, `tuning`, and `portmap`).
 ```jsonc
 {
   "cniVersion": "1.1.0",
@@ -158,7 +161,7 @@ Plugins may define additional fields that they accept and may generate an error 
       // plugin specific parameters
       "bridge": "cni0",
       "keyA": ["some more", "plugin specific", "configuration"],
-      
+
       "ipam": {
         "type": "host-local",
         // ipam specific
@@ -374,8 +377,6 @@ Resources may, for example, include:
 - Firewall rules
 
 A plugin SHOULD remove as many stale resources as possible. For example, a plugin should remove any IPAM reservations associated with attachments not in the provided list. The plugin MAY assume that the isolation domain (e.g. network namespace) has been deleted, and thus any resources (e.g. network interfaces) therein have been removed.
-
-Garbage collection is a per-network operation. If a plugin manages resources shared across multiple networks, it must only remove stale resources known to belong to the network provided in the `GC `action.
 
 Plugins should generally complete a `GC` action without error. If an error is encountered, a plugin should continue; removing as many resources as possible, and report the errors back to the runtime.
 
--- a/cnitool/cnitool.go
+++ b/cnitool/cnitool.go
@@ -68,7 +68,7 @@ func main() {
 	if netdir == "" {
 		netdir = DefaultNetDir
 	}
-	netconf, err := libcni.LoadConfList(netdir, os.Args[2])
+	netconf, err := libcni.LoadNetworkConf(netdir, os.Args[2])
 	if err != nil {
 		exit(err)
 	}
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -77,13 +77,13 @@ type PluginConfig struct {
 }
 
 type NetworkConfigList struct {
-	Name                       string
-	CNIVersion                 string
-	DisableCheck               bool
-	DisableGC    bool
+	Name                   string
+	CNIVersion             string
+	DisableCheck           bool
+	DisableGC              bool
 	LoadOnlyInlinedPlugins bool
-	Plugins                    []*PluginConfig
-	Bytes                      []byte
+	Plugins                []*PluginConfig
+	Bytes                  []byte
 }
 
 type NetworkAttachment struct {
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -67,18 +67,23 @@ type RuntimeConf struct {
 	CacheDir string
 }
 
-type NetworkConfig struct {
-	Network *types.NetConf
+// Deprecated: Use PluginConfig instead of NetworkConfig, the NetworkConfig
+// backwards-compat alias will be removed in a future release.
+type NetworkConfig = PluginConfig
+
+type PluginConfig struct {
+	Network *types.PluginConf
 	Bytes   []byte
 }
 
 type NetworkConfigList struct {
-	Name         string
-	CNIVersion   string
-	DisableCheck bool
+	Name                       string
+	CNIVersion                 string
+	DisableCheck               bool
 	DisableGC    bool
-	Plugins      []*NetworkConfig
-	Bytes        []byte
+	LoadOnlyInlinedPlugins bool
+	Plugins                    []*PluginConfig
+	Bytes                      []byte
 }
 
 type NetworkAttachment struct {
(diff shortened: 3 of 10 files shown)


SPEC.md
```diff
@@ -146,48 +147,61 @@ Plugins that consume any of these configuration keys should respect their intend
...
-      },
-      "sysctl": {
-        "net.core.somaxconn": "500"
-      }
-    },
-    {
-        "type": "portmap",
-        "capabilities": {"portMappings": true}
-    }
-  ]
```
s1061123, line 187: (I already mentioned in meeting, but add it for further discussion)

Is your PR completely changes the current style (deprecate current single JSON config) and introduce new one to replace? or Introduce new one additionally which co-exists with current style?

bleggett:
One question here - we are still carrying non-conflist format support in `libcni` - it's been ages since we dropped that from the spec, and I would suggest we actively drop support for that old format entirely as part of this, and cutting a new semver.

Otherwise things get quite messy with essentially 3 generations of config file support implemented in libcni, which I don't think we need to carry.

bleggett:
Basic impl up for people to concretely argue with, which should help. Note that there are no tests yet, I want to get a feeling for what people think before I add tests.

s1061123:
> One question here - we are still carrying non-conflist format support in libcni - it's been ages since we dropped that from the spec, and I would suggest we actively drop support for that old format entirely as part of this, and cutting a new semver.

Does this means CNI no longer support old CNI version which supports conf file? If so, all version will be deprecated other than 1.0.0.

bleggett:
> > One question here - we are still carrying non-conflist format support in libcni - it's been ages since we dropped that from the spec, and I would suggest we actively drop support for that old format entirely as part of this, and cutting a new semver.
> 
> Does this means CNI no longer support old CNI version which supports conf file? If so, all version will be deprecated other than 1.0.0.

I'm soliciting good arguments around why we should keep `libcni` support for parsing a discontinued format that has been deprecated for 3 years :D

In semver, you increase the major version to indicate to consumers of your {spec, library} that you are breaking backwards compatibility. We did that three years ago, but kept back compat parsing in libcni to ease the transition to 1.0.0.

Anyone still using the non-list format hasn't been following the spec for the last 3 years. This makes `libcni` harder to understand, and results in all implementation consumers having to juggle different types of files and manage upconverts (all of them have code for this).

I don't see a reason to release a new version of `libcni` with support for a non-current version of the spec at this point. It made sense for the 1.0.0 release itself, to let people convert to the 1.0.0 format, but for subsequent releases, we have no reason to support it.

If a libcni consumer genuinely needs to parse the old conf format, they can take an old release of `libcni`

bleggett:
@squeed PTAL - inverted flag as discussed.

SPEC.md
```diff
@@ -109,13 +109,15 @@ A network configuration consists of a JSON object with the following keys:
...
 - `cniVersions` (string list): List of all CNI versions which this configuration supports. See [version selection](#version-selection) below.
-- `name` (string): Network name. This should be unique across all network configurations on a host (or other administrative domain).  Must start with an alphanumeric character, optionally followed by any combination of one or more alphanumeric characters, underscore, dot (.) or hyphen (-).
+- `name` (string): Network name. This should be unique across all network configurations on a host (or other administrative domain).  Must start with an alphanumeric character, optionally followed by any combination of one or more alphanumeric characters, underscore, dot (.) or hyphen (-). Must not contain characters disallowed in file paths. A path segment (such as a filesystem directory) with the same name as the network name, containing one or more plugin configuration JSON objects for that network, should exist at the same level as the network configuration object itself.
```
squeed, line 112: I would drop the last sentence; this portion of the spec does not really concern configuration loading and parsing.

That the network name must be a valid file component is a useful (and implicit) restriction, so that's a good idea.

SPEC.md
```diff
@@ -109,13 +109,15 @@ A network configuration consists of a JSON object with the following keys:
...
 - `disableCheck` (boolean): Either `true` or `false`.  If `disableCheck` is `true`, runtimes must not call `CHECK` for this network configuration list.  This allows an administrator to prevent `CHECK`ing where a combination of plugins is known to return spurious errors.
-- `plugins` (list): A list of CNI plugins and their configuration, which is a list of plugin configuration objects.
+- `disableLoadPluginsFromPath` (boolean): Either `true` or `false`. If `false` (default), indicates [plugin configuration objects](#plugin-configuration-objects) should be loaded from a sibling directory with the same name as the network `name` field. These sibling directories should exist at the same path as the network configuration itself. Any valid plugin configuration objects within a named sibling directory will be appended to the final list of `plugins` for that network name. If set to `true`, plugin configuration objects in sibling directories will be ignored. If `plugins` is not present in the network configuration, `disableLoadPluginsFromPath` cannot be set to `true`.
```
squeed, line 114: I've been thinking about this a lot, and I'm of two minds as to whether or not we should allow configuring this or not.

This is not a network-specific option, rather a something the administrator would choose to do. That is to say, administrators should say (via ContainerD's configuration) whether or not they would like to disable drop-ins.

Given that this expresses administrative intent, and the whole purpose of this exercise is to allow for a cleaner meshing between vendor-supplied configurations, maybe we get rid of this option altogether? Thoughts?

SPEC.md
bleggett: > I would drop the last sentence; this portion of the spec does not really concern configuration loading and parsing.

Fair enough.

> That the network name must be a valid file component is a useful (and implicit) restriction, so that's a good idea.

Yeah, the intent was to explain _why_ the network name must be a valid filename, since saying it should and not explaining why feels a bit weird. Doing so does necessarily make config loading and parsing requirements more explicit/opinionated, tho.

SPEC.md
bleggett: > Given that this expresses administrative intent, and the whole purpose of this exercise is to allow for a cleaner meshing between vendor-supplied configurations, maybe we get rid of this option altogether? Thoughts?

It is mostly here for API compatibility with runtimes.

If we don't have this flag in the network config, we have to alter `libcni` config file loading APIs more drastically and backwards compatibility between network configs with inline plugins and network configs with subdirectory plugins gets pushed entirely to the container runtimes to manage.

Which feels a bit gross since it's our library, and our config, and our spec.

Putting the flag in the netconfig like this makes the back compat story really straightforward and will make integrating this into container runtimes pretty trivial, even if they don't change the libcni APIs they use, so for those reasons I'd prefer to keep it.


That being said, since it is an opt-in disable, a future spec revision could (and arguably should) drop it, which would force us to deprecate some of our libcni APIs and make container runtimes load netconfs and plugin confs themselves, separately and directly (they can do this today as well, if they choose to use the `NetworkPluginConfFromBytes` APIs added here)

SPEC.md
```diff
@@ -109,13 +109,15 @@ A network configuration consists of a JSON object with the following keys:
...
+- `plugins` (list): A list of inlined [plugin configuration objects](#plugin-configuration-objects). If this key is populated with inlined plugin objects, and `disableLoadPluginsFromPath` is true, the final set of plugins for a network must consist of all the plugin objects in this list, merged with all the plugins loaded from the sibling folder with the same name as the network.
 
 #### Plugin configuration objects:
-Plugin configuration objects may contain additional fields than the ones defined here.
-The runtime MUST pass through these fields, unchanged, to the plugin, as defined in section 3.
+All plugin configuration objects present in a directory with the same name as a valid network configuration must be parsed, and each plugin with a parsable configuration object must be invoked.
```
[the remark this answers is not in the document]
bleggett: Yep, your proposed wording is fine with me, this is really about what `libcni` does.

I do think we need (and I think you have talked about) docs that explain the difference between the spec and the actual libcni.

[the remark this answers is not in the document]
bleggett: Reworded this, ty!

SPEC.md
bleggett: Removed the sentence, ty!

SPEC.md
bleggett: In light of the other spec changes tho, if you want to make this `disableLoadPlugins` or `disablePluginAggregation`, open to that.

bleggett: Consensus here was: https://hackmd.io/jU7dQ49dQ86ugrXBx1De9w?both#2024-04-08

1. Keep opt-out in config
2. Rename this value to be loading-source-agnostic

bleggett:
Split above into 3 commits:

1. SPEC update for loading non-inlined plugin configs: https://github.com/containernetworking/cni/pull/1052/commits/99181ae7ff3bd2fb60fa348381325e9006a5ed75
1. Use type aliases to let us use `godoc` to mark certain pre-1.0 types as deprecated with dochints
1. `libcni` update for loading non-inlined plugin configs: https://github.com/containernetworking/cni/pull/1052/commits/9a1d54e79db80c7215f8f561f727d906a51f274d

bleggett:
Also threw up https://github.com/istio/istio.io/pull/14826/files based on some of the discuss here, as a start.

SPEC.md
```diff
@@ -109,13 +109,15 @@ A network configuration consists of a JSON object with the following keys:
...
+- `plugins` (list): A list of inlined [plugin configuration objects](#plugin-configuration-objects). If this key is populated with inlined plugin objects, and `loadOnlyInlinedPlugins` is true, the final set of plugins for a network must consist of all the plugin objects in this list, merged with all the plugins loaded from the sibling folder with the same name as the network.
 
 #### Plugin configuration objects:
-Plugin configuration objects may contain additional fields than the ones defined here.
-The runtime MUST pass through these fields, unchanged, to the plugin, as defined in section 3.
+Runtimes may aggregate plugin configuration objects from multiple sources, and must unambiguously associate each loaded plugin configuration object with a single, valid network configuration. All aggregated plugin configuration objects must be parsed, and each plugin with a parsable configuration object must be invoked.
```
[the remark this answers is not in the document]
bleggett: I effectively mean "validated" and was using "parsed" as shorthand for that, I'll just say `validated`.

SPEC.md
```diff
@@ -109,13 +109,15 @@ A network configuration consists of a JSON object with the following keys:
...
 - `disableCheck` (boolean): Either `true` or `false`.  If `disableCheck` is `true`, runtimes must not call `CHECK` for this network configuration list.  This allows an administrator to prevent `CHECK`ing where a combination of plugins is known to return spurious errors.
-- `plugins` (list): A list of CNI plugins and their configuration, which is a list of plugin configuration objects.
+- `loadOnlyInlinedPlugins` (boolean): Either `true` or `false`. If `false` (default), indicates [plugin configuration objects](#plugin-configuration-objects) should be loaded from a sibling directory with the same name as the network `name` field. These sibling directories should exist at the same path as the network configuration itself. Any valid plugin configuration objects within a named sibling directory will be appended to the final list of `plugins` for that network name. If set to `true`, plugin configuration objects in sibling directories will be ignored. If `plugins` is not present in the network configuration, `loadOnlyInlinedPlugins` cannot be set to `true`.
```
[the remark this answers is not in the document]
bleggett: Ugh, this wording is wrong now anyway, ty.

SPEC.md
```diff
@@ -146,18 +148,68 @@ Plugins that consume any of these configuration keys should respect their intend
...
+  "capabilities": {
+    "mac": true
+  },
+  "sysctl": {
+    "net.core.somaxconn": "500"
+  }
+}
+```
+
+Network configuration with one inlined plugin conf, and one loaded plugin conf:
```
squeed, line 198: Do you want to provide a bit more context for this example? Or maybe just drop it?

bleggett: I am leaning towards dropping it and moving it to https://github.com/containernetworking/cni/pull/1081

squeed: Heck yeah, totally approve.

MikeZappa87:
> One question here - we are still carrying non-conflist format support in `libcni` - it's been ages since we dropped that from the spec, and I would suggest we actively drop support for that old format entirely as part of this, and cutting a new semver.
> 
> Otherwise things get quite messy with essentially 3 generations of config file support implemented in libcni, which I don't think we need to carry.

This would need to be coordinated :-/ However we should do this unless people have a case for a CNI network configuration that executes a single CNI binary.

SPEC.md
bleggett: ...but given that we reference it in several existing spots, I'd rather do that in #1081

squeed (approved):
Looks basically good, but please split in to 3-or-so commits (spec changes, go renames, and implementation). Thanks!

bleggett:
> Looks basically good, but please split in to 3-or-so commits (spec changes, go renames, and implementation). Thanks!

Yep, had it like that at one point. Fixed up, ty!

squeed (approved):

squeed:
@bleggett this looks good to go, we just want to merge https://github.com/containernetworking/cni/pull/1090 and cut v1.2.1 first. Then we can merge this!

bleggett:
I'll resync this once #1090 + #1097 is in

bleggett:
@squeed @s1061123 PTAL - should be good from my perspective to merge when you wish.

squeed (approved):

PR #1053, opened 2023-12-13
spec: add attachment and endpoint definitions to specification

MikeZappa87:
We started this back in #903 however I am closing that and opening this one. Two comments by Casey/Henry should be reviewed.


SPEC.md
```diff
@@ -69,11 +69,13 @@ Released versions of the spec are available as Git tags.
...
 This document proposes a generic plugin-based networking solution for application containers on Linux, the _Container Networking Interface_, or _CNI_.
 
-For the purposes of this proposal, we define three terms very specifically:
+For the purposes of this proposal, we define six terms very specifically:
 - _container_ is a network isolation domain, though the actual isolation technology is not defined by the specification. This could be a [network namespace][namespaces] or a virtual machine, for example.
 - _network_ refers to a group of endpoints that are uniquely addressable that can communicate amongst each other. This could be either an individual container (as specified above), a machine, or some other network device (e.g. a router). Containers can be conceptually _added to_ or _removed from_ one or more networks.
 - _runtime_ is the program responsible for executing CNI plugins.
 - _plugin_ is a program that applies a specified network configuration.
+- _attachment is the operation of applying a network configuration to a container, which results in an endpoint.
+- _endpoint is a uniquely addressable, container-scoped connection to a single network.
```
squeed, line 78: we don't currently use endpoint in the SPEC; we use attachment. For example, see the GC section.

While I think that endpoint is a good bit of verbiage, we don't currently use it.

MikeZappa87: Should I remove it?

s1061123, line 78: > _endpoint is a uniquely addressable, container-scoped connection to a single network.

I'm just thinking v4/v6 network (dual-stack). In such case, _endpoint may addressed by two address. Should we change 'uniquely identified', not 'uniquely addressable'?

s1061123 (changes requested):

SPEC.md
squeed: Another thing that came up in the meeting: the spec makes no statements about addressability; it already has a notion of what is the primary key. So, we should say something like "uniquely identifiable" instead of addressable.

(aside: it's possible to have a container with no IPs at all :-) )

MikeZappa87: I made the change to say 'identifiable'

SPEC.md
```diff
@@ -69,11 +69,13 @@ Released versions of the spec are available as Git tags.
...
 This document proposes a generic plugin-based networking solution for application containers on Linux, the _Container Networking Interface_, or _CNI_.
 
-For the purposes of this proposal, we define three terms very specifically:
+For the purposes of this proposal, we define six terms very specifically:
 - _container_ is a network isolation domain, though the actual isolation technology is not defined by the specification. This could be a [network namespace][namespaces] or a virtual machine, for example.
 - _network_ refers to a group of endpoints that are uniquely addressable that can communicate amongst each other. This could be either an individual container (as specified above), a machine, or some other network device (e.g. a router). Containers can be conceptually _added to_ or _removed from_ one or more networks.
 - _runtime_ is the program responsible for executing CNI plugins.
 - _plugin_ is a program that applies a specified network configuration.
+- _attachment is the operation of applying a network configuration to a container, which results in an endpoint.
+- _endpoint is a uniquely identifiable, container-scoped connection to a single network.
```
henry118, line 78: both keywords here are missing the `_` suffix and the resulting syntax is not italics. otherwise lgtm.

PR #1060, opened 2024-02-01
Add MTU to CNI result

SchSeba:
This commit allow CNIs to expose the MTU of interface

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -567,6 +567,7 @@ Plugins must output a JSON object with the following keys upon a successful `ADD
 - `interfaces`: An array of all interfaces created by the attachment, including any host-level interfaces:
     - `name` (string): The name of the interface.
     - `mac` (string): The hardware address of the interface (if applicable).
+    - `mtu`: (uint) The MTU of the interface (if applicable).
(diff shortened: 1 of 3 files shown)


bleggett (approved):
There was a good point brought up that we could put this in a metadata struct, but I think MTU is pretty unambiguously a core property of the interface so it makes sense here to me.

Another good point was brought up that if MTU changes, this value could be stale, but I think that's generally true of most things here.

SPEC.md
```diff
@@ -567,6 +567,7 @@ Plugins must output a JSON object with the following keys upon a successful `ADD
 - `interfaces`: An array of all interfaces created by the attachment, including any host-level interfaces:
     - `name` (string): The name of the interface.
     - `mac` (string): The hardware address of the interface (if applicable).
+    - `mtu`: The MTU of the interface (if applicable).
```
s1061123, line 570: Could you add type in the fields? i.e. `(uint)`

s1061123 (changes requested):

SPEC.md
SchSeba: nice catch thanks!

s1061123 (approved):
/lgtm

Issue #1055, opened 2024-01-04
Failed to destroy the network because of empty cached cni result.


sofat1989:
```
failed to "KillPodSandbox" for "1b77faa7-5bc3-4b87-a56b-97918872b6f4" with KillPodSandboxError: "rpc error: code = Unknown desc = failed to destroy network for sandbox \"5bae61e683f7e59902f458514e9af1515d1c1c71ff069e9d4a2008a011307d04\": failed to get network \"generic-veth\" cached result: decoding version from network config: unexpected end of JSON input"


276776240 -rw------- 1 root root    0 Nov 27 23:16 generic-veth-5bae61e683f7e59902f458514e9af1515d1c1c71ff069e9d4a2008a011307d04-eth1
276779765 -rw------- 1 root root    0 Nov 27 23:16 istio-cni-5bae61e683f7e59902f458514e9af1515d1c1c71ff069e9d4a2008a011307d04-eth2
```

```
reboot   system boot  5.15.0-26-generi Mon Nov 27 23:20 - 00:10 (30+00:49)
reboot   system boot  5.15.0-26-generi Mon Nov 27 23:16 - 00:10 (30+00:53)
reboot   system boot  5.15.0-26-generi Mon Nov 27 23:13 - 00:10 (30+00:56)
```

Recently we met an issue. When the containerd is adding the networks for pods, but at that time, the system is rebooted. It will leave some empty files under /var/lib/cni/results. And after the system is up again, but the pods are deleted by the users. containerd will try to read the cached result to destroy the networks. It will fail directly.  

For the cached file, the data will be put in the prevResult field. It seems that a lot of plugins only care about the previous result of the other plugins chained together. And it doesn't care about the previous result of itself. 

Can we delay reporting the error of cached result and try to delete the network with cachedResult=nil ?
 
```
// DelNetwork executes the plugin with the DEL command
 func (c *CNIConfig) DelNetwork(ctx context.Context, net *NetworkConfig, rt *RuntimeConf) error {
        var cachedResult types.Result
+       var cachedResultErr error

        // Cached result on DEL was added in CNI spec version 0.4.0 and higher
        if gtet, err := version.GreaterThanOrEqualTo(net.Network.CNIVersion, "0.4.0"); err != nil {
@@ -567,12 +571,14 @@ func (c *CNIConfig) DelNetwork(ctx context.Context, net *NetworkConfig, rt *Runt
        } else if gtet {
[...]

PR #1066, opened 2024-02-15
Set cache dir to volatile dir

henry118:
This was meant to be a question but since it's a pretty easy change I'll just use a PR to kick off the discussion.

The question is that **should CNI cache result persist across node restart?**

My understanding is that most networking setups will not persist across node reboot, and libcni will not recover those network setups either. Pods/container recovery after reboot is expected to be driven by the runtime. Today in k8s, pods are normally rescheduled once a node is restarted, therefore any cached CNI result will immediately go stale. It could be a problem if the same node continues to get new pods scheduled on it.

Should we set the cache dir to something volatile, say `/run/cni` or `/var/run/cni`?

Likewise I had https://github.com/containerd/containerd/issues/9825 to enable the customization of this dir in runtime. Just unsure where the best place of the change should be...

xref: https://github.com/containerd/containerd/issues/9825
xref: https://github.com/containernetworking/cni/issues/1055

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -37,7 +37,7 @@ import (
 )
 
 var (
-	CacheDir = "/var/lib/cni"
+	CacheDir = "/var/run/cni"
 	// slightly awkward wording to preserve anyone matching on error strings
 	ErrorCheckNotSupp = fmt.Errorf("does not support the CHECK command")
 )


henry118:
@squeed @MikeZappa87 or others?

MikeZappa87:
I am not certain of the backstory to why the cache ended up in /var/lib vs /var/run. @squeed @dcbw

alaypatel07:
Interested in answers to the questions here. Running into issues where cni cache directory is corrupted with 0 bytes files

alaypatel07:
@squeed There is already an issue for the error we are running into https://github.com/containernetworking/cni/issues/1055.

What would break if upon reboot the results directory is deleted?

henry118:
This issue is likely a result of OS crash, where dirty pages were lost before persisted to disk. containerd had similar issue with its runtime data and had to introduce an *optional* `fsync` as a safeguard (https://github.com/containerd/containerd/pull/9401). Can we do something similar in libcni?

alaypatel07:
@henry118 so I understand the issue better, if case of OS crash, are all the files in `/var/lib/cni/results` expected to be of size 0 or just some of them which were dirtied?

henry118:
Opened #1072 as alternative solution. Closing this thread.

PR #1072, opened 2024-03-07
tolerate invalid cni caches for deletion

henry118:
Fixes: #1055
Closes: #1066

In `DelNetworkList` function, ignore the errors returned by `getCachedResult` and continue to invoke plugins for deletion as best effort. Also updated the test case to adapt this new behavior.

Cached results are deleted from disk in the following two cases:
1. `getCachedResult` returns an error; OR
2. `getCachedResult` succeeds and no errors are returned from plugins

Cached results are not deleted if (I think this will allow clients to retry the deletion)
1. `getCachedResult` succeeds AND 
2. Plugin fails


henry118:
@squeed @s1061123 PTAL

squeed (approved):

PR #1076, opened 2024-03-22
cnitool: refactor, add GC, convert to workspace

squeed:
This refactors cnitool to be a modern cobra-style program. It also adds STATUS and GC support.

Finally it uses go workspaces to allow for separate `go.mod` files.


cnitool/main.go
```diff
@@ -0,0 +1,267 @@
...
+	for _, netns := range args[1:] {
+		pair := strings.Split(netns, ":")
+		ifname := "eth0"
+		if len(pair) > 1 {
+			ifname = pair[1]
+		}
+
+		validAttachments = append(validAttachments, types.GCAttachment{
+			IfName:      ifname,
+			ContainerID: containerID(netns),
```
maiqueb, line 238: this is wrong AFAIU - you're checking the containerID based on the "full" netns value, which can include the name of the interface.

PR #1089, opened 2024-05-13
supplement ns building files for darwin OS

Icarus9913:
Since The PR #890 add ns validation, it introduced the OS building incompatibility into `skel` usage.  

And you can also see that under https://github.com/containernetworking/cni/tree/main/pkg/ns folder, we only have `linux` and `windows` which make it inconvenient to do some Unit Test for the `darwin` OS.

So, this PR supplement ns building files for darwin OS. Then you can also test the `skel` buildings in the `darwin` OS


squeed (changes requested):
This seems to have some unrelated changes -- can you remove those please?

Icarus9913:
> This seems to have some unrelated changes -- can you remove those please?

Done

squeed (approved):

PR #1095, opened 2024-05-27
libcni: always delete the cache on conflist for CNI DEL

maiqueb:
This aligns the call with `DelNetwork`, and allows CRIO tests to bump - [0].

[0] - https://github.com/cri-o/cri-o/pull/8207#issuecomment-2129937741

### DIFF
--- a/libcni/api.go
+++ b/libcni/api.go
@@ -598,9 +598,7 @@ func (c *CNIConfig) DelNetworkList(ctx context.Context, list *NetworkConfigList,
 		}
 	}
 
-	if cachedResult != nil {
-		_ = c.cacheDel(list.Name, rt)
-	}
+	_ = c.cacheDel(list.Name, rt)
 
 	return nil
 }


libcni/api.go
```diff
@@ -598,9 +598,7 @@ func (c *CNIConfig) DelNetworkList(ctx context.Context, list *NetworkConfigList,
 		}
 	}
 
-	if cachedResult != nil {
-		_ = c.cacheDel(list.Name, rt)
-	}
+	_ = c.cacheDel(list.Name, rt)
```
aojea, line 601: why should delete if there is no cachedResult?

maiqueb: - the result is ignored; if there is no cached result, nothing bad will happen
- that is the [same behavior](https://github.com/containernetworking/cni/blob/c04330ee9e0bb932e1e3eda757813e2f212e746b/libcni/api.go#L668) as in `DelNetwork`
- it allows us to [bump the ocicni library in crio](https://github.com/cri-o/cri-o/pull/8207#issuecomment-2129937741) to use the status verbs now.

The alternatives would be to stop checking the results in the cache in the crio tests (since the tests use CNI 0.3.1 ...) or to fix a lot of tests to use CNI version >= 0.4.0 .

This was discussed in the [CNI maintainers meeting](https://hackmd.io/jU7dQ49dQ86ugrXBx1De9w) may 27th 2024, and seemed the simpler way forward, given it is totally innocuous.

aojea: > * that is the [same behavior](https://github.com/containernetworking/cni/blob/c04330ee9e0bb932e1e3eda757813e2f212e746b/libcni/api.go#L668) as in `DelNetwork`

that seems a fair point

/lgtm

squeed (approved):

PR #1097, opened 2024-06-03
Fix faulty json.Marshal behavior for embeds types.NetConf

### LINKED ISSUE #1096: Faulty json.Marshal behavior for structs that embeds types.NetConf since v1.2.0
containernetworking/cni#1039 changed the behavior of json.Marshal for struct that embeds `types.NetConf`. All fields in your struct except `types.NetConf` will be ignored when calling json.Marshal.

Here are examples:
With CNI v1.2.0: https://go.dev/play/p/L4feVRtl5S4
With CNI v1.1.2: https://go.dev/play/p/4RrrgKJUEN3

Users implementing their own CNI plugins could face the issue when updating their `github.com/containernetworking/cni` dependency to v1.2.0.

containernetworking/cni#1039 could be reverted to fix the problem.

s1061123:
Fix #1096


LionelJouin (commented):
This way the users would need to change `types.NetConf` with `types.NetConfType`: https://go.dev/play/p/j2wjAgA_s-K
The current way would still not work: https://go.dev/play/p/fHVOmTrH8p_K

s1061123:
@LionelJouin Updated to avoid to change user side code.

LionelJouin (approved):
Looks good: https://go.dev/play/p/7ncmsg8-uL0

s1061123:
@squeed PTAL?

PR #1098, opened 2024-06-10
SPEC: add warning about preserving shared resources for GC

squeed:
For plugins that may share resources or resource pools across networks, we should make it explicitly clear that GC must only clean up resources known to be owned by the calling network.


s1061123 (approved):
LGTM

Issue #1101, opened 2024-06-27
Error in specification regarding GC??


kanpov:
CNI spec says extra JSON key in GC is `cni.dev/attachments` yet libcni uses `cni.dev/valid-attachments`? What is it then if the spec doesn't even correspond to the official runtime?

kanpov:
I'm asking this because I'm developing a CNI runtime and this lack of clarity is a problem

squeed:
Oops! That is a very silly mistake indeed.

My plan is to change the spec to match libcni, as libni is probably more directly meaningful. In the unlikely event that there is spec-conformant code, I will set both variables. Make sense?

kanpov:
> Oops! That is a very silly mistake indeed.
> 
> My plan is to change the spec to match libcni, as libni is probably more directly meaningful. In the unlikely event that there is spec-conformant code, I will set both keys. Make sense?

Yes, probably, but I think the spec should keep `cni.dev/attachments` and not be changed, but only libcni is changed to actually adhere the spec. What do you think?

bleggett:
My $0.02 is that 

- `cni.dev/valid-attachments` is a better name anyway
- We don't have back compat concerns at this point for this feature

So changing the spec and throwing the back compat concern into `libcni` like #1103 does puts us into a better state.

kanpov:
> My $0.02 is that 
> 
> - `cni.dev/valid-attachments` is a better name anyway
> - We don't have meaningful back compat concerns at this point for this feature
> 
> So tweaking the spec and throwing the back compat concern into `libcni` like #1103 does puts us into a better state.

Sure, but there's no way this change doesn't require a version bump (that is, if CNI follows semver) to 1.2

PR #1103, opened 2024-07-01
SPEC, libcni: harmonize GC valid-attachment key

squeed:
The SPEC said valid attachments should be `cni.dev/attachments`, whereas libcni used `cni.dev/valid-attachments`. Oops.

This changes the SPEC to match libcni, *and* has libcni set both variables, just in case.

Fixes: #1101

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -387,7 +387,7 @@ The runtime MUST NOT use GC as a substitute for DEL. Plugins may be unable to cl
 
 The runtime must provide a JSON-serialized plugin configuration object (defined below) on standard in. It contains an additional key;
 
-- `cni.dev/attachments` (array of objects): The list of **still valid** attachments to this network:
+- `cni.dev/valid-attachments` (array of objects): The list of **still valid** attachments to this network:
     - `containerID` (string): the value of CNI_CONTAINERID as provided during the CNI ADD operation
     - `ifname` (string): the value of CNI_IFNAME as provided during the CNI ADD operation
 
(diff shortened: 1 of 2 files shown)


squeed:
@kanpov thanks for noticing this!

libcni/api.go
```diff
@@ -817,6 +817,8 @@ func (c *CNIConfig) GCNetworkList(ctx context.Context, list *NetworkConfigList,
 		}
 		if args != nil {
 			inject["cni.dev/valid-attachments"] = args.ValidAttachments
+			// GH-1101: spec used incorrect variable name
+			inject["cni.dev/attachments"] = args.ValidAttachments
```
s1061123, line 821: Looks good to me from cording perspective, but how about to change `#1101` from JIRA-ish name?

SPEC.md
```diff
@@ -387,7 +387,7 @@ The runtime MUST NOT use GC as a substitute for DEL. Plugins may be unable to cl
 
 The runtime must provide a JSON-serialized plugin configuration object (defined below) on standard in. It contains an additional key;
 
-- `cni.dev/attachments` (array of objects): The list of **still valid** attachments to this network:
+- `cni.dev/valid-attachments` (array of objects): The list of **still valid** attachments to this network:
```
bleggett, line 390: :+1:

bleggett (approved):

libcni/api.go
squeed: Fixed.

MikeZappa87 (approved):

PR #1155, opened 2025-04-07
libcni: remove some deprecation warnings

squeed:
We were a bit over-eager with the deprecations; these types and aliases are doing no harm :-).


dougbtv (approved):
lgtm!

PR #1173, opened 2025-11-10
libcni: add OpenTelemetry trace context propagation

### LINKED ISSUE #561: libcni should support distributed tracing
Broadly, it needs to pass a span identifier across from the caller to the plugin.

Probably something to do with OpenTracing.  Details left as an exercise.

jipperinbham:
This enables distributed tracing support by passing span identifiers from the caller to CNI plugins through environment variables. When a valid span context exists, it is propagated using the TRACEPARENT, TRACESTATE, and BAGGAGE environment variables following the OpenTelemetry specification [OTEP 0258](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/0258-env-context-baggage-carriers.md).

This allows CNI plugins to participate in distributed traces, helping with performance analysis and debugging across the plugin execution lifecycle. I does not create any spans and instead leaves that responsibility to the caller and CNI plugins. Below is an example for extracting the trace context via environment variables:

```Go
func extractTraceparent(ctx context.Context) context.Context {
	traceParent, ok := os.LookupEnv("TRACEPARENT")
	if !ok {
		return ctx
	}

	traceParts := strings.Split(traceParent, "-")
	if len(traceParts) != 4 {
		return ctx
	}

	traceID, err := trace.TraceIDFromHex(traceParts[1])
	if err != nil {
		return ctx
	}

	spanID, err := trace.SpanIDFromHex(traceParts[2])
	if err != nil {
		return ctx
	}

	flagBytes, err := hex.DecodeString(traceParts[3])
	if err != nil {
		return ctx
	}

	var ts trace.TraceState
	if traceState, ok := os.LookupEnv("TRACESTATE"); ok {
		if ts, err = trace.ParseTraceState(traceState); err != nil {
			// no-op
		}
	}

	// traceBaggage, err := baggage.Parse(os.Getenv("BAGGAGE"))
	// if err != nil {
	// 	slog.With("error", err.Error()).Error(fmt.Sprintf("Failed to parse baggage %s", err.Error()))
	// }

	scc := trace.SpanContextConfig{
		TraceID:    traceID,
		SpanID:     spanID,
		TraceFlags: trace.TraceFlags(flagBytes[0]),
		TraceState: ts,
		Remote:     true,
	}

	return trace.ContextWithSpanContext(ctx, trace.NewSpanContext(scc))
}
``` 

Fixes #561


squeed (approved):

go.mod
```diff
@@ -1,20 +1,22 @@
 module github.com/containernetworking/cni
 
-go 1.21
+go 1.23.0
```
jipperinbham, line 3: I'm guessing this is causing the CI failure. @squeed would you prefer I rollback to 1.21?

MikeZappa87:
@jipperinbham can you fix the linter and we can move this forward?

jipperinbham:
@MikeZappa87 thanks for the ping! I've updated this work and https://github.com/containerd/go-cni/pull/132 to having passing CI.

PR #1179, opened 2026-01-04
perf(libcni): optimize config file filtering performance

lucming:


squeed:
Why would we want this? We want to find all files...

lucming:
We should break as soon as we find a match because each file will only have one relevant suffix anyway.

PR #1183, opened 2026-04-21
fix: move exec inside loop

### LINKED ISSUE #855: Run `c := exec.CommandContext(ctx, pluginPath)` inside the for loop
https://github.com/containernetworking/cni/blob/master/pkg/invoke/raw_exec.go#L34
```
	stdout := &bytes.Buffer{}
	stderr := &bytes.Buffer{}
	c := exec.CommandContext(ctx, pluginPath)
	c.Env = environ
	c.Stdin = bytes.NewBuffer(stdinData)
	c.Stdout = stdout
	c.Stderr = stderr

	// Retry the command on "text file busy" errors
	for i := 0; i <= 5; i++ {
		err := c.Run()

		// Command succeeded
		if err == nil {
			break
		}

		// If the plugin is currently about to be written, then we wait a
		// second and try it again
		if strings.Contains(err.Error(), "text file busy") {
			time.Sleep(time.Second)
			continue
		}

		// All other errors except than the busy text file
		return nil, e.pluginErr(err, stdout.Bytes(), stderr.Bytes())
	}

	// Copy stderr to caller's buffer in case plugin printed to both
	// stdout and stderr for some reason. Ignore failures as stderr is
	// only informational.
	if e.Stderr != nil && stderr.Len() > 0 {
		_, _ = stderr.WriteTo(e.Stderr)
	}
	return stdout.Bytes(), nil
```
It seems like that response contains unexpected values if it fails with "text file busy" and then succeeds.
Should we update it as below?

```
	var stdout *bytes.Buffer
	var stderr *bytes.Buffer

	// Retry the command on "text file busy" errors
	for i := 0; i <= 5; i++ {
                stdout = &bytes.Buffer{}
                stderr = &bytes.Buffer{}
                c := exec.CommandContext(ctx, pluginPath)
                c.Env = environ
                c.Stdin = bytes.NewBuffer(stdinData)
                c.Stdout = stdout
                c.Stderr = stderr

		err := c.Run()

		// Command succeeded
		if err == nil {
			break
		}

		// If the plugin is currently about to be written, then we wait a
		// second and try it again
		if strings.Contains(err.Error(), "text file busy") {
			time.Sleep(time.Second)
			continue
		}

		// All other errors except than the busy text file
		return nil, e.pluginErr(err, stdout.Bytes(), stderr.Bytes())
	}

	// Copy stderr to caller's buffer in case plugin printed to both
	// stdout and stderr for some reason. Ignore failures as stderr is
	// only informational.
	if e.Stderr != nil && stderr.Len() > 0 {
		_, _ = stderr.WriteTo(e.Stderr)
	}
	return stdout.Bytes(), nil
```

QxBytes:
Aims to resolve https://github.com/containernetworking/cni/issues/855

Moves exec command into the for loop so each retry gets a new instance of the command. Previously, we would retry using the same exec command which would always fail due to https://cs.opensource.google/go/go/+/refs/tags/go1.26.2:src/os/exec/exec.go;l=146 (command cannot be reused after calling `Run` on it).

Reran the unit test several dozen times without any flakiness.

### DIFF
--- a/pkg/invoke/raw_exec.go
+++ b/pkg/invoke/raw_exec.go
@@ -32,12 +32,14 @@ type RawExec struct {
 }
 
 func (e *RawExec) ExecPlugin(ctx context.Context, pluginPath string, stdinData []byte, environ []string) ([]byte, error) {
-	var stdout, stderr *bytes.Buffer
+	stdout := &bytes.Buffer{}
+	stderr := &bytes.Buffer{}
 
 	// Retry the command on "text file busy" errors
 	for i := 0; i <= 5; i++ {
-		stdout = &bytes.Buffer{}
-		stderr = &bytes.Buffer{}
+		stdout.Reset()
+		stderr.Reset()
+
 		c := exec.CommandContext(ctx, pluginPath)
 		c.Env = environ
 		c.Stdin = bytes.NewBuffer(stdinData)
--- a/pkg/invoke/raw_exec.go
+++ b/pkg/invoke/raw_exec.go
@@ -32,16 +32,18 @@ type RawExec struct {
 }
 
 func (e *RawExec) ExecPlugin(ctx context.Context, pluginPath string, stdinData []byte, environ []string) ([]byte, error) {
-	stdout := &bytes.Buffer{}
-	stderr := &bytes.Buffer{}
-	c := exec.CommandContext(ctx, pluginPath)
-	c.Env = environ
-	c.Stdin = bytes.NewBuffer(stdinData)
-	c.Stdout = stdout
-	c.Stderr = stderr
+	var stdout, stderr *bytes.Buffer
 
 	// Retry the command on "text file busy" errors
 	for i := 0; i <= 5; i++ {
+		stdout = &bytes.Buffer{}
+		stderr = &bytes.Buffer{}
+		c := exec.CommandContext(ctx, pluginPath)
+		c.Env = environ
+		c.Stdin = bytes.NewBuffer(stdinData)
+		c.Stdout = stdout
+		c.Stderr = stderr
+
 		err := c.Run()
 
 		// Command succeeded
--- a/pkg/invoke/raw_exec_test.go
+++ b/pkg/invoke/raw_exec_test.go
@@ -18,6 +18,9 @@ import (
 	"bytes"
 	"context"
 	"os"
+	"os/exec"
+	"runtime"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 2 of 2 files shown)


MikeZappa87:
@squeed @LionelJouin this might have been under the radar for awhile

pkg/invoke/raw_exec.go
```diff
@@ -32,16 +32,18 @@ type RawExec struct {
...
-	c.Env = environ
-	c.Stdin = bytes.NewBuffer(stdinData)
-	c.Stdout = stdout
-	c.Stderr = stderr
+	var stdout, stderr *bytes.Buffer
 
 	// Retry the command on "text file busy" errors
 	for i := 0; i <= 5; i++ {
+		stdout = &bytes.Buffer{}
+		stderr = &bytes.Buffer{}
```
jellonek: Why you are creating new objects instead of just calling https://pkg.go.dev/bytes#Buffer.Reset at the start of each iteration?

pkg/invoke/raw_exec.go
```diff
@@ -32,16 +32,18 @@ type RawExec struct {
...
+		c := exec.CommandContext(ctx, pluginPath)
+		c.Env = environ
```
jellonek, line 44: is there a possibility that c.Env could change between iterations?

pkg/invoke/raw_exec.go
```diff
@@ -32,16 +32,18 @@ type RawExec struct {
...
+		c.Stdin = bytes.NewBuffer(stdinData)
```
jellonek, line 45: why you are recreating it again and again, while it could be simply resetted?

jellonek (commented):
IMO good idea + nice tests, while implementation - i'm not sure about it...

pkg/invoke/raw_exec.go
QxBytes: Updated the PR

pkg/invoke/raw_exec.go
QxBytes: `c.Env` would come from `environ` which I don't expect to change during the lifetime of the call.

pkg/invoke/raw_exec.go
QxBytes: If I resetted like `stdout` or `stderr` I'd need to write the stdinData back in right? I expect the retries case to be hit very rarely.

MikeZappa87:
@QxBytes can you rebase off the latest?

squeed (approved):
this seems reasonable. Can you rebase?

QxBytes:
@MikeZappa87  @squeed rebased off the latest

PR #1190, opened 2026-07-05
types: add gws field to Route for ECMP support

yushoyamaguchi:
Add a GWs []net.IP field to the Route struct to enable ECMP (Equal-Cost Multi-Path) routing via multiple next hops.

When both GW and GWs are set, GW is treated as an additional next hop appended after GWs. The NextHops() method returns the combined list for use by plugin implementations.

Also document the gws field and its merge behavior in SPEC.md.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -585,6 +585,7 @@ Plugins must output a JSON object with the following keys upon a successful `ADD
 - `routes`: Routes created by this attachment:
     - `dst`: The destination of the route, in CIDR notation
     - `gw`: The next hop address. If unset, a value in `gateway` in the `ips` array may be used.
+    - `gws` : A list of next hop addresses for ECMP (Equal-Cost Multi-Path) routing. Each entry is an IP address string. Use this instead of `gw` when the route has multiple gateways.
     - `mtu` (uint): The MTU (Maximum transmission unit) along the path to the destination.
     - `advmss` (uint): The MSS (Maximal Segment Size) to advertise to these destinations when establishing TCP connections.
     - `priority` (uint): The priority of route, lower is higher.
--- a/SPEC.md
+++ b/SPEC.md
@@ -585,6 +585,7 @@ Plugins must output a JSON object with the following keys upon a successful `ADD
 - `routes`: Routes created by this attachment:
     - `dst`: The destination of the route, in CIDR notation
     - `gw`: The next hop address. If unset, a value in `gateway` in the `ips` array may be used.
+    - `gws` : A list of next hop addresses for ECMP (Equal-Cost Multi-Path) routing. Each entry is an IP address string. Use this instead of `gw` when the route has multiple gateways. Duplicate IP addresses in `gws` are not allowed.
     - `mtu` (uint): The MTU (Maximum transmission unit) along the path to the destination.
     - `advmss` (uint): The MSS (Maximal Segment Size) to advertise to these destinations when establishing TCP connections.
     - `priority` (uint): The priority of route, lower is higher.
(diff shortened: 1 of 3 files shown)


pkg/types/types.go
```diff
@@ -228,6 +234,17 @@ func (r *Route) Copy() *Route {
 	return route
 }
 
+// NextHops returns the combined gateway list for ECMP routing.
+// If GW is set, it is appended after GWs.
+func (r *Route) NextHops() []net.IP {
+	gws := make([]net.IP, len(r.GWs))
+	copy(gws, r.GWs)
+	if r.GW != nil {
```
squeed: Can you check to see if `r.GWs` contains `r.GW`? If so, it should not be appended (only unique IPs).

yushoyamaguchi: I've fixed!

SPEC.md
```diff
@@ -585,6 +585,7 @@ Plugins must output a JSON object with the following keys upon a successful `ADD
 - `routes`: Routes created by this attachment:
     - `dst`: The destination of the route, in CIDR notation
     - `gw`: The next hop address. If unset, a value in `gateway` in the `ips` array may be used.
+    - `gws` (list of strings, optional): A list of next hop addresses for ECMP (Equal-Cost Multi-Path) routing. Each entry is an IP address string. If both `gw` and `gws` are set, `gw` is treated as an additional next hop appended after the entries in `gws`. If only `gw` is set, behavior is unchanged from routes without `gws`.
```
s1061123: In case of `gw` and `gws` case, usually plugin author tend to contain all ECMP routes into `gws`, not split into `gw` + `gws`. In addition, the consumer of Result object need to add `gw` and `gws` for all route inforation and it could be (a bit) burden. Hence I would like to know the background of this design. Could you please let me know why you design it?

In such `gw` and `gws` case, at least to me, I would like to 

- use `gws` for ECMP supported Result consumer
- use `gw` for non-ECMP supported Result consumer (i.e. plugin just returns one of them)

s1061123: Specification document needs to specify what is required and what should NOT. In this case, how about to contain same IP address in `gws`? Is that still valid?

s1061123: > If only `gw` is set, behavior is unchanged from routes without `gws`. 

In this section, we describe `gws`. So could you make `gws` as a subject?

s1061123 (changes requested):

SPEC.md
yushoyamaguchi: @s1061123 
As you said, I think that it is **NOT** desirable to write nexthops in both `gw` and `gws` fields, too.
(I think we should write this to SPEC, so I'll add the description after this discussion.)

In the community MTG, we talk about what we should do when **user define both `gw` and `gws`**.
However, we talk about only two option.
- Ignore `gw`.
- Merge `gw` to `gws`

and we choose merging way.

Do you think `plugins` should return **ERROR** in such case ?

Otherwise, do you think we don't need to mention such case in the SPEC?

yushoyamaguchi: Sorry, I've fixed.

yushoyamaguchi: @s1061123 
I've fixed as you described in last MTG.
- Add an explanation to `SPEC.md` that `gws` should be used instead of `gw` when configuring ECMP routes.
- Remove the description of how to handle cases where both `gw` and `gws` are defined, since this behavior should not be defined in the spec (It will be decided when implementing in `plugins`).

yushoyamaguchi: I've fixed.

squeed (approved):
Looks good to me. I think the spec changes are reasonable.

yushoyamaguchi:
@s1061123 
Sorry, could you review, again?

mlguerrero12 (approved):

wizhaoredhat (approved):
LGTM

s1061123 (approved):

PR #1192, opened 2026-08-18
Skip nil elements when converting result versions

thc1006:
## What happens

`[null]` is valid JSON, so a result like `{"ips":[null]}` unmarshals into a slice that holds a nil pointer. When that result is converted to another spec version, the IP, interface, and route converters read the element directly and hit a nil pointer dereference. `convertIPConfigFrom040` at `pkg/types/100/types.go:149` reaches for `from.Address` on a nil `*IPConfig`, and `convertInterfaceFrom040`, `convertIPConfigTo040`, `convertInterfaceTo040`, and the two inline loops in `convertTo02x` (`pkg/types/040/types.go`) do the same. The neighbouring DNS and route paths do not crash because they go through the nil-safe `Copy()`.

## Why it matters

The null does not have to be malicious. It reaches the converters on ADD through `result.GetAsVersion`, on a chained plugin through `NewResultFromResult(conf.PrevResult)`, and on CHECK, DEL, and GC through `getCachedResult`, which re-converts a cached result. A single buggy or hand-written plugin result, a crafted `prevResult`, or a tampered cache file panics the library, and on the cached-result path that panic can land inside the runtime that called into CNI. It is not reachable from unauthenticated network input.

## Reproducer

```go
r, _ := create.CreateFromBytes([]byte(`{"cniVersion":"0.4.0","ips":[null]}`))
r.GetAsVersion("1.0.0")
// panic: runtime error: invalid memory address or nil pointer dereference
```

## The change

Skip nil elements in the six conversion loops so a null entry is dropped rather than dereferenced, which keeps these paths in line with the nil-safe `Copy()` siblings. The added specs cover the up-convert (0.4.0 to 1.0.0), the down-convert (1.0.0 to 0.4.0), and the 0.4.0 to 0.2.0 route path; each panics before this change and passes after.

### DIFF
--- a/pkg/types/040/types.go
+++ b/pkg/types/040/types.go
@@ -149,7 +149,11 @@ func convertTo02x(from types.Result, toVersion string) (types.Result, error) {
 		DNS:        *fromResult.DNS.Copy(),
 	}
 
+	// A JSON null in the source array unmarshals to a nil element; skip it.
 	for _, fromIP := range fromResult.IPs {
+		if fromIP == nil {
+			continue
+		}
 		// Only convert the first IP address of each version as 0.2.0
 		// and earlier cannot handle multiple IP addresses
 		if fromIP.Version == "4" && toResult.IP4 == nil {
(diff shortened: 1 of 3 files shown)


thc1006:
Duplicate of #1191, which I opened first and which has slightly broader test coverage. Closing this in favor of that one.

